PR #13, opened 2025-07-30
Commands that have colored output are treated as interactive so git-ai doesn't kill the colors.

mm-zacharydavison:
I noticed that `git-ai` killed colors for things like `git status` in my terminal (`fish`).
I've added those commands to the interactive set so that colors get maintained (by using `spawn()`).

Let me know if this is the wrong solution, I'm happy to fix it another way.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -366,6 +366,7 @@ fn get_default_remote(repo: &git2::Repository) -> Option<String> {
 
 fn proxy_to_git(args: &[String]) {
     // Check if this is an interactive command that needs special handling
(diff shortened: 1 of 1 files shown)


acunniffe:
This looks right @mm-zacharydavison. I'd rather pass as much through directly as we can to preserve existing behaviors. Thanks for writing it this way. 

Merging and releasing now. Look for it in 0.2.4

acunniffe:
It's out https://github.com/acunniffe/git-ai/actions/runs/16731687992

PR #44, opened 2025-09-30
First class support for Cursor via Hooks Preset

acunniffe:
The Cursor team enabled hooks feature flag for some git-ai contributors. It's in 1.7.x https://cursor.com/nightlydownload and will launch more broadly soon. 

We reported a few bugs/limitations to the team that we expect them to fix

- [ ] Unlike claude code, cursor does not use the system shell's configuration (looks like they just run bash). This makes it difficult to reliably call user installed binaries x-platform. Hoping this is resolved, for now a mac compatible solution `zsh -c 'source ~/.zprofile; source ~/.zshrc;` is being used for testing
- [x] Shared a preference for`beforeFileEdit`. _It's on their roadmap and hoping it makes the cut for public release_ 
- [ ] Clarification - will these hooks work with `cursor-agent `CLI or just the IDE.

To test, add the following to `~/.cursor/hooks.json`

```json
{
  "version": 1,
  "hooks": {
    "beforeSubmitPrompt": [
      {
        "command": "zsh -c 'source ~/.zprofile; source ~/.zshrc; git-ai checkpoint cursor --hook-input \"$(cat)\"' >/dev/null 2>&1"
      }
    ],
    "afterFileEdit": [
      {
        "command": "zsh -c 'source ~/.zprofile; source ~/.zshrc; git-ai checkpoint cursor --hook-input \"$(cat)\"' >/dev/null 2>&1"
      }
    ]
  }
}
```

You'll notice both of these commands are the same -- for the cursor preset git-ai is responsible for deciding whether it should mark code as AI or human based on the JSON stdin. It's not safe to assume people have jq or other JSON tools on their shell so we're insourcing that concern. When the `hook_event_name` is  `beforeSubmitPrompt` a human checkpoint is created, when it is `afterFileEdit` an AI checkpoint is created. 
```
{
  "conversation_id": string // an identifier for the conversation as a whole
  "hook_event_name": string // the name of the hook this is being called in response to
  "workspace_roots": string[] // The roots of the folders in the workspace.  Usually just one, but multiroot can have many
}
```

Known issues: 

- [ ] Race condition. The last assistant message is not persisting until after the hook runs. Planning to fetch latest messages when preparing the authorship logs in post-commit. The many seconds between code being generated, reviewed, tested and committed should guarantee changes have hit the SQLlite on disk.

### DIFF
--- a/src/commands/checkpoint_agent/agent_preset.rs
+++ b/src/commands/checkpoint_agent/agent_preset.rs
@@ -16,7 +16,8 @@ pub struct AgentCheckpointFlags {
 
 pub struct AgentRunResult {
     pub agent_id: AgentId,
-    pub transcript: AiTranscript,
+    pub is_human: bool,
+    pub transcript: Option<AiTranscript>,
     pub repo_working_dir: Option<String>,
 }
 
@@ -79,7 +80,8 @@ impl AgentCheckpointPreset for ClaudePreset {
 
         Ok(AgentRunResult {
             agent_id,
-            transcript,
+            is_human: false,
+            transcript: Some(transcript),
             // use default.
             repo_working_dir: None,
         })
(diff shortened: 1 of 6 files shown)


src/git/post_commit.rs
```diff
@@ -47,13 +48,16 @@ pub fn post_commit(repo: &Repository, force: bool) -> Result<(String, Authorship
...
-    let filtered_working_log = filter_untracked_files(repo, &parent_working_log)?;
+    let mut filtered_working_log = filter_untracked_files(repo, &parent_working_log)?;
 
     debug_log(&format!(
         "Working log entries: {}",
         filtered_working_log.len()
     ));
 
+    // mutates inline
+    CursorPreset::update_cursor_conversations_to_latest(&mut filtered_working_log)?;
```
acunniffe, line 59: fix for race condition

src/commands/checkpoint_agent/agent_preset.rs
```diff
@@ -792,13 +495,10 @@ impl CursorPreset {
                         if let Some(capability_type) = bubble_content.get("capabilityType") {
                             if let Some(cap_num) = capability_type.as_i64() {
                                 let model = match cap_num {
+                                    // @todo Aidan to figure out the rest of these mappings
+                                    30 => "gpt-5-codex",
```
acunniffe, line 506: This list is incomplete. Have to generate a line or two of code with every model and figure out the rest. Coming back to this later in the week

acunniffe:
Screenshot 2025-09-30 at 1 58 32 PM

This isn't super stable...updated and now workspace_root is missing so no way to anchor ourself to a specific repo. We could get the workspace via the SQL db, conversations have an edge to it, but that's definitely more hacky.

PR #60, opened 2025-10-02
install hooks support for cursor and claude code

acunniffe:
- Cursor has released hooks publicly, so we're adding it to `install-hooks`. 
- Claude code hooks updated 
- `--dry-run` implemented so users can see proposed edits before they are made

Screenshot 2025-10-02 at 12 59 22 PM


acunniffe:
@svarlamov we def want a headless version of this for enterprise setup, but for independent developers should I add a flag here to make this interactive? 

>> "welcome to git ai..." let's install your hooks

"here are the changes we want to make to claude code settings ... approve (y) (n)..."

PR #68, opened 2025-10-03
Handle git aliased commands

svarlamov:


acunniffe:
Great approach!

acunniffe:
@svarlamov is this going to be merged?

svarlamov:
> @svarlamov is this going to be merged?

Soon! I just want to build the git config cache before I do this because it's a lot of extra calls potentially

svarlamov:
Closing because I think we need to have a centralized way to manage the git config cache so that we don't have to make a `git` call for every config we want to test. It seems like it'd be easy to build this ourselves or potentially use a git-oxide rust library

PR #98, opened 2025-10-08
Bug/fix notes merging behavior

### LINKED ISSUE #95: Conflicting authorship files
### Description:

`Git-ai blame` doesn't work properly in the fresh clone when 2 branches have changes in the same file.

### Steps to reproduce
- Create 2 clones of a repository
- Create one branch in both of the clones with different names
- Modify the same file with AI in both of the clones and commit changes
- Push first branch to remote
- Push second branch to remote
- Create another clone of the repo
- Fetch both branches from remote

### Observed behaviour
- First branch doesn't show correct `git-ai blame <file>` 
- `git notes --ref ai show <commit sha>` doesn't work for latest commit in first branch
- `git-ai blame <file>` and `git notes` work correctly in second branch. 
- Fetching git notes refs in first clone introduces conflict
```
  atnes-test-repo git:(my-test-1) git fetch origin refs/notes/ai:refs/notes/ai
From bitbucket.org:atlassian/atnes-test-repo
 ! [rejected]        refs/notes/ai -> refs/notes/ai  (non-fast-forward)
```

svarlamov:
Fixes #95 

When we migrated to git notes the other day, we failed to handle the new problem of merging notes. Unlike the independent commit-based AI refs, git notes are trees and so even though each note is still unrelated to the other notes, the tree structure itself needs to get merged. The problem was hidden for a while, because the push hook was passing a prop that would override the tree if it diverged.

This PR solves this by having clients merge their local ai note tree (using the `union` method) with the latest remote ai note tree. It also removes the force flag on the push. It makes the push a bit slower (need to fetch remote notes before pushing our own), but I think there are many possible optimizations to be made still.

@acunniffe Can you please make sure that this plays nicely with the rebase/squash/etc. work?

More detailed AI-generated commit summary below (I think it does a good job of explaining the small but somewhat complex changes)

## Problem Statement

When two independent clones of a repository create AI-authored commits on different branches and push to the same remote, the git notes containing authorship metadata conflict. This manifests as:

1. `git-ai blame <file>` fails to show AI attribution for the first branch pushed
2. `git notes --ref ai show <commit>` returns no results for affected commits
3. Manual fetch of notes produces non-fast-forward rejection errors
4. Fresh clones only receive notes from the most recent push, losing historical authorship data

## Root Cause

The issue stems from using a **force refspec** (`+refs/notes/ai:refs/notes/ai`) for both push and fetch operations. Git notes are stored as a tree structure under `refs/notes/ai`, where each note is a blob mapped to a commit SHA. When using force push:

1. Clone A pushes branch `feature-1` → Creates notes tree T1 on remote
2. Clone B independently pushes branch `feature-2` → **Force-overwrites** remote with tree T2
3. Tree T1 is completely lost, orphaning all notes for commits in `feature-1`

This occurs because notes form a single shared tree structure, not independent per-commit blobs. Each push/fetch operation affects the entire tree, making force operations destructive.

## Solution

Implemented a **merge-based synchronization workflow** that prevents data loss:

### Architecture

1. **Tracking Refs Pattern** - Remote notes are fetched into isolated tracking refs (`refs/notes/ai-remote/{remote}`) rather than directly into `refs/notes/ai`
2. **Automatic Merging** - After fetching, tracking refs are merged into local notes using `git notes merge -s union`
3. **Non-Force Push** - Notes are pushed using `refs/notes/ai:refs/notes/ai` (without `+` prefix), requiring fast-forward

### Implementation Details

**1. Helper Functions ([src/git/refs.rs](../src/git/refs.rs#L142-L201))**
```rust
- tracking_ref_for_remote() // Generates refs/notes/ai-remote/{remote}
- sanitize_remote_name()     // Ensures valid ref names (replaces special chars)
- ref_exists()               // Checks ref existence via show-ref
- merge_notes_from_ref()     // Executes git notes merge -s union
- copy_ref()                 // Initial setup for new tracking refs
```

**2. Fetch Hook ([src/commands/fetch_hooks.rs](../src/commands/fetch_hooks.rs#L49-L98))**
```
fetch origin +refs/notes/ai:refs/notes/ai-remote/origin
  ↓
if refs/notes/ai exists:
  git notes --ref=ai merge -s union refs/notes/ai-remote/origin
else:
  git update-ref refs/notes/ai refs/notes/ai-remote/origin
```

**3. Push Hook ([src/commands/push_hooks.rs](../src/commands/push_hooks.rs#L56-L117))**
```
Pre-push:
  fetch origin +refs/notes/ai:refs/notes/ai-remote/origin
  git notes --ref=ai merge -s union refs/notes/ai-remote/origin

Post-push:
  git push origin refs/notes/ai:refs/notes/ai  # No force!
```

## Merge Strategy: Union

Uses `git notes merge -s union` which concatenates notes content. This is safe because:
- Each commit SHA is globally unique → notes for different commits never conflict
- Authorship logs are JSON blobs → concatenation doesn't corrupt structure
- Alternative strategies like `cat_sort_uniq` could deduplicate lines but aren't needed for JSON data

## Workflow Example

**Before (Buggy):**
```
Clone A: Creates note for commit abc123 → Push → Remote has {abc123}
Clone B: Creates note for commit def456 → Push → Remote has {def456} (abc123 LOST!)
Clone C: Fetch → Only sees {def456}
```

**After (Fixed):**
```
Clone A: Creates note for commit abc123 → Push → Remote has {abc123}
Clone B: Creates note for commit def456 → Pre-push fetch+merge → Local has {abc123, def456}
        Push → Remote has {abc123, def456}
Clone C: Fetch → Receives {abc123, def456} → Merge → All notes preserved
```

### 1. `git push --mirror`
**Status:** Partially handled
- The post-push hook is **skipped** for `--mirror` operations to avoid double-pushing notes (mirror already pushes all refs)
- If a user hasn't recently fetched notes before running `--mirror`, their `refs/notes/ai` may be stale and overwrite remote notes
- **Mitigation:** Mirror operations are rare (typically for backups/migrations) and users performing them generally expect to push their exact local state
- **Future improvement:** Could add a pre-push hook to fetch+merge before mirror, but adds complexity for minimal benefit

### 2. Tracking Ref Pollution
**Status:** Not an issue
- Tracking refs are stored at `refs/notes/ai-remote/{remote}` which is outside standard push namespaces
- Git push behavior by namespace:
  - `git push` → only `refs/heads/*` (branches)
  - `git push --all` → only `refs/heads/*`
  - `git push --tags` → only `refs/tags/*`
  - `git push --mirror` → all refs (but hook skips this case)
- Tracking refs remain local-only unless explicitly pushed via `git push origin refs/notes/ai-remote/*` (extremely unlikely)

### 3. Race Conditions
**Status:** Acceptable failure mode
- If two clones push simultaneously, one may fail with non-fast-forward error
- **Behavior:** Failed push logs error via `debug_log()` but doesn't fail user's operation (best-effort)
- User can retry push, which will fetch+merge and succeed
- This is standard git behavior for concurrent pushes

### 4. Merge Conflicts
**Status:** Theoretically possible, practically impossible
- The `union` strategy auto-resolves by concatenation
- Conflicts could only occur if the same commit somehow has different notes in tracking ref vs local ref
- This can't happen in normal operation since notes are created deterministically from authorship logs
- If it did occur, merge would fail and log error (best-effort handling)

## Backward Compatibility

- **No migration required** - Existing repositories automatically upgrade on first fetch/push
- **Legacy refspec preserved** - `AI_AUTHORSHIP_REFSPEC` constant retained for potential future use
- **Graceful degradation** - If notes sync fails (network issues, etc.), errors are logged but user operations proceed

## Testing Considerations

The fix resolves the exact scenario from the bug report:
1. Two clones modify the same file on different branches
2. Both push to remote
3. Fresh clone fetches both branches
4. `git-ai blame` now works correctly for **both** branches
5. No non-fast-forward errors when fetching notes

## Performance Impact

- **Fetch operations:** Adds one `git fetch` + one `git notes merge` per fetch (negligible overhead)
- **Push operations:** Adds one `git fetch` + one `git notes merge` + one `git push` (best-effort, non-blocking)
- **Storage:** Tracking refs add minimal overhead (~one ref per remote, typically just `origin`)

## Security Considerations

- All git operations run with `core.hooksPath=/dev/null` to prevent recursive hook invocation
- Remote names are sanitized using `sanitize_remote_name()` to prevent ref injection attacks
- Operations are best-effort and cannot fail user's git commands

### DIFF
--- a/scripts/benchmarks/git/benchmark_notes_merge.py
+++ b/scripts/benchmarks/git/benchmark_notes_merge.py
@@ -1,0 +1,228 @@
+#!/usr/bin/env python3
+"""
+Benchmark git notes fetch and merge performance at various scales.
+
+This script simulates the git-ai workload:
+- Single notes ref (refs/notes/ai)
+- Fetch into tracking ref (refs/notes/ai-remote/origin)
+- Merge from tracking ref into local refs/notes/ai using "ours" strategy
+- Notes contain JSON-like content (simulating authorship logs)
+"""
+
+import subprocess
+import tempfile
+import shutil
+import time
+import os
+import sys
+import json
+from pathlib import Path
+
+
+def run_git(cmd, cwd, env=None):
+    """Run a git command and return the result."""
+    full_cmd = ["/opt/homebrew/bin/git"] + cmd
+    result = subprocess.run(
+        full_cmd,
+        cwd=cwd,
+        capture_output=True,
+        text=True,
+        env={**os.environ, **(env or {})}
+    )
+    if result.returncode != 0:
+        print(f"Command failed: {' '.join(full_cmd)}")
+        print(f"Error: {result.stderr}")
+    return result
+
+
+def generate_authorship_note(commit_hash, index):
+    """Generate a realistic authorship note JSON."""
+    return json.dumps({
+        "metadata": {
+            "schema_version": "3.0",
+            "commit": commit_hash,
+            "timestamp": "2025-10-08T12:00:00Z"
+        },
+        "session": {
+            "id": f"session_{index}",
+            "checkpoints": [
+                {
+                    "type": "user_prompt",
+                    "content": f"Implement feature {index}",
+                    "timestamp": "2025-10-08T12:00:00Z"
+                },
+                {
+                    "type": "ai_response",
+                    "content": f"Here's the implementation for feature {index}...",
+                    "timestamp": "2025-10-08T12:01:00Z"
+                }
+            ]
+        },
+        "authorship": {
+            "ai_percentage": 0.75,
+            "human_percentage": 0.25
+        }
+    }, indent=2)
+
+
+def create_test_repo(base_dir, name, num_commits, num_notes):
+    """Create a test repository with specified number of commits and notes."""
+    repo_path = base_dir / name
+    repo_path.mkdir()
+
+    print(f"Creating {name} with {num_commits} commits and {num_notes} notes...")
+
+    run_git(["init", "-b", "main"], repo_path)
+    run_git(["config", "user.name", "Test User"], repo_path)
+    run_git(["config", "user.email", "test@example.com"], repo_path)
+
+    # Create commits
+    test_file = repo_path / "test.txt"
+    commits = []
+    for i in range(num_commits):
+        test_file.write_text(f"Commit {i}\n")
+        run_git(["add", "test.txt"], repo_path)
+        run_git(["commit", "-m", f"Commit {i}"], repo_path)
+
+        # Get the commit hash
+        result = run_git(["rev-parse", "HEAD"], repo_path)
+        commits.append(result.stdout.strip())
+
+        if (i + 1) % 1000 == 0:
+            print(f"  Created {i + 1} commits...")
+
+    # Add notes to commits using refs/notes/ai
+    for i, commit in enumerate(commits[:num_notes]):
+        note_content = generate_authorship_note(commit, i)
+        run_git(["notes", "--ref=ai", "add", "-f", "-m", note_content, commit], repo_path)
+
+        if (i + 1) % 1000 == 0:
+            print(f"  Created {i + 1} notes...")
+
+    print(f"  ✓ Created {name}")
+    return repo_path
+
+
+def clone_repo(source, dest):
+    """Clone a repository."""
+    run_git(["clone", str(source), str(dest)], source.parent)
+    # Fetch notes into tracking ref (simulating git-ai's fetch pattern)
+    tracking_ref = "refs/notes/ai-remote/origin"
+    run_git(["fetch", "origin", f"+refs/notes/ai:{tracking_ref}"], dest)
+    # Copy tracking ref to local notes ref
+    run_git(["update-ref", "refs/notes/ai", tracking_ref], dest)
+
+
+def benchmark_fetch_merge(repo_path, num_new_notes):
+    """Benchmark fetch and merge operations matching git-ai's workload."""
+    print(f"\nBenchmarking with {num_new_notes} new commits and notes...")
+
+    # Get the remote repo path (origin)
+    result = run_git(["remote", "get-url", "origin"], repo_path)
+    remote_path = Path(result.stdout.strip())
+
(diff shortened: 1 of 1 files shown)


AtnesNess:
Thank you for quick fix! 

> (need to fetch remote notes before pushing our own

I'm curious whether pulling notes history can get longer over time so the performance becomes terrible? Say we have 1000 commits per day, at some point pulling git-notes might be pretty expensive operation

svarlamov:
Benchmarks:

```
============================================================
SUMMARY
============================================================
Total Notes     New Notes    Fetch (s)    Merge (s)    Total (s)   
------------------------------------------------------------
1,000           10           0.072        0.007        0.078       
1,000           100          0.127        0.007        0.134       
1,000           500          0.302        0.008        0.310       
1,000           1000         0.270        0.010        0.280 
------------------------------------------------------------
10,000          10           0.109        0.007        0.116       
10,000          100          0.233        0.008        0.241       
10,000          500          0.437        0.008        0.445       
10,000          1000         0.506        0.010        0.516 
------------------------------------------------------------
50,000          10           5.450        0.008        5.458       
50,000          100          5.678        0.008        5.686       
50,000          500          2.817        0.008        2.825       
50,000          1000         0.587        0.010        0.596
------------------------------------------------------------
100,000          10          18.242       0.009        18.251
100,000          100         16.334       0.009        16.343
100,000          500         18.392       0.009        18.401
100,000          1000        3.353        0.011        3.364
```

Questions:
**Will merge be slow?**
No, merges, even on huge repos, are really fast. This was our biggest concern as there wasn't a lot of clarity from docs, etc. as to how large git note tree merges would be handled. This was also the only net-new task that git-ai wasn't already doing previous to this PR.

**Will fetch be slow?**
Git makes fetch really fast even on large repos, especially when you're mostly synced up already. Git has a protocol to ensure minimal data transfer and the maintenance keeps the look up fast server-side. You can see that the fetch time drops precipitously after the server-side maintenance/repacking is done, and had I gave it more time it would probably drop even further. On average, fetch time should scale linearly with _new_ notes only. Shouldn't be an issue, especially with well-optimized git servers.

PR #101, opened 2025-10-09
[wip] Implemented git rebase and git reset authorship log reconstruction

acunniffe:
local support for rewriting authorship after
- [x] rebase interactive and non-interactive mode
- [x] reset `--hard` `--soft` and `--mixed`

### DIFF
--- a/docs/limitations.mdx
+++ b/docs/limitations.mdx
@@ -39,7 +39,7 @@ title: Compatibility and Limitations
 | After resolving git conflicts, attribution is correct | ✅ |
 | Rebase correctly merges attribution | ✅ |
 | reset `--soft` and `--mixed` maintains correct attribution | ✅ |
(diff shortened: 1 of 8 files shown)


acunniffe:
@svarlamov can you give any color on why windows tests were failing / if it's important for us to get them working?

svarlamov:
> @svarlamov can you give any color on why windows tests were failing / if it's important for us to get them working?

The test was using a shell script which won't work on Windows. Poked a bit at trying to use a windows batch test, but a couple passes with the AI didn't help and I don't have a machine to iterate on right now. I think it's fine to revisit later -- it's not a functionality issue with our code AFAICT

acunniffe:
💪 makes sense. I'm going to adjust how the test rig works. Thought of a lot of ways to improve it this morning.

PR #102, opened 2025-10-10
Massive improvement to commit performance 

acunniffe:
`git-ai` was starting to feel unusable is massive repos. 

We've adjusted how we diff in pre and post commit to optimize the final checkpointing and unstaging of authorship

**before**
JPEG image

**after**
Screenshot 2025-10-10 at 2 12 41 PM


acunniffe:
succeeded by https://github.com/acunniffe/git-ai/pull/104

PR #114, opened 2025-10-20
make sync authorship work for bare or normal repos

acunniffe:
Improving the new sync authorship functions so they play nicer with bare repos. Doing some work on webhooks / CI bots now where this had suddenly become very important

### DIFF
--- a/src/git/sync_authorship.rs
+++ b/src/git/sync_authorship.rs
@@ -1,7 +1,6 @@
 use crate::git::refs::{
     AI_AUTHORSHIP_PUSH_REFSPEC, copy_ref, merge_notes_from_ref, ref_exists, tracking_ref_for_remote,
 };
-use crate::git::repository::CommitRange;
 use crate::{
     error::GitAiError,
     git::{cli_parser::ParsedGitInvocation, repository::exec_git},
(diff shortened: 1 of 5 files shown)


src/git/sync_authorship.rs
```diff
@@ -43,18 +42,63 @@ pub fn fetch_remote_from_args(
...
     let tracking_ref = tracking_ref_for_remote(&remote_name);
+
+    debug_log(&format!(
+        "fetching authorship notes for remote '{}' to tracking ref '{}'",
+        remote_name, tracking_ref
+    ));
+
+    // First, check if the remote has refs/notes/ai using ls-remote
+    // This is important for bare repos where the refmap might not be configured
+    let mut ls_remote_args = repository.global_args_for_exec();
```
acunniffe, line 59: @svarlamov is this cool to use here? Before we were passing parsed CLI args down into the fetch and push function -- when running on the server it didn't have those, but does seem to work with global_args_for_exec because I've been using

```
pub fn find_repository_in_path(path: &str) -> Result<Repository, GitAiError> {
    let global_args = vec!["-C".to_string(), path.to_string()];
    return find_repository(&global_args);
}
```

svarlamov: Technically, I think that we should keep passing the parsed_args. This is so that any global args the user had when they ran fetch, continue being respected.

I see the complexity with getting a parsed args object tho. You can generate them from a string with `parse_git_cli_args` or maybe just have a server version of the function without them?

acunniffe: Yeah but won't they get pulled in if it's been invoked as the git wrapper? 

```
    pub fn global_args_for_exec(&self) -> Vec<String> {
        let mut args = self.global_args.clone();
        if !args.iter().any(|arg| arg == "--no-pager") {
            args.push("--no-pager".to_string());
        }
        args
    }
```

Or do you mean the command args will be missed w/ this approach?

PR #124, opened 2025-10-22
Fixed some git compat issues causing stats to not work in bare repos

acunniffe:
In CI or Webhooks (where people want to accumulate git-ai stats), it's common to make a bare clone. Some of the git execs we were making were breaking for this use case

### DIFF
--- a/src/commands/git_ai_handlers.rs
+++ b/src/commands/git_ai_handlers.rs
@@ -405,7 +405,6 @@ fn handle_stats(args: &[String]) {
     if let Some(range) = commit_range {
         match range_authorship::range_authorship(range, true) {
             Ok(stats) => {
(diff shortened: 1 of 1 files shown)


src/git/refs.rs
```diff
@@ -53,59 +53,40 @@ pub fn get_commits_with_notes_from_list(
...
-    for line in stdout.lines() {
-        if let Some((sha, author)) = line.split_once('|') {
-            commit_authors.insert(sha.to_string(), author.to_string());
-        }
-    }
-
-    // Check which commits have notes
-    let mut args = repo.global_args_for_exec();
-    args.push("cat-file".to_string());
-    args.push("--batch-check=%(objectname)".to_string());
```
acunniffe, line 79: This was being too clever

PR #131, opened 2025-10-23
agent preset v1

acunniffe:
Added generic agent preset and documentation for quicker integrations


test-agent-preset-v1.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
```
acunniffe, line 1: @svarlamov if you want to test this out locally you could run this shell script

PR #163, opened 2025-10-28
parallelize fetch | push | pull

acunniffe:
Since notes syncing won't contend with regular fetch | push | pull requests there's no reason to do them serially. Performance minded teams are asking for every ounce of optimization and this seems like a low risk way to improve our posture.


svarlamov (approved):
Looks great and definitely noticed the speed up

PR #167, opened 2025-10-29
Fix PR refs not being persisted during GitHub CI squash workflow.

mm-zacharydavison:
While adding integration tests, I noticed that if we didn't explicitly fetch the refs into a local ref, they'd be lost during the CI.
This fix explicitly stores the PR refs fetched from GitHub in `refs/github/pr/{}`.

I've also added:
- integration test
- `install_github_ci_workflow` for installing github ci workflows into integration test repos
- `get_workflow_logs` for fetching workflow logs  for debugging integration tests
- `wait_for_workflows` for waiting for GH actions to complete

### DIFF
--- a/tests/github/deleted_branch_pr_test.rs
+++ b/tests/github/deleted_branch_pr_test.rs
@@ -71,7 +71,7 @@ fn test_squash_merge_with_deleted_branch() {
     println!("✅ Squash merged and deleted branch");
 
     // Wait for GitHub CI workflow to complete
-    test_repo.wait_for_workflows(&pr_number, 300)
+    test_repo.wait_for_workflows(300)
         .expect("GitHub CI workflow failed or timed out");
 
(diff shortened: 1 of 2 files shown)


src/ci/github.rs
```diff
@@ -83,12 +83,13 @@ pub fn get_github_ci_context() -> Result<Option<CiContext>, GitAiError> {
     // Fetch PR commits using GitHub's special PR refs
     // This is necessary because the PR branch may be deleted after merge
     // but GitHub keeps the commits accessible via pull/{number}/head
+    // We store the fetched commits in a local ref to ensure they're kept
     exec_git(&[
         "-C".to_string(),
         clone_dir.clone(),
         "fetch".to_string(),
         "origin".to_string(),
```
mm-zacharydavison: actually, i shouldn't assume the remote is called `origin`.

PR #171, opened 2025-10-30
Add 'upgrade' command and automatic updates.

mm-zacharydavison:
Resolves https://github.com/acunniffe/git-ai/issues/87

Adds support for release channels (`latest`, `next`) per #198 

```
git-ai upgrade              // checks for updates and upgrades if there is a new version
git-ai upgrade --force      // force re-installs the latest version
```

Hooks on fetch and push run automated version checks in background processes. Unless auto updates are disabled, it will automatically update as well.

If an update is available and auto updates are disabled, it will print:

```
A new version of git-ai is available: (1.1.2) → (1.1.3)
Run 'git-ai upgrade' to upgrade to the latest version.
```

`git-ai upgrade` delegates to the existing `install.sh` or `install.ps1` script.

- [x] Install testing
- [x] Config testing
- [x] Automated check/install testing
- [x] Update install scripts to stop overwriting config if it already exists (both windows and bash)
- [ ] Windows testing or temporarily disable on Windows

### DIFF
--- a/src/commands/upgrade.rs
+++ b/src/commands/upgrade.rs
@@ -7,6 +7,12 @@ use std::io::IsTerminal;
 use std::sync::atomic::{AtomicBool, AtomicU64, Ordering};
 use std::time::{Duration, SystemTime, UNIX_EPOCH};
 
+#[cfg(windows)]
+use std::os::windows::process::CommandExt;
+
+#[cfg(windows)]
+const CREATE_NO_WINDOW: u32 = 0x08000000;
+
 const UPDATE_CHECK_INTERVAL_HOURS: u64 = 24;
 const INSTALL_SCRIPT_URL: &str =
     "https://raw.githubusercontent.com/acunniffe/git-ai/main/install.sh";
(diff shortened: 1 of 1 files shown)


mm-zacharydavison:
Sorry guys, I've got a full plate this week, so don't have time to look at the proposed changes until next week, documenting here in-case anyone else wants to pick it up in the meantime:

1. Add options to disable automated version checks and auto updates (forward-looking feature) via the git-ai config (for enterprises that are using custom releases/MDM)
2. Can we put the version check an async thread that runs on fetch+pull+push pre command books? Similar to this: https://github.com/acunniffe/git-ai/blob/feat/sentry/src/commands/hooks/fetch_hooks.rs#L16. The thinking here is that these commands naturally take a bit longer, so it’s more likely to finish and we won’t add any startup latency. IDEs run fetch in the background a lot, so it should get picked up pretty regularly.
3. When auto updates are disabled, only output the update alert when we’re being run in an interactive terminal (this will make sure it doesn’t impact output parsing for IDEs, etc. that run git/git-ai directly and are reliant on certain output formats). When auto updates are enabled, I don’t think we need to output anything (just update silently).
4. We’ve gotten reports of users getting rate limited on [api.github.com](http://api.github.com/) endpoints when running on corporate networks. Is there any way to use a plain [github.com](http://github.com/) raw endpoint or something like that?

docs/enterprise-configuration.mdx
```diff
@@ -21,6 +21,9 @@ All options in `config.json` are optional and will fall back to default values i
 | `exclude_repositories` | `Pattern[]` | Exclude `git-ai` from these remotes. Supports glob patterns (e.g., `https://github.com/myorg/*`) | If a repository is present in both allow and exclude lists, exclusion takes precedence |
 | `telemetry_oss` | `"off"` | Disable OSS performance metrics and error logging sent to Git AI maintainers | Defaults to enabled |
 | `telemetry_enterprise_dsn` | `string` | A Sentry DSN to use to send yourself performance metrics and error logging | Defaults to none |
+| `disable_version_checks` | `boolean` | Skip automated version checks that would otherwise run on fetch/pull/push | `false` |
+| `disable_auto_updates` | `boolean` | Keep checking for updates but never install them automatically | `false` |
+| `update_channel` | `"latest" \| "next"` | Release channel to follow (`latest` = stable, `next` = prerelease) | `"latest"` |
```
acunniffe, line 26: Thinking through the case where enterprises fork and MDM Git AI. I don't expect they will won't want our release channels to update binaries after we distribute them.

Can we default these to `true` (disabled) and set up an `option_env!("OSS_BUILD")` that toggles these to false in OSS release pipeline? So if it came from the OSS Channel it auto updates, if it doesn't come from our release channel it doesn't try to auto update by default. 

I imagine enterprises might want their own channels too, but TBD. Many will fork and build and hit this footgun. Seems like a lot of these companies have ways to control what's on each dev's bin

acunniffe (approved):

PR #178, opened 2025-11-01
OSS + Enterprise Sentry support

acunniffe:
This PR introduces support for logging performance metrics + errors to Sentry compatible endpoints

OSS releases will have telemetry enabled by default starting in `1.0.15`. There's a flag in `config.json` to disable it

```
   "telemetry_oss": "off"
```

Enterprises can set their own DSN using 

```
   "telemetry_enterprise_dsn": "<dsn>"
```

Or by building from source and setting `SENTRY_ENTERPRISE=<dsn> cargo build`

--

**what's logged?**

- Any errors from a pre commit or post commit hook (should not include user data).  
- Performance violations. if we're outside our target overhead for wrapping git commands, the timings for pre/post commit hooks will be logged. 
- operating system, version
- remote URLs of repos

--

**performance impact**

because git-ai is a git wrapper and does not want to introduce any major overhead to a git command. So we're trading realtime logs for performance. Logs are buffered in an append-only`ai/log` file (one per pid to prevent locking). Logs are then sent to the server only when you do a git fetch or git push. A background process is enqueued to mop up the logs and send them off without slowing those commands down.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -750,7 +750,7 @@ dependencies = [
 
 [[package]]
 name = "git-ai"
-version = "1.0.14"
+version = "1.0.15"
 dependencies = [
  "assert_cmd",
  "chrono",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "git-ai"
-version = "1.0.14"
+version = "1.0.15"
 edition = "2024"
 
 


src/observability/wrapper_performance_targets.rs
```diff
@@ -0,0 +1,85 @@
+use std::{ops::Add, time::Duration};
+
+use serde_json::json;
+
+use crate::{
+    authorship::working_log::CheckpointKind, observability::log_performance, utils::debug_log,
+};
+
+pub fn log_performance_target_if_violated(
```
acunniffe, line 12: @svarlamov let's sync on defining our internal targets.

svarlamov: Sounds good. The biggest thing I noticed in playing around with it is that we probably need to set a floor on the 10% (at least for now). So like `max(10% or 70ms)`

```
➜  git-ai git:(feat/sentry) GIT_AI_DEBUG_PERFORMANCE=1 git commit
On branch feat/sentry
Your branch is up to date with 'origin/feat/sentry'.

nothing to commit, working tree clean
[git-ai (perf)] ᕽ Performance target violated for command: commit. Total duration: 51ms, Git duration: 7ms. Pre-command: 44ms, Post-command: 0ms.
```

svarlamov (approved):
Everything looks good. Left some thoughts on the thresholds

PR #179, opened 2025-11-02
Improve checkpoint and post-commit performance 

acunniffe:
This PR fixes a few bugs
- [x] Checkpoints were not using the edited_files optimizations for commits with no AI edits. 
- [x] Post commit was still creating initial checkpoints instead of an `INITIAL` attestation entry. On Chromium we went from 2600ms to 242ms
- [x] Expensive calls to `git diff` that weren't needed

Still todo
- [x] Some Amend and reset tests are failing. need to debug

### DIFF
--- a/src/authorship/authorship_log_serialization.rs
+++ b/src/authorship/authorship_log_serialization.rs
@@ -1068,103 +1068,105 @@ mod tests {
         assert_eq!(entry.line_ranges[1], LineRange::Range(8, 10));
     }
 
-    #[test]
-    fn test_metrics_calculation() {
-        use crate::authorship::attribution_tracker::{Attribution, LineAttribution};
-        use crate::authorship::transcript::{AiTranscript, Message};
-        use crate::authorship::working_log::{
-            AgentId, Checkpoint, CheckpointKind, WorkingLogEntry,
-        };
-        use std::time::{SystemTime, UNIX_EPOCH};
+    // Commenting out because working log to authorship helper deprecated in favor
+    // of virtual attribution
+    // #[test]
+    // fn test_metrics_calculation() {
+    //     use crate::authorship::attribution_tracker::{Attribution, LineAttribution};
+    //     use crate::authorship::transcript::{AiTranscript, Message};
+    //     use crate::authorship::working_log::{
+    //         AgentId, Checkpoint, CheckpointKind, WorkingLogEntry,
+    //     };
+    //     use std::time::{SystemTime, UNIX_EPOCH};
 
-        // Create an agent ID
-        let agent_id = AgentId {
-            tool: "cursor".to_string(),
-            id: "test_session".to_string(),
-            model: "claude-3-sonnet".to_string(),
-        };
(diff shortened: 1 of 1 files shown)


src/commands/checkpoint.rs
```diff
@@ -352,31 +357,45 @@ fn get_all_files(
...
+/// Get all files that should be tracked, including those from previous checkpoints and INITIAL attributions
+///
 fn get_all_tracked_files(
     repo: &Repository,
     _base_commit: &str,
     working_log: &PersistedWorkingLog,
     edited_filepaths: Option<&Vec<String>>,
 ) -> Result<Vec<String>, GitAiError> {
-    let mut files = get_all_files(repo, edited_filepaths)?;
+    let mut files: HashSet<String> = edited_filepaths
```
acunniffe, line 361: @svarlamov per our slack discussion, this is the simplified checkpoint logic

src/commands/git_ai_handlers.rs
```diff
@@ -264,7 +264,25 @@ fn handle_checkpoint(args: &[String]) {
...
+                    // Find the git repository
+                    let repo = match find_repository_in_path(&working_dir) {
+                        Ok(repo) => repo,
+                        Err(e) => {
+                            eprintln!("Failed to find repository: {}", e);
+                            std::process::exit(1);
+                        }
+                    };
+                    match repo.get_staged_and_unstaged_filenames() {
+                        Ok(filenames) => {
```
acunniffe: @svarlamov per our slack discussion, this is how we can support mock_ai going forward

acunniffe: In chromium, this path takes 2 seconds, but then the actual checkpoint takes 200 ms.

src/utils.rs
```diff
@@ -9,14 +9,16 @@ static DEBUG_PERFORMANCE_ENABLED: std::sync::OnceLock<bool> = std::sync::OnceLoc
...
+            || std::env::var("GIT_AI_DEBUG_PERFORMANCE").unwrap_or_default() == "1"
     })
 }
 
 fn is_debug_performance_enabled() -> bool {
-    is_debug_enabled() && *DEBUG_PERFORMANCE_ENABLED.get_or_init(|| {
-        std::env::var("GIT_AI_DEBUG_PERFORMANCE").unwrap_or_default() == "1" || std::env::var("GIT_AI_DEBUG_PERFORMANCE").unwrap_or_default() != "0"
-    })
+    is_debug_enabled()
+        && *DEBUG_PERFORMANCE_ENABLED
```
acunniffe, line 20: I think this was broken @svarlamov

svarlamov:
Yup your change is much better. Reason my change looked weird is I was trying to keep performance log default true in the case that debug logging is true, unless performance log is explicitly disabled (=0). That’s how it was working initially. I’d much rather have it default to off though :)

Means we can also remove the flag from the GH workflows since setting it to zero is a no-op now.

Sent via Superhuman iOS ( ***@***.*** )

PR #202, opened 2025-11-08
Fix windows INITIAL contributions bugs

acunniffe:
- [x] @acunniffe and @svarlamov pair on final failing test for dirty tab changes
- [x] @acunniffe to remove all the loggers from debugging

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -40,6 +40,6 @@ jobs:
             ${{ runner.os }}-cargo-
 
       - name: Run tests
-        run: cargo test -- --test-threads=1
+        run: cargo test
         env:
           CARGO_INCREMENTAL: 0


acunniffe:
Hey @svarlamov - I made a bunch of changes to how we handle paths on windows. There's only one test in ai_tab I can't get to pass. It seems like it's all being processed properly and the checkpoints are correct. 

Few questions: 
What's supposed to happen to a dirty edit when you stage all and commit, but the file isn't saved? If it's in the IDE buffer it feels weird to stage initial contributions. I kinda of feel like on the final human checkpoint (`is_pre_commit == true`) we need to fall back to what's on disk / in the index. Sure you'd lose some AI tab attributions, but I think persisting those beyond the commit boundary is challenging and will affect all the staged/unstaged checks we do

What is the `test_ai_tab_e2e_handles_dirty_files_map` testing? It looks like you write the AI contents to disk right before you run the checkpoint? So it's not really dirty? Or are you testing the human file's dirty state doesn't interfere. 

Bit confused so if it's all right can I leave this to you to finish or pair? 

```
files for mock_ai: ["README.md"]
pathspec_filter: Some(["src/lib.rs"])
files: ["src/lib.rs", "README.md"]
human Test User changed 1 of the 2 file(s) that have changed since the last commit (1 already checkpointed)
Checkpoint completed in 1.416725417s

git_ai checkpoint output: checkpoint_kind: AiTab
filenames for mock_ai: {"src/lib.rs", "README.md"}
files for mock_ai: ["src/lib.rs", "README.md"]
pathspec_filter: Some(["src/lib.rs"])
files: ["README.md", "src/lib.rs"]
ai_tab github-copilot-tab changed 1 of the 2 file(s) that have changed since the last commit (1 already checkpointed)
Checkpoint completed in 927.586458ms

Checkpoints before commit: 2
Checkpoint 0: kind=Human, entries=1
  File: README.md, attributions=2
Checkpoint 1: kind=AiTab, entries=1
  File: src/lib.rs, attributions=2
```

svarlamov:
> Hey @svarlamov - I made a bunch of changes to how we handle paths on windows. There's only one test in ai_tab I can't get to pass. It seems like it's all being processed properly and the checkpoints are correct.
> 
> Few questions: What's supposed to happen to a dirty edit when you stage all and commit, but the file isn't saved? If it's in the IDE buffer it feels weird to stage initial contributions. I kinda of feel like on the final human checkpoint (`is_pre_commit == true`) we need to fall back to what's on disk / in the index. Sure you'd lose some AI tab attributions, but I think persisting those beyond the commit boundary is challenging and will affect all the staged/unstaged checks we do

Yup, in the current design it gets lost unless it's saved before the commit is run.

> What is the `test_ai_tab_e2e_handles_dirty_files_map` testing? It looks like you write the AI contents to disk right before you run the checkpoint? So it's not really dirty? Or are you testing the human file's dirty state doesn't interfere.

Since ai tab completions aren't saved to disk when they are accepted and the file may be dirty prior to acceptance as well, it's testing that the entire ai tab completions flow uses the contents from the dirty files. Finally, pre commit, the file is "saved" with the AI tab completion changes and commit is run. It's expected that the human pre commit checkpoint will retain the AI tab attributions since the files weren't changed since the last time that they were seen as dirty files.

PR #210, opened 2025-11-11
Print git-ai version when installing (for CI debugging).

mm-zacharydavison:
While debugging the GitHub CI squash workflow, I found it difficult to know which version of `git-ai` a workflow was using.

This PR updates the `install.sh` script so that on install, the version installed is printed.


acunniffe:
@mm-zacharydavison one more thing to know. We also started saving Git AI version to Authorship logs https://github.com/acunniffe/git-ai/pull/197

PR #225, opened 2025-11-19
Add enable/disable commands for git-ai wrapper

AtnesNess:
## Summary

This PR adds `git-ai enable` and `git-ai disable` commands to provide a mechanism to disable the git-ai wrapper for git operations in case git-ai blocks execution or creates other problems. This allows users to bypass git-ai and use the underlying git binary directly when needed.

## Problem

When git-ai is installed as a wrapper (the recommended enterprise deployment method), it intercepts all `git` commands. In some cases, git-ai may block execution or create problems that prevent normal git operations. Users need a reliable way to disable the wrapper to bypass git-ai and use the underlying git binary directly.

## Solution

This PR adds two new commands:

1. **`git-ai disable`**: Disables the git-ai wrapper by renaming the git binary wrapper, allowing all `git` commands to bypass git-ai and execute the underlying git binary directly
2. **`git-ai enable`**: Re-enables the git-ai wrapper by restoring the git binary wrapper

The implementation works by:
- When disabled: The git wrapper binary is renamed to `git.disabled`, so `git` commands no longer route through git-ai
- When enabled: The `git.disabled` binary is renamed back to `git`, restoring git-ai interception

The commands include user-friendly error messages and help text that clearly explain their purpose without exposing implementation details.

## Changes

- Added `enable` and `disable` command handlers in `src/commands/git_ai_handlers.rs`
- Added `enable_git_ai()` and `disable_git_ai()` utility functions in `src/utils.rs`
- Added `is_git_ai_disabled()` check function to detect wrapper state
- Updated help text to document the new commands
- Added logic to allow enable/disable commands to work even when git-ai is disabled
- Added user-friendly error messages that don't expose implementation details

## Usage

Users can now disable git-ai when needed:

```bash
# Disable git-ai wrapper (bypasses git-ai, uses underlying git directly)
git-ai disable

# Re-enable git-ai wrapper
git-ai enable
```

When disabled, all `git` commands will bypass git-ai and execute the underlying git binary directly, allowing users to continue working even if git-ai is experiencing issues.

## Testing

- Verified that help text displays correctly
- Verified that error messages are clear and user-friendly
- The underlying functionality remains unchanged (still uses file renaming mechanism)

### DIFF
--- a/src/commands/git_ai_handlers.rs
+++ b/src/commands/git_ai_handlers.rs
@@ -157,8 +157,8 @@ fn print_help() {
     eprintln!("  git-path           Print the path to the underlying git executable");
     eprintln!("  upgrade            Check for updates and install if available");
     eprintln!("    --force               Reinstall latest version even if already up to date");
-    eprintln!("  enable             Enable git-ai (renames git.disabled back to git)");
-    eprintln!("  disable            Disable git-ai (renames git to git.disabled)");
+    eprintln!("  enable             Enable proxying of git commands to git-ai");
+    eprintln!("  disable            Disable proxying of git commands to git-ai");
     eprintln!("  version, -v, --version     Print the git-ai version");
     eprintln!("  help, -h, --help           Show this help message");
     eprintln!("");
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -152,7 +152,7 @@ pub fn enable_git_ai() -> Result<(), GitAiError> {
     
     if !git_disabled.exists() {
         return Err(GitAiError::Generic(
-            "git-ai is already enabled (git.disabled not found)".to_string(),
+            "git-ai is already enabled".to_string(),
         ));
     }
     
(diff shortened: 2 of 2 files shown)


acunniffe:
We're not going to implement functionality like this in the core project. I don't think it's any more discoverable than telling someone to rewrite their shell profile or invoking `git-og` via a symlink and it'll create some weird issues if users don't restart all their shells after running them. We suggest enterprise install scripts add a symlink like the oss profile does ie `git-og` https://github.com/acunniffe/git-ai/blob/9d2d2e30a7e5d134adbf009e476246d2dccfbc34/install.sh#L183

This still works because
a) in both approaches user must know about git-ai and have tribal knowledge about why they might want to bypass it
b) reliable...`git-og` doesn't wrap without hooks, it skips right to git (just like this approach)
c) `git-og` is favorable because you don't need to turn it off and on (2 commands sandwiching your work). It's more explicit too. HAVE git do THIS.

PR #229, opened 2025-11-22
Single push when pushing

acunniffe:
Some organizations require signing commits. Usually this means the user is prompted to provide a passcode, fingerprint or other factor when they push. Because Git AI was firing off a second push (just notes) in parallel they were getting asked twice. 

We've fixed this by appending the AI notes refspec to user's push commands. This code + tests cover all the cases we could think of and has reasonable guards in place. 

Screenshot 2025-11-22 at 8 16 58 AM

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -45,7 +45,7 @@ jobs:
           CARGO_INCREMENTAL: 0
 
   test-ignored:
-    name: Test ignored tests on Ubuntu
+    name: Test SCM e2e tests on just Ubuntu
     runs-on: ubuntu-latest
 
     steps:


svarlamov:
Screenshot 2025-11-22 at 12 37 17 PM

Haven't gone too deep, but it seems to cause git push to fail on any repo that doesn't have AI notes yet

acunniffe:
> Screenshot 2025-11-22 at 12 37 17 PM
> Haven't gone too deep, but it seems to cause git push to fail on any repo that doesn't have AI notes yet

Great catch. I'll check if it exists in the sync that runs in pre_push_command and add + if needed

acunniffe:
Should be working now 
Screenshot 2025-11-22 at 1 55 40 PM

PR #233, opened 2025-11-23
skip loading copilot chat session data in codespaces and devcontainers

svarlamov:


acunniffe:
Why are we skipping? does it not exist in these contexts?

svarlamov:
> Why are we skipping? does it not exist in these contexts?

Correct. Have some ideas for how to pull it through, but it'll take quite a bit of work so pushing this now so that at least authorship tracking will work for now

PR #271, opened 2025-12-07
Internal calls to blame and rewrite ops no longer fail when in subdirs

acunniffe:
Most of the Git AI repository helpers assume they are running in the root of the repository. Noticed some weird behavior around authorship rewrite codepaths in subdirectories so opened this PR to locks in that assumption. 

Note: This does not change any behavior when we wrap and pass arguments to git

### DIFF
--- a/src/authorship/range_authorship.rs
+++ b/src/authorship/range_authorship.rs
@@ -4,6 +4,7 @@ use std::collections::HashSet;
 use serde::Deserialize;
 use serde::Serialize;
 
+use crate::authorship::rebase_authorship::filter_pathspecs_to_ai_touched_files;
 use crate::authorship::stats::{CommitStats, stats_for_commit_stats, stats_from_authorship_log};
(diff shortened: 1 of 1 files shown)


src/git/repository.rs
```diff
@@ -1799,6 +1799,16 @@ pub fn find_repository(global_args: &Vec<String>) -> Result<Repository, GitAiErr
...
 
+    // Rewrite global_args if -C path doesn't match the actual workdir/
+    // So every git command can assume it's being run in repo root
+    let mut global_args = global_args.clone();
+    if global_args.len() == 2 && global_args[0] == "-C" {
+        let workdir_str = workdir.display().to_string();
+        if global_args[1] != workdir_str {
+            global_args[1] = workdir_str;
+        }
+    }
```
acunniffe, line 1810: this doesn't change invocation arguments. It's only affecting the behavior of our execs to git and hardening the assumption they're running in root

PR #298, opened 2025-12-19
Disable prompt storage by default

acunniffe:
Recently a few teams have shared that their secrets / PII scanners have been going off because their developers have been putting data into prompts that shouldn't be shared with the Coding Agents / LLMs / Git AI. 

All the Git AI features that require raw prompts are experimental and limited in their usage, so we think it makes sense to disable prompt sharing by default and make it easy to turn back on per-repo. 

This PR

- [x] Toggles off prompt saving by default
- [x] introduces `git-ai config` command to make it easier to update `config.json`
- [x] Toggle off prompt saving in repos without remotes unless user has `*` for allowed repos
- [x] Document `git-ai config` command in CLI reference

### DIFF
--- a/docs/reference.mdx
+++ b/docs/reference.mdx
@@ -236,6 +236,87 @@ Automatically configure Claude Code, Cursor and GitHub Copilot to send authorshi
 git-ai install-hooks
 ```
 
+##### `config`
+
+View and manage git-ai configuration settings.
+
+```bash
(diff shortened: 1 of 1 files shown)


svarlamov:
Awesome work! Tested on my end and everything worked great -- just added two todos that we discussed in the PR description

PR #302, opened 2025-12-22
Retain working log authorship during a fast-forward pull

acunniffe:
Fixed https://github.com/acunniffe/git-ai/issues/301


acunniffe:
@svarlamov these are a bit tough to test. I'm confident in the above but want to get a method for deterministically testing ops that include remotes. 

How do you feel about this direction? Won't test network/performance but the core behaviors seem tractable https://chatgpt.com/share/6949590b-8ef4-8011-83b1-04b77ae4e587

svarlamov:
Sounds good! I think this is also how git itself tests with remotes

Sent via Superhuman ( ***@***.*** )

PR #316, opened 2026-01-03
inline when git plugin

acunniffe:
Added model name + hover to the after for AI lines 

Screenshot 2026-01-03 at 10 13 07 AM


agent-support/vscode/src/blame-lens-manager.ts
```diff
@@ -890,6 +901,78 @@ export class BlameLensManager {
     });
   }
 
+  /**
+   * Update the after-text decoration for the current cursor line.
+   * Shows "[View $MODEL Thread]" with hover content for AI-authored lines.
+   */
+  private updateAfterTextDecoration(
```
wangkaizhuang: 牛逼，这样看方便多了

PR #335, opened 2026-01-08
gitlab token priority fix

acunniffe:
Prefer user-provided gl token


acunniffe:
Tested on my personal GL org, works

Issue #249, opened 2025-11-30
Store git-ai version in checkpoints


svarlamov:

svarlamov:
Thanks @Krishnachaitanyakc! Closing

PR #348, opened 2026-01-14
Store git-ai version in checkpoints

Krishnachaitanyakc:
Fix for #249 

Add git_ai_version field to the Checkpoint struct to track which version of git-ai created each checkpoint. This is useful for:
- Debugging issues with specific versions
- Understanding feature availability when checkpoints were created
- Supporting potential migration logic for future format changes

The field uses serde attributes for backward compatibility with older checkpoints and avoids serializing when None to keep JSON compact.


acunniffe:
Looks good @Krishnachaitanyakc. simple + useful. Officially a contributor now🏆 ! 

There's a new OSS release coming in the next day or two and this will be included.

PR #352, opened 2026-01-15
Add Nix flake packaging with modules and documentation

jwiegley:
## Summary

This PR adds comprehensive Nix flake support for installing and managing git-ai on NixOS, nix-darwin, and Home Manager systems.

### Features

**Nix Flake (`flake.nix`)**
- `rustPlatform.buildRustPackage` build for the git-ai binary
- Wrapped version with automatic `~/.git-ai/config.json` initialization
- `git` wrapper script using `exec -a git` to preserve argv[0] for proper passthrough detection
- `git-og` bypass script for direct access to real git
- Development shell with full Rust toolchain
- NixOS module with system-wide activation scripts
- Home Manager module for user-level hooks and config management
- Nixpkgs overlay for easy importing

**Documentation (`README-nix.md`)**
- Quick start commands for trying and installing git-ai
- Complete flake outputs reference
- Installation methods: Home Manager (recommended), nix-darwin, NixOS system-wide, direct package, overlay
- Module options tables for both NixOS and Home Manager modules
- Development shell instructions

### Bug Fixes

**argv[0] preservation for git passthrough**
- When git-ai is invoked as `git`, it checks argv[0] to determine whether to pass through to real git or handle git-ai commands
- The Nix wrapper now uses `exec -a git` to explicitly set argv[0], ensuring proper passthrough behavior
- Without this fix, `git status` would incorrectly show "Unknown git-ai command: status"

**Extension detection without CLI in PATH**
- During Home Manager/nix-darwin activation, PATH doesn't include directories like `/opt/homebrew/bin` where editor CLIs live
- Extension detection now checks `~/.cursor/extensions/` and `~/.vscode/extensions/` directories directly
- This eliminates false warnings when extensions are already installed but the CLI isn't in PATH

### Screenshot

This is what it looks like when I switch to my new system configuration with `git-ai` hooked in:

image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -901,7 +901,7 @@ dependencies = [
 
 [[package]]
 name = "git-ai"
-version = "1.0.35"
+version = "1.0.36"
 dependencies = [
  "assert_cmd",
  "chrono",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "git-ai"
-version = "1.0.35"
+version = "1.0.36"
 edition = "2024"
 
 [dependencies]


acunniffe:
@jwiegley thanks for the first contribution. Left a comment or two but overall looks good. 

One unknown to me re: Nix conventions. We have some side effects here (install-hooks) that are currently not easy to rollback. Is it considered bad not to cleanup every side effect when you turn one of these profiles off? We're working on uninstall hooks re: https://github.com/acunniffe/git-ai/issues/349 but isn't out yet

jwiegley:
Yes, better than an install hook function that modifies a config file, would be to declare the entire config file in Nix itself and have it then replace the config file with a symlink into the Nix store where the "true" version would live. This permits versioning and rollbacks across the whole system. It is expected, when running `nix rollback`, that you go back to the "last known good" state of the system, in every respect.

The downside, of course, is that this approach would wipe out any existing user settings, requiring the user to express those settings in Nix as well if they really wanted to keep them.

tylergets:
@jwiegley was there ever code in this pr? were interested in just the rust build section of this

jwiegley:
I'm not sure what happened here. Somehow the main branch got reset to just being the main branch of Git-AI. I'll need to re-create this PR using the new Nix support branch

jwiegley:
Recreated in https://github.com/git-ai-project/git-ai/pull/450

PR #365, opened 2026-01-17
fix: filter out non-existent commits in grep_ai_notes for shallow clones

satyamtg:
Fixes orphaned git notes issue where git-ai stats would fail or return 0 AI lines when some commits referenced in git notes don't exist locally.

This happens in:
- Shallow clones (--depth=N)
- Partial clones missing commits from other branches
- Repos after force-push/rebase where old commits are gone

The fix:
1. Use git cat-file -e to verify each commit exists before sorting
2. Filter out non-existent commits from the list
3. Only pass existing commits to git log --no-walk
4. Handle edge cases (no commits exist, only one exists)

---

<a href="https://app.devin.ai/review/git-ai-project/git-ai/pull/365" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/src/git/refs.rs
+++ b/src/git/refs.rs
@@ -325,12 +325,51 @@ pub fn grep_ai_notes(repo: &Repository, pattern: &str) -> Result<Vec<String>, Gi
     // If we have multiple results, sort by commit date (newest first)
     if shas.len() > 1 {
         let sha_vec: Vec<String> = shas.into_iter().collect();
+        
+        eprintln!("[DEBUG] Found {} commits with pattern '{}' in git notes", sha_vec.len(), pattern);
+        for sha in &sha_vec {
+            eprintln!("[DEBUG]   - {}", sha);
+        }
+        
+        // FILTER OUT NON-EXISTENT COMMITS using git cat-file -e
+        let existing_shas: Vec<String> = sha_vec
+            .into_iter()
+            .filter(|sha| {
+                let mut args = repo.global_args_for_exec();
+                args.push("cat-file".to_string());
+                args.push("-e".to_string());
+                args.push(sha.clone());
+                
+                let exists = exec_git(&args).is_ok();
+                if !exists {
+                    eprintln!("[DEBUG] Filtering out non-existent commit: {}", sha);
+                }
+                exists
+            })
+            .collect();
+        
+        eprintln!("[DEBUG] After filtering: {} existing commits", existing_shas.len());
+        for sha in &existing_shas {
+            eprintln!("[DEBUG]   - {}", sha);
+        }
+        
+        if existing_shas.is_empty() {
+            return Err(GitAiError::Generic(
+                "No existing commits found in git notes for pattern".to_string()
+            ));
+        }
+        
+        if existing_shas.len() == 1 {
+            return Ok(existing_shas);
+        }
+        
+        // Sort only existing commits by date
         let mut args = repo.global_args_for_exec();
         args.push("log".to_string());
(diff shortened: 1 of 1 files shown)


satyamtg:
Fixes #366

svarlamov:
Hey @satyamtg first of all, thank you so much for contributing to Git AI! Always makes my day to see issues followed by contributions!

Can you help me understand a bit clearer how all of this is supposed to work and how exactly it addresses the issue in #366 

Here's my mental model: since git-ai stats is based on git log, if you don't have the commit (e.g., a shallow clone), it would never try to fetch the authorship note for a commit that doesn't exist. Also, if you were to request stats for a commit that doesn't exist, the command will error out -- I don't think there's any other way to do this since Git doesn't know the commit exists on remote, it just knows that it doesn't have it. What's your thinking on this?

I think we should update our limitations page to more explicitly call this out, but we intentionally have the stats command silently 'fail' (or I should say render data on a best effort basis). This is also how many of git's commands work for shallow clones (for ex, try `git blame <file>` on a shallow clone, it doesn't warn you that it'll use only local data)

satyamtg:
Hey @svarlamov , so basically I was just trying out things with git-ai. I made 4 commits on a branch, using a conversation_id (say X) with my AI agent tool. Then I paused, and pushed the 4 commits (say A, B, C and D) to remote. After that, I basically reset the branch to the base branch state, and continuing the same conversation ID X, did 2 commits with AI changes, and a human commit. Then force pushed these 3 commits to the same branch. (Say commits P, Q, R). Once this was done, I just pulled the branch on my other computer (just that branch) + the git notes at ref = ai. I tried running `git-ai stats P..R`, but it gave 0 AI attributions. But on the original computer, it gave 30 lines of AI code. Note that enough data was present on the 2nd computer where I just wanted to fetch the AI stats originally.    

So, in theory, this seems to be a bug to me. I dug deep and found that when you run stats with a range, it tries to get the prompt from AI notes, then the commits linked to that prompt. In this case, A, B, C, D were also linked. What I feel is this is how it runs - 
1. git ai blame on files at commit C to see who authored each line
2. Discover prompt IDs embedded in the blame output (e.g., ai:prompt-abc-123)
3. Searches git notes to find which commits have that prompt ID
4. Gets data from those commits to finally track attribution.

For this, at one point it would run `git log --format=%H --date-order --no-walk A B C D P Q` 
Reason being the conversation ID didn't actually change, it had A B C and D also attached to it. 

```
git log --format=%H --date-order --no-walk 6226137a9d051752bfd1ea17364a8159efbc2c2c
fatal: bad object 6226137a9d051752bfd1ea17364a8159efbc2c2c
```

Hope this clears the issue. My fix just makes sure that this git log command doesn't fail. 

The idea, as you already stated, is to do best effort attribution. But in my case, even that didn't happen.

I've already tested this fix with my exact scenario, and it works fine.

acunniffe:
Hey @satyamtg can you possibly provide 
- the contents of the Git Notes you're seeing filtered out `  eprintln!("[DEBUG] Filtering out non-existent commit: {}", sha);` 
- or a test bash script / integration test using the `TestRepo` struct we have use everywhere? 

I think you stumbled on a real + broader set of stats issues where reusing the same prompt AFTER making a commit.

svarlamov:
Thank you for the contribution @satyamtg -- I'm going to close this for now as the codebase has evolved a lot since these changes were initially proposed and it seems like under normal circumstances this has not been coming up for users.

PR #382, opened 2026-01-20
feat: Add cursor on Linux support

dragove:
Add cursor on Linux support

I'm not familiar with rust, I tested it on my local machine and seems working

Please directly edit this pr or just commit new change ignoring my change to support Linux if my commit is not satisfied.

### DIFF
--- a/src/commands/checkpoint_agent/agent_presets.rs
+++ b/src/commands/checkpoint_agent/agent_presets.rs
@@ -916,7 +916,18 @@ impl CursorPreset {
                 .join("User"))
         }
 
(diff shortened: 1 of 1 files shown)


svarlamov:
Do you know if this covers both Cursor CLI and Desktop for linux?

dragove:
> Do you know if this covers both Cursor CLI and Desktop for linux?

Only desktop is supported. I only added the configuration path of cursor on linux. Thus the original code can find it.

svarlamov:
Sounds good -- merging!

PR #383, opened 2026-01-21
Better support for Git Clients

acunniffe:
Many devs use Git Clients, which (when being pointed at a `git` binary often run a `--version` check and look for `libexec`

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,8 @@
 <img src="https://github.com/acunniffe/git-ai/raw/main/assets/docs/git-ai.png" align="right"
      alt="Git AI by acunniffe/git-ai" width="100" height="100" />
 
+
+
 </div>
 <div>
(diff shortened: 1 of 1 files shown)


svarlamov:
```
➜  testing-git-10 git:(master) ✗ git --exec-path

thread 'main' panicked at src/git/repository.rs:1837:52:
called `Option::unwrap()` on a `None` value
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
➜  testing-git-10 git:(master) ✗ /opt/homebrew/bin/git --exec-path
/opt/homebrew/opt/git/libexec/git-core
➜  testing-git-10 git:(master) ✗
```

acunniffe:
Spoke to Sasha - move symlink logic into the install-hooks, and fix panic. 
Verify if Fork uses the exec symlinks before doing a pass-through / alias wrapper

PR #384, opened 2026-01-21
Introduce a --json flag to the status command to enable JSON output

Karn:
Output the commit stats and checkpoints as JSON via a new, `--json`, flag to enable use cases such as parsing state prior to committing.

### DIFF
--- a/src/commands/status.rs
+++ b/src/commands/status.rs
@@ -6,9 +6,11 @@ use crate::error::GitAiError;
 use crate::git::find_repository;
 use crate::git::repo_storage::InitialAttributions;
 use crate::git::repository::Repository;
+use serde::Serialize;
 use std::collections::HashSet;
 use std::time::{SystemTime, UNIX_EPOCH};
 
+#[derive(Serialize)]
 struct CheckpointInfo {
     time_ago: String,
     additions: u32,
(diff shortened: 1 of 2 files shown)


src/commands/status.rs
```diff
@@ -17,14 +19,33 @@ struct CheckpointInfo {
...
 
-pub fn handle_status(_args: &[String]) {
-    if let Err(e) = run_status() {
+#[derive(Serialize)]
+struct StatusOutput {
+    stats: CommitStats,
+    checkpoints: Vec<CheckpointInfo>,
+}
+
+pub fn handle_status(args: &[String]) {
```
Karn, line 28: I kept changes here light, I think it would be preferential to move this function into `src/commands/git_ai_handlers.rs` to match the `handle_stats` implementation, but that would introduce additional changes that feel out of scope.

Happy to make this change in a follow-up PR.

src/commands/status.rs
```diff
@@ -17,14 +19,33 @@ struct CheckpointInfo {
     is_human: bool,
 }
 
-pub fn handle_status(_args: &[String]) {
-    if let Err(e) = run_status() {
+#[derive(Serialize)]
+struct StatusOutput {
+    stats: CommitStats,
+    checkpoints: Vec<CheckpointInfo>,
+}
```
Karn, line 26: I can also replace this with a `json!` macro and remove the intermediate struct if that is more appropriate.

acunniffe: I think this it's ok to do this here since there is not already a return struct from this function.

src/commands/status.rs
acunniffe: This is ok for now. We can't use a CLI parsing library for the git wrapper, but we should be using it more for the `git-ai` commands. I'll have a Claude do some cleanup here eventually and use Clap for this sort of code. That's a bigger project.

acunniffe (approved):

acunniffe:
All set @Karn -- happy first contribution 🥳 . Welcome to the crew. 

Will ship in a release later today. I'm very curious what you plan to use this for if you can share? I imagined this more as a troubleshooting command (answers: are my agents making checkpoints?), but sure there's more one can do with the data.

PR #391, opened 2026-01-22
feat: add support for droid

dragove:
AI crafted code, working for me. Any edits are appreciated.

droid hooks reference can be found [here](https://docs.factory.ai/reference/hooks-reference)

the document shows it uses snake_case. However it actually use camelCase.

### DIFF
--- a/tests/droid.rs
+++ b/tests/droid.rs
@@ -55,7 +55,9 @@ fn test_parse_droid_jsonl_transcript() {
         match message {
             Message::User { timestamp, .. }
             | Message::Assistant { timestamp, .. }
-            | Message::ToolUse { timestamp, .. } => {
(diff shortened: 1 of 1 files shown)


svarlamov:
@dragove Thank you for your contribution! I think we're getting really close here. Any chance you could:

1. Merge the latest git-ai `main` branch back into your fork
2. Implement the install hooks flow for droid as well. We require all agent integrations to either be first-party or have an automated hooks installer setup. I would point your coding agent here: https://github.com/git-ai-project/git-ai/tree/main/src/mdm/agents

dragove:
@svarlamov 
I've rebased onto main and modify some logic to get better experience.
install-hooks logic is implemented and prevent droid import hooks from claude.
Model field is not  present from droid hooks. So model will always be `unknown`

PR #398, opened 2026-01-23
feat: Add Fish shell support in installer

### LINKED ISSUE #386: Add support for Fish shell
Fish shell is not supported on current version, but it quite popular alternative shell. Could we add support for it? Thanks

Right now it install hooks in ZSH even when Fish is used

Krishnachaitanyakc:
Fixes #386

- Enhanced detect_shell() to detect Fish shell via $SHELL env var, ~/.config/fish/config.fish existence, and $FISH_VERSION variable
- Use fish_add_path -g for PATH modification (idiomatic Fish syntax)
- Create ~/.config/fish/ directory if it doesn't exist before writing config
- Maintains backward compatibility with bash and zsh users

### DIFF
--- a/tests/e2e/installer-shell-detection.bats
+++ b/tests/e2e/installer-shell-detection.bats
@@ -1,0 +1,372 @@
+#!/usr/bin/env bats
+
+# BATS test file for install.sh shell detection functionality
+# https://github.com/bats-core/bats-core
+#
+# These tests verify that the installer correctly detects and configures
+# PATH in multiple shell configuration files.
+
+setup() {
+    # Create a temporary directory for each test
+    export TEST_TEMP_DIR="$(mktemp -d)"
+    export ORIGINAL_DIR="$(pwd)"
+    export ORIGINAL_HOME="$HOME"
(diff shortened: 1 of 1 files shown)


svarlamov:
Thank you for your contribution @Krishnachaitanyakc! I just realized that we only install in one shell, but for example, I now have both zsh and fish installed on my laptop. Do you think there's a sane way to make sure that we end up in the PATH on all the user's shells? Curious what your thoughts are on this

Krishnachaitanyakc:
@svarlamov One option I can think of is, instead of detecting which shell is primary and modifying only that one, we can detect all shells that have config files present and modify each of them. What do you think?

svarlamov:
@Krishnachaitanyakc I think that's a great idea -- can you update the PR? Excited to get this merged :)

Krishnachaitanyakc:
@svarlamov done, can you review it?

svarlamov:
Just reviewed and tested a bunch -- looks great! I'll merge as soon as the CI finishes

PR #399, opened 2026-01-24
sp: implementation, overridden, et al

tnm:
Fix some spellings of `implement*` and `overridden`, and `SHOULD`. The `_overridden` change is notable because it schema-level; note below.

### DIFF
--- a/specs/git_ai_standard_v3.0.0.md
+++ b/specs/git_ai_standard_v3.0.0.md
@@ -730,3 +730,7 @@ When amend is used during a rebase or other operation:
 - Implementations of 3.0.0 or later SHOULD NOT attempt to process earlier versions
 - Implementations > 3.0.0 MUST process earlier versions, provided they are valid and match the schema they advertise
 
+### Errata
+
+E-001: The field name overriden_lines was introduced as a typographical error in v3.0.0 and shipped in the reference git-ai implementation, where it became canonical. In v4.x, this field WILL be renamed to overridden_lines.
+


specs/git_ai_standard_v3.0.0.md
```diff
@@ -166,7 +166,7 @@ Each entry in the `prompts` object MUST contain:
 | `total_additions` | integer | REQUIRED | Total lines added by this session |
 | `total_deletions` | integer | REQUIRED | Total lines deleted by this session |
 | `accepted_lines` | integer | REQUIRED | Lines accepted in the final commit |
-| `overriden_lines` | integer | REQUIRED | Lines that were later modified by human |
+| `overridden_lines` | integer | REQUIRED | Lines that were later modified by human |
```
tnm, line 169: Wanted to flag this since it's technically a schema-level typo fix, not just a documentation change. Of course ultimately your call, but could:

- treat as erratum, or
- bump the schema version, or
- state that the misspelled field was never valid and this is an intentional break.

Erratum approach feels right especially this early. I can add it here if you'd like.

acunniffe: Good catch and right call - erratum for sure. Verified the error is in the actual source as well so we definitely don't want to ship a break here, but will resolve in next major version.

svarlamov:
Thank you so much for your contributions @tnm! Would love to have your thoughts on updates to the spec here as well: https://docs.google.com/document/d/1p5su6SHW5350HAEFEov7OvT_hLNLyCSNBPj9cZ46ptw/edit?usp=sharing

specs/git_ai_standard_v3.0.0.md
acunniffe: I added an Errata section at the bottom of the spec

PR #379, opened 2026-01-20
feat: Add multi-repository workspace support for checkpoints

Krishnachaitanyakc:
This change enables git-ai to work correctly in workspaces where the root directory is not a git repository but contains multiple independent git repositories as subdirectories. #357 

Changes:
- Add find_repository_for_file() to locate the git repo containing a file by walking up the directory tree (with submodule detection)
- Add group_files_by_repository() to group edited files by their containing git repository
- Update checkpoint handler to detect multi-repo workspaces and process each repository separately with its own subset of edited files
- Convert workspace-relative paths to absolute paths for correct repo detection
- Continue processing remaining repos if one fails (graceful degradation)
- Add detailed logging for multi-repo checkpoint progress
- Add comprehensive test suite for multi-repo workspace scenarios

This enables proper attribution tracking when editing files across multiple repositories in a single session.

PR #402, opened 2026-01-24
fixes critical bug w/ multi-repo support

acunniffe:
fixes a regression introduced by https://github.com/git-ai-project/git-ai/pull/379. Stop checkpoints from being added for files outside of a repo.


acunniffe:
@svarlamov I know you gave some offline feedback about adding the check in both places, but reviewing `post_commit` hook, I think line covers us https://github.com/git-ai-project/git-ai/blob/3084cb5c9744df7e01a50fd84da63048d6f7d5a4/src/authorship/post_commit.rs#L62

PR #407, opened 2026-01-27
Feat/git config native

### LINKED ISSUE #376: Git config native implementation
Instead of calling `git config` to get config values every time (adds 10-40ms of overhead for each call, which can _really_ add up), build or find a rust native implementation.

This has positive performance implications for git-ai overall, and enables us to build features like the one we didn't merge in #68 for git aliased commands that are currently not being handled correctly by git-ai

bernoussama:
Fixes #376 
- Replaced `git config` commands invocations with gix-config native Rust implementation
- Added `GixError` type
- Added gix_config tests, ensuring same output as `git config` commands

### DIFF
--- a/tests/gix_config_tests.rs
+++ b/tests/gix_config_tests.rs
@@ -225,8 +225,8 @@ fn test_config_falls_back_to_global() {
     let repo = TestRepo::new();
 
     // Use a unique key to avoid conflicts with real config
-    let test_key = "gitaitest.globalfallback";
-    let global_value = "GLOBAL_TEST_VALUE_12345";
+    let test_key = "gitaici.globalcheck";
+    let global_value = "GLOBAL_CI_VALUE_12345";
 
     // Set a global value for our test key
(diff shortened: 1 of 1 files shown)


bernoussama:
Hey @svarlamov ,
1. Yes it should work if not inside a repo it'll fallback to env and global config as per gix_config docs, but it should be tested, will include a test for it, good observation.
2. I will look into TestRepo and use it for testing
3. For the serial macro, I'll use TestRepo for tests and see how it goes

bernoussama:
about current git config implementation: 
- It'll work outside of a repo, as i added a fallback to global Git config file if not inside a Git repo. This scenario remains untested however, because it would involve modifying env variables requiring these tests to be serial.

bernoussama:
> LGTM! I added some tests for bare clones, etc. but it was pretty complete. Thanks!

I see what you did to test global config fallback, modifying the global config directly also works, nice.
Happy to help!

PR #413, opened 2026-01-28
chore: replace mentions of the old repo path with the new organization

EDM115:
obviously didn't replaced "acunniffe" where it still did make sense, i.e. git commit author, submodule or calendly link  
as a bonus, updated a documentation link that no longer exists to the website

I made this PR while reviewing the powershell script before installing, and then I noticed that the repo name mismatched with the actual one, probably after you moved the repo to its own orga.  
While GitHub provides automatic link redirection between the old and new name, it is still a good idea to update it ^^  
goes with git-ai-project/action#2 too


svarlamov:
Great catch and thank you for your first contribution!

PR #430, opened 2026-01-30
Login token install hook events

Ryan-Adami:


install.sh
```diff
@@ -276,7 +276,11 @@ success "You can now run 'git-ai' from your terminal"
 INSTALLED_VERSION=$(${INSTALL_DIR}/git-ai --version 2>&1 || echo "unknown")
 echo "Installed git-ai ${INSTALLED_VERSION}"
 
-# Install hooks
+# Login user with install token if provided
+if [ -n "${INSTALL_NONCE:-}" ] && [ -n "${API_BASE:-}" ]; then
+    ${INSTALL_DIR}/git-ai exchange-nonce || true
```
svarlamov: Noticed here and in the windows one -- continuing the installation isn't a bad idea, but I do think we need to at least re-prompt the user to login and show an error message if this failed

file.txt
```diff
@@ -0,0 +1,2 @@
+foobar
```
svarlamov: tidy

src/observability/flush.rs
```diff
@@ -268,24 +270,37 @@ fn cleanup_old_logs(logs_dir: &PathBuf) {
...
+            dirs.push(global_logs_dir);
         }
+    }
 
-        if !current.pop() {
-            break;
+    // Legacy: per-repo logs (.git/ai/logs) - check current repo only
+    if let Ok(mut current) = std::env::current_dir() {
+        loop {
+            let git_dir = current.join(".git");
```
svarlamov: Don't change this here (since we want to just copy the old impl for backwards-compat), but I think we have utils for getting current repo git dir. And notably, this impl would break in worktrees for ex

svarlamov: Good that we're moving to the global location for the logs

src/observability/flush.rs
```diff
@@ -268,24 +270,37 @@ fn cleanup_old_logs(logs_dir: &PathBuf) {
...
-        let git_dir = current.join(".git");
-        if git_dir.exists() && git_dir.is_dir() {
-            let logs_dir = git_dir.join("ai").join("logs");
-            if logs_dir.exists() && logs_dir.is_dir() {
-                return Some((current.clone(), logs_dir));
-            }
+    // Primary: global logs directory (~/.git-ai/internal/logs)
+    if let Some(home) = dirs::home_dir() {
+        let global_logs_dir = home.join(".git-ai").join("internal").join("logs");
+        if global_logs_dir.exists() && global_logs_dir.is_dir() {
```
svarlamov: shouldnt we auto-create the entire path if it doesn't exist? Technically not guaranteed that any of these folders will exist so we try to always make sure it's all there

src/commands/install_hooks.rs
```diff
@@ -60,8 +127,15 @@ pub fn run(args: &[String]) -> Result<HashMap<String, String>, GitAiError> {
     // Run async operations with smol and convert result
     let statuses = smol::block_on(async_run_install(&params, dry_run, verbose))?;
 
-    // Spawn background metrics flush
-    spawn_background_metrics_db_flush();
+    // Flush logs immediately so install metrics are captured right away (only if logged in)
```
svarlamov: IIRC the flush function should take care of the login state

src/commands/install_hooks.rs
```diff
@@ -41,6 +104,10 @@ pub fn to_hashmap(statuses: HashMap<String, InstallStatus>) -> HashMap<String, S
 
 /// Main entry point for install-hooks command
 pub fn run(args: &[String]) -> Result<HashMap<String, String>, GitAiError> {
+    // Set up global log context (for logging outside git repos)
+    // Logs go to ~/.git-ai/internal/logs/{PID}.log
+    set_global_log_context();
```
svarlamov: Is this the original pattern? I don't think we should have to call a fn to initialize log ctx (it should either happen on startup (not per-command) or just in the log logic directly

src/commands/exchange_nonce.rs
```diff
@@ -0,0 +1,129 @@
...
+        .map_err(|e| format!("\x1b[33mWarning: Failed to store credentials: {}\x1b[0m", e))?;
+
+    eprintln!("\x1b[32mSuccessfully logged in\x1b[0m");
+    Ok(())
+}
+
+fn format_error(_prefix: &str, _detail: String, install_page_url: Option<&str>) -> String {
+    if let Some(url) = install_page_url {
+        format!(
+            "\x1b[33mNonce exchange failed. Get a new install command from:\n  {}\x1b[0m",
```
svarlamov: Can we be a bit more human-friendly with the msg?

src/authorship/post_commit.rs
```diff
@@ -412,7 +412,6 @@ fn record_commit_metrics(
 ) {
     use crate::metrics::{record, CommittedValues, EventAttributes};
 
-    // Build parallel arrays: index 0 = "all" (aggregate), index 1+ = per tool/model
```
svarlamov, line 415: can we leave it?

src/commands/exchange_nonce.rs
```diff
@@ -0,0 +1,129 @@
...
+use crate::api::client::ApiContext;
+use crate::auth::types::StoredCredentials;
+use crate::auth::CredentialStore;
+
+use serde::Deserialize;
+use std::time::{SystemTime, UNIX_EPOCH};
+
+/// Token response from the OAuth endpoint
+#[derive(Debug, Deserialize)]
+struct TokenResponse {
```
svarlamov: don't we already define these models in the existing oauth client?

src/commands/exchange_nonce.rs
```diff
@@ -0,0 +1,129 @@
...
+        eprintln!("{}", e);
+        // Don't exit with error - install should continue even if login fails
+    }
+}
+
+fn exchange_nonce(nonce: &str, api_base: &str, install_page_url: Option<&str>) -> Result<(), String> {
+    eprintln!("Exchanging install nonce for credentials...");
+
+    // Build the token request
+    let url = format!("{}/worker/oauth/token", api_base.trim_end_matches('/'));
```
svarlamov: i haven't reviewed the entire auth flow yet, but isn't this the same endpoint we use for normal login? If so can't we just reuse that code? I'm just really weary of adding more impls of almost the same code (I know the agents _really_ struggle with reuse)

svarlamov (changes requested):
Awesome work! Congrats on your first Git AI PR! Added a ton of thoughts but mostly small stuff/code reuse. I think the last big thing is just renaming 'nonce' to install login token or whatever makes more sense

PR #462, opened 2026-02-05
feat: add ignore patterns support for stats command

### LINKED ISSUE #461: # Feature Request: Add ignore patterns support for `stats` command
# Feature Request: Add ignore patterns support for `stats` command

## Problem
Currently, `git-ai stats` includes all files when calculating AI vs human code statistics, including:
- Lock files (package-lock.json, Cargo.lock, go.sum, etc.)
- Minified/bundled files (*.min.js, *.bundle.js)
- Generated code (protobuf files, GraphQL generated files, etc.)
- Build artifacts that are committed to repos

This makes statistics inaccurate because these files are auto-generated and shouldn't count towards "real coding work".

## Proposed Solution

Add support for ignoring file patterns in stats calculations, with three configuration levels:

### 1. CLI flag (highest priority)
```bash
git-ai stats --ignore "*.lock" "*.min.js" "dist/**"
git-ai stats HEAD --ignore "*.generated.*"
```

### 2. Project-level config (team shared)
Create `.git-ai-ignore` in repository root (gitignore syntax):
```
# Lock files
*.lock
go.sum

# Build artifacts  
*.min.js
*.bundle.js

# Generated code
*.generated.*
*_pb.go
```

### 3. Global config (user defaults)
In `~/.git-ai/config.json`:
```json
{
  "stats_ignore_patterns": [
    "*.lock",
    "*.min.js",
    "*.generated.*"
  ]
}
```

All three levels merge together (additive).

## Use Cases

**Monorepo with multiple languages**: Each team can define their own `.git-ai-ignore` committed to the repo.

**Personal preferences**: Users can set global defaults without modifying each project.

**Ad-hoc filtering**: Quick one-off stats with `--ignore` flag.

## Implementation Notes

The ignore logic should:
- Only affect `git-ai stats` command (not checkpoint, blame, diff)
- Use glob pattern matching (like .gitignore)
- Skip entire files from statistics (both AI and human lines)
- Work with both `stats <commit>` and `stats <range>` modes

## Benefits

✅ More accurate AI/human code ratio  
✅ Exclude noise from lock files and generated code  
✅ Team consistency via committed `.git-ai-ignore`  
✅ Flexible: global defaults + per-project overrides  

## Example

Before:
```bash
$ git-ai stats HEAD
you  ████████░░░░░░░░░░░░ ai
     65%              35%
# Includes 5000 lines from package-lock.json
```

After:
```bash
$ git-ai stats HEAD
you  ████████████████░░░░ ai
     80%              20%  
# package-lock.json excluded via .git-ai-ignore
```

Would this be a useful addition to git-ai?

txf0096:
Fixes #461

## Summary

This PR implements ignore patterns support for the `stats` command, allowing users to exclude auto-generated files (lock files, minified code, generated code) from AI/human statistics.

## Implementation

Added three configuration levels (all additive):

### 1. CLI flag
```bash
git-ai stats --ignore "*.lock" "*.min.js" "dist/**"
git-ai stats HEAD --ignore "*.generated.*"
```

### 2. Project-level config
Create `.git-ai-ignore` in repository root (gitignore syntax):
```
# Lock files
*.lock
go.sum

# Build artifacts  
*.min.js
*.bundle.js

# Generated code
*.generated.*
*_pb.go
```

### 3. Global config
In `~/.git-ai/config.json`:
```json
{
  "stats_ignore_patterns": [
    "*.lock",
    "*.min.js",
    "*.generated.*"
  ]
}
```

## Changes

- Add `src/authorship/ignore_patterns.rs` module for loading patterns from files
- Add `stats_ignore_patterns` field to `Config` struct
- Merge patterns in `handle_stats()` before calling stats/range functions
- Only affects `stats` command (not checkpoint, blame, diff)
- Uses glob pattern matching (like .gitignore)
- Skips entire files from statistics (both AI and human lines)

## Benefits

✅ More accurate AI/human code ratio  
✅ Exclude noise from lock files and generated code  
✅ Team consistency via committed `.git-ai-ignore`  
✅ Flexible: global defaults + per-project overrides  

## Example

Before:
```bash
$ git-ai stats HEAD
you  ████████░░░░░░░░░░░░ ai
     65%              35%
# Includes 5000 lines from package-lock.json
```

After:
```bash
$ git-ai stats HEAD
you  ████████████████░░░░ ai
     80%              20%  
# package-lock.json excluded via .git-ai-ignore
```

## Testing

- Compiled successfully with `cargo build`
- Pattern loading logic tested with multiple sources
- Backward compatible (defaults to no patterns if not configured)

---

<a href="https://app.devin.ai/review/git-ai-project/git-ai/pull/462" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/src/authorship/ignore_patterns.rs
+++ b/src/authorship/ignore_patterns.rs
@@ -1,0 +1,123 @@
+use std::fs;
+use std::path::Path;
+
+use crate::config::Config;
+use crate::git::repository::Repository;
+use crate::utils::debug_log;
+
+/// Load ignore patterns from various sources and merge with CLI args
+/// Priority order: global config > project file > CLI args (all are additive)
+///
+/// Sources:
+/// 1. Global config: ~/.git-ai/config.json (stats_ignore_patterns field)
+/// 2. Project file: .git-ai-ignore in repo root
+/// 3. CLI args: --ignore patterns passed by user
+pub fn load_ignore_patterns_from_files(repo: &Repository) -> Vec<String> {
+    let mut patterns = Vec::new();
+
+    // 1. Load from global config (~/.git-ai/config.json)
+    let global_config_patterns = Config::get().stats_ignore_patterns();
+    if !global_config_patterns.is_empty() {
+        debug_log(&format!(
+            "Loaded {} patterns from ~/.git-ai/config.json",
+            global_config_patterns.len()
+        ));
+        patterns.extend_from_slice(global_config_patterns);
+    }
+
+    // 2. Load project ignore (.git-ai-ignore in workdir root)
+    if let Some(project_patterns) = load_project_ignore(repo) {
+        debug_log(&format!(
+            "Loaded {} patterns from .git-ai-ignore",
+            project_patterns.len()
+        ));
+        patterns.extend(project_patterns);
+    }
+
+    patterns
+}
+
+/// Load project ignore from .git-ai-ignore in workdir root
+fn load_project_ignore(repo: &Repository) -> Option<Vec<String>> {
+    let workdir = repo.workdir().ok()?;
+    let path = workdir.join(".git-ai-ignore");
+    read_ignore_file(&path)
+}
+
+/// Read ignore file and parse patterns
+/// Supports comments (#) and blank lines
+fn read_ignore_file(path: &Path) -> Option<Vec<String>> {
+    if !path.exists() {
+        return None;
+    }
+
+    let content = match fs::read_to_string(path) {
+        Ok(content) => content,
+        Err(e) => {
+            debug_log(&format!("Failed to read ignore file {:?}: {}", path, e));
+            return None;
+        }
+    };
+
+    let patterns: Vec<String> = content
+        .lines()
+        .map(|line| line.trim())
+        .filter(|line| !line.is_empty() && !line.starts_with('#'))
+        .map(|line| line.to_string())
+        .collect();
+
+    if patterns.is_empty() {
+        None
+    } else {
+        Some(patterns)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_read_ignore_file_with_comments() {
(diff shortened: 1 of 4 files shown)


svarlamov:
Thank you for the contribution! I like the direction, but wondering would you be open to narrowing the scope to only CLI flags?

I don't think we're quite ready to decide on ignore files and config changes just yet. Would that solve the issue for you still?

svarlamov (changes requested):
Scope here's a bit too large for us right now. Can we make this PR just include passing CLI args for stats rather than config and pre-repo ignore files too?

txf0096:
> Scope here's a bit too large for us right now. Can we make this PR just include passing CLI args for stats rather than config and pre-repo ignore files too?

CLI flag support for ignore patterns already exists in git-ai (developers can pass patterns via command-line arguments). However, in practice, we've found this approach doesn't scale well for our use case:

  The Problem:
  - Many projects need to ignore 10-20+ patterns (lock files, generated code, vendor dirs, minified assets, etc.)
  - Requiring developers to type/remember: git-ai stats --ignore "*.lock" "*.min.js" "vendor/**" "dist/**" "*.generated.ts" ... every time is impractical
  - This leads to inconsistent statistics across team members (some remember patterns, others don't)
  - Statistics become unreliable and difficult to compare

  Why Config Files Matter:
  - .git-ai-ignore (project-level): Ensures consistent ignore rules across all team members
  - ~/.git-ai/config.json (global): Allows users to set personal defaults (e.g., always ignore *.lock)

  I understand starting with just CLI flags keeps the scope smaller. However, without config file support, this feature won't actually solve our real-world problem - developers won't use CLI flags consistently enough to make statistics reliable.

svarlamov:
Hey @txf0096! We just release support for setting up your ignores using the same format that GitHub respects for generated files: https://docs.github.com/en/repositories/working-with-files/managing-files/customizing-how-changed-files-appear-on-github

Ref #516

Issue #509, opened 2026-02-12
Pre-built binaries require GLIBC 2.29+, incompatible with older Linux systems


Persephones-hades101:
# Issue: Pre-built binaries require GLIBC 2.29+, incompatible with older Linux systems

## Environment
- **OS**: CentOS 8 / RHEL 8 (GLIBC 2.28)
- **git-ai version**: v1.1.2 (latest pre-built binary)
- **Installation method**: `curl -sSL https://usegitai.com/install.sh | bash`

## Problem
The pre-built Linux binaries fail to run on systems with GLIBC 2.28 or earlier:

```bash
$ git-ai --version
/users/sudhasi2/.git-ai/bin/git-ai: /lib64/libm.so.6: version `GLIBC_2.29' not found (required by /users/sudhasi2/.git-ai/bin/git-ai)
/users/sudhasi2/.git-ai/bin/git-ai: /lib64/libc.so.6: version `GLIBC_2.29' not found (required by /users/sudhasi2/.git-ai/bin/git-ai)
/users/sudhasi2/.git-ai/bin/git-ai: /lib64/libc.so.6: version `GLIBC_2.30' not found (required by /users/sudhasi2/.git-ai/bin/git-ai)
```

Installation completes successfully but the binary is unusable.

## Impact
This affects users on:
- CentOS 8 / RHEL 8 (GLIBC 2.28)
- Debian 10 (GLIBC 2.28)
- Ubuntu 18.04 LTS (GLIBC 2.27)
- Other enterprise Linux systems with older GLIBC versions

These systems are still widely deployed in corporate environments.

## Current Workaround
Building from source works fine:
```bash
# Install Rust
export CARGO_HOME=/workspace/rust/cargo
export RUSTUP_HOME=/workspace/rust/rustup
curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs/ | sh -s -- -y --no-modify-path

# Build git-ai
git clone https://github.com/git-ai-project/git-ai.git
cd git-ai
cargo build --release
cp target/release/git-ai ~/.git-ai/bin/git-ai
```

The locally-built binary works perfectly with GLIBC 2.28.

### Option 1: Build on older system (Recommended)
Build pre-built binaries on a system with GLIBC 2.27 or 2.28. Binaries built against older GLIBC versions work on newer systems (forward compatibility).

### Option 2: Provide multiple binary versions
Offer separate downloads for different GLIBC versions:
[...]

PR #513, opened 2026-02-12
fix: use musl static linking to eliminate GLIBC dependency for Linux binaries

svarlamov:
# fix: use musl static linking to eliminate GLIBC dependency for Linux binaries

## Summary

Addresses #509 — pre-built Linux binaries currently require GLIBC 2.29+ (built on `ubuntu:20.04` / GLIBC 2.31), which breaks on CentOS 8, RHEL 8, Debian 10, and other older Linux systems.

Instead of lowering the GLIBC floor (which would require an EOL build image), this PR switches Linux release builds to **musl static linking**, producing fully self-contained binaries with **zero GLIBC dependency**.

**Changes:**
1. **`release.yml`**: Switch Linux build targets from `x86_64-unknown-linux-gnu` / `aarch64-unknown-linux-gnu` to their `-musl` equivalents. Docker build image changed from `ubuntu:20.04` to `ubuntu:22.04` (maintained). Install `musl-tools` instead of `libssl-dev`.
2. **`release.yml`**: Replace the old GLIBC version logging with a static-linking assertion — the release build will now fail if `ldd` reports the binary as dynamically linked.
3. **`install.sh`**: No functional changes (net diff is a cosmetic blank line).

Artifact names (`git-ai-linux-x64`, `git-ai-linux-arm64`) are unchanged — this is a drop-in replacement from the user's perspective.

- [ ] **musl build has NOT been tested end-to-end**: PR CI does not exercise the release workflow (`workflow_dispatch` only). A dry-run release build (`workflow_dispatch` with `dry_run: true`) is **strongly recommended** before merging to confirm the musl build succeeds for both x64 and arm64.
- [ ] **`rusqlite` bundled SQLite compilation with musl**: `rusqlite` compiles SQLite from C source using `cc`. Verify `musl-gcc` (from `musl-tools`) correctly compiles this on both x64 and arm64. This is the most likely build failure point.
- [ ] **Static linking verification pattern**: The `ldd` output check uses `grep -q 'not a dynamic executable\|statically linked'`. Confirm this matches the actual `ldd` output for musl binaries on the Ubuntu 22.04 runner (the exact message can vary by `ldd` version).
- [ ] **Binary size and runtime behavior**: musl static binaries are typically larger and use musl's allocator instead of glibc's. For a CLI tool this should be negligible, but worth a quick sanity check (run the binary, check `--version`, try a commit).

**Recommended test plan:**
1. Trigger a dry-run release build to confirm both x64 and arm64 musl builds succeed
2. Download the resulting binaries and verify with `file <binary>` (should say "statically linked") and `ldd <binary>` (should say "not a dynamic executable")
3. Test the binary on a system with old GLIBC (e.g., CentOS 8 / Debian 10) to confirm it runs

### Notes
- Link to Devin run: https://app.devin.ai/sessions/ca466e0eab5d40a0a4f8b365e6dfd7bc
- Requested by: @svarlamov

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -29,15 +29,15 @@ jobs:
       matrix:
         include:
           - os: ubuntu-22.04
-            target: x86_64-unknown-linux-gnu
+            target: x86_64-unknown-linux-musl
             artifact_name: git-ai-linux-x64
             use_docker: true
-            docker_image: debian:10
+            docker_image: ubuntu:22.04
           - os: ubuntu-22.04-arm
-            target: aarch64-unknown-linux-gnu
+            target: aarch64-unknown-linux-musl
             artifact_name: git-ai-linux-arm64
             use_docker: true
-            docker_image: debian:10
+            docker_image: ubuntu:22.04
           - os: windows-latest
             target: x86_64-pc-windows-msvc
             artifact_name: git-ai-windows-x64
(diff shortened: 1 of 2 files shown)


svarlamov:
is there any chance that this will break the builds for newer os's/ubuntu/etc. any potential issues you can think of?

svarlamov:
is there a way to do this without using an EOL'd builder image?

SampsonKY:
I need this！！！

PR #536, opened 2026-02-17
removed serial testing library.

### LINKED ISSUE #264: Add test harness for setting env vars in test repo
Instead of the flakey serial test lib approach used in https://vscode.dev/github/acunniffe/git-ai/blob/fix/dec-3-25-cves-and-ignore-prompts-fix/tests/github_copilot.rs

- [ ] Make it easy to set arbitrary env vars on test repo
- [ ] Update the tests
- [ ] Remove the serial test lib

stuartsessions:
# PR: Make tests parallel-safe and remove `serial_test`
Fixes #264 

Summary
- Add thread-local test environment overrides and HOME override to `src/mdm/utils.rs`.
- Add per-`TestRepo` environment injection (`env_vars`, `add_env`, `set_envs`, `clear_envs`) to `tests/repos/test_repo.rs` so spawned child processes receive test-specific envs.
- Update runtime code to consult test-aware helper (`env_test_proxy`) where tests previously mutated `std::env` (notably opencode + checkpoint presets and prompt utilities).
- Replace in-test `std::env::set_var` usages in several tests with thread-local overrides (`set_test_env_override`) or per-repo env injection.
- Remove the `serial_test` dev-dependency and `#[serial]` usage from tests.
- Normalize blame output for `--contents` mode: map `Not Committed Yet` → `External file (--contents)` to stabilize tests.

Why
- Many integration tests mutated the process-global environment which caused races and flakiness when tests ran in parallel. Thread-local overrides plus per-repo env injection let tests simulate environment variables without changing global `std::env`.

Files changed (high level)
- `src/mdm/utils.rs` — added `TEST_HOME_OVERRIDE`, `TEST_ENV_OVERRIDES`, `set_test_home_override`, `set_test_env_override`, `get_test_env_override`, `env_test_proxy`.
- `src/git/test_utils/tmp_repo.rs` — new file extracting `TmpRepo` and `TmpFile` structs from mod.rs with integrated unit tests for environment variable isolation.
- `src/git/test_utils/mod.rs` — cleaned up to re-export `TmpRepo` and `TmpFile` from tmp_repo module; removed ~1300 lines of duplicated struct and method definitions.
- `Cargo.toml` — removed `serial_test` dev-dependency.
- `src/commands/checkpoint_agent/opencode_preset.rs`, `src/commands/checkpoint_agent/agent_presets.rs`, `src/authorship/prompt_utils.rs` — switched env reads to use `env_test_proxy`.
- Tests updated: `tests/github_copilot.rs`, `tests/opencode.rs`, `src/mdm/agents/codex.rs`, and others to use test overrides instead of `std::env::set_var`.

Workflow Changes
- `.github/workflows/tests.yml` - changed the testing tool to use 'cargo-nextest' for improved performance on integration tests. changed test workflow to perform unit and integration tests concurrently.

Notes & follow-ups
- **TmpRepo refactor**: Extracted `TmpRepo` and `TmpFile` structs into their own file (`src/git/test_utils/tmp_repo.rs`) to improve code organization and added integrated unit tests (`test_repo_specific_env_vars`, `test_env_vars_isolated_per_repo`) that validate per-repo environment variable isolation. This matches the pattern of per-repo env injection used in `TestRepo` and supports parallel test execution.
- These helpers are test-only; production behavior falls back to the real process environment when no override is set.

---

<a href="https://app.devin.ai/review/git-ai-project/git-ai/pull/536" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/src/commands/checkpoint.rs
+++ b/src/commands/checkpoint.rs
@@ -277,22 +277,22 @@ pub fn run(
 
     if show_working_log {
         if checkpoints.is_empty() {
-            debug_log("No working log entries found.");
+            eprintln!("No working log entries found.");
         } else {
-            debug_log("Working Log Entries:");
-            debug_log(&"=".repeat(80).to_string());
+            eprintln!("Working Log Entries:");
+            eprintln!("{}", "=".repeat(80));
             for (i, checkpoint) in checkpoints.iter().enumerate() {
-                debug_log(&format!("Checkpoint {}", i + 1));
-                debug_log(&format!("  Diff: {}", checkpoint.diff));
-                debug_log(&format!("  Author: {}", checkpoint.author));
-                debug_log(&format!(
+                eprintln!("Checkpoint {}", i + 1);
+                eprintln!("  Diff: {}", checkpoint.diff);
+                eprintln!("  Author: {}", checkpoint.author);
+                eprintln!(
                     "  Agent ID: {}",
                     checkpoint
                         .agent_id
                         .as_ref()
                         .map(|id| id.tool.clone())
                         .unwrap_or_default()
-                ));
+                );
 
                 // Display first user message from transcript if available
                 if let Some(transcript) = &checkpoint.transcript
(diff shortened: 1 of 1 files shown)


stuartsessions:
The code comparison is functionally the same to what was pre-existing: if a process variable is found, use that. Otherwise, use temp_dir(). The unsafe variable set was in a conditional statement that only got reached if there wasn't an existing env var.

The "GIT_AI_TEST_DB_PATH" is a test var, so shouldn't be set in user's code.

svarlamov:
Hey @stuartsessions thanks for the contribution! Watching the tests now -- is it possible that things actually got slower?

stuartsessions:
Entirely possible. It definitely didn't really speed up the integration test workflow, but running unit and integration tests separately should speed it up a little. It could also be that the overhead of doing a build step separate from the unit/integration testing introduced overhead which makes it slower now...

I tested the changes locally using `cargo test` and it ran just fine, so the tests will work fine without nextest.
I think the best way to change testing times will be to change the underlying process of tests. Performing integration tests on a per-crate basis, making fewer but more encompassing tests, etc. Or go all in on gitoxide or libgit2, but I think your intuition about teh overhead of calling git in subprocesses is correct.

stuartsessions:
I changed my main branch to work on a different issue not realizing it caused the PR to close. I can reopen this, just need to finagle my fork.
@svarlamov considering it doesn't improve the windows tests much, do we still want to include these changes?

svarlamov:
Hey @stuartsessions thanks for checking in! Probably not honestly...

We've also got HUGE updates to testing that were recently added to main and then there's even more coming in #530 with the hooks feature. Maybe after #530 lands, we could add some logic to split windows tests into 2 so that they could run in parallel?

PR #569, opened 2026-02-21
fix: handle cross-repo checkpoints when CWD is a different git repo

svarlamov:
# fix: handle cross-repo checkpoints when CWD is a different git repo

## Summary

When an AI agent working from `repos/repo-1` edits files in `repos/repo-2`, the checkpoint was only written to repo-1's working log. Files in repo-2 were silently filtered out by `path_is_in_workdir` in `checkpoint.rs`, so all changes showed as "human" when later committing in repo-2.

**Root cause:** In `handle_checkpoint()`, when the CWD is itself a git repo, `needs_file_based_repo_detection` is `false` and the code falls to the single-repo path. The single-repo checkpoint naturally ignores files outside its workdir — but nothing ever processes those external files.

**Fix:** After running the local repo checkpoint, detect any edited files outside the current repo's workdir, group them by their containing repository via `group_files_by_repository`, and run separate checkpoints for each external repo. Passes `None` as workspace boundary since external files are by definition outside the local workspace.

### Updates since initial revision
- Applied Devin Review suggestion: use repo `workdir` (not CWD `repository_working_dir`) when resolving relative external paths to absolute, fixing an inconsistency where the containment check and absolute path generation used different base directories.
- Fixed Rust 1.93.0 `rustfmt` formatting differences.
- Collapsed nested `if` to satisfy `clippy::collapsible_if` lint.
- **Removed `std::process::exit(0)` from the `Err` arm** of the local checkpoint match. Previously, a local checkpoint failure would immediately terminate the process, preventing cross-repo checkpoints from ever executing. Now the failure is tracked via `local_checkpoint_failed`, cross-repo checkpoints run regardless, and `exit(0)` is called afterward (preserving the existing behavior of not propagating checkpoint errors to callers). This mirrors the multi-repo workspace error handling at lines 692–701, which continues processing other repos on failure.
- Moved `spawn_background_flush()` out of the `Ok` arm to after cross-repo processing, so metrics are flushed for cross-repo checkpoints too.

- [ ] **Manually verify with a real agent in a multi-repo workspace.** Tests use `mock_ai` which doesn't exercise the full agent preset pipeline. From `repos/repo-1`, have an AI (Claude Code, Cursor, etc.) edit a file in `repos/repo-2`, then commit in repo-2 and run `git-ai blame` to confirm AI attribution.
- [ ] **Verify the `exit(0)` deferral is safe.** The local checkpoint `Err` arm no longer exits immediately — cross-repo checkpoints run first, then `exit(0)`. Confirm no callers depend on the process exiting before cross-repo side effects occur.
- [ ] **Check `AgentRunResult` clone cost.** `agent_run_result.as_ref().cloned()` copies the full struct including transcript. For agents with large transcripts, this could be non-trivial memory. Consider whether only the needed fields should be extracted.
- [ ] **Confirm `group_files_by_repository(&external_files, None)` with no workspace boundary is acceptable.** This walks up the filesystem unbounded looking for `.git` dirs. It only runs for files already identified as external, but verify no edge cases with symlinks or deeply nested paths cause issues.

### Notes
- The three new tests were confirmed to **fail before the fix** and **pass after**.
- CI passes on Ubuntu and macOS (all modes). Windows test jobs were still in progress at time of last check.
- [Link to Devin run](https://app.devin.ai/sessions/3550851e623b46fbb98592c11468acfa)
- Requested by @svarlamov

### DIFF
--- a/src/commands/git_ai_handlers.rs
+++ b/src/commands/git_ai_handlers.rs
@@ -834,6 +834,7 @@ fn handle_checkpoint(args: &[String]) {
         agent_run_result,
         false,
     );
+    let local_checkpoint_failed = checkpoint_result.is_err();
     match checkpoint_result {
         Ok((_, files_edited, _)) => {
             let elapsed = checkpoint_start.elapsed();
@@ -850,7 +851,6 @@ fn handle_checkpoint(args: &[String]) {
                 "checkpoint_kind": format!("{:?}", checkpoint_kind)
             });
             observability::log_error(&e, Some(context));
-            std::process::exit(0);
         }
     }
 
@@ -923,6 +923,10 @@ fn handle_checkpoint(args: &[String]) {
 
     if checkpoint_kind != CheckpointKind::Human {
         observability::spawn_background_flush();
+    }
+
+    if local_checkpoint_failed {
+        std::process::exit(0);
     }
 }
 


svarlamov:
devin fix the lints

svarlamov:
Devin fix: In src/commands/git_ai_handlers.rs, the Err arm of the checkpoint_result match (lines 843-854) calls std::process::exit(0), which prevents the cross-repo checkpoint processing on lines 856-921 from ever executing when the local checkpoint fails. To fix this, remove the std::process::exit(0) call and instead let execution fall through to the cross-repo processing block. You may want to track whether the local checkpoint succeeded/failed and still report the error, but the key change is to NOT exit the process so that cross-repo checkpoints can still run. This would be consistent with the multi-repo workspace path at lines 692-701 which continues processing other repos on failure. After the cross-repo processing, you can exit with code 0 if the local checkpoint failed (to preserve the existing behavior of not propagating checkpoint errors to callers).

Issue #344, opened 2026-01-13
Git commit causes memory overflow


wangkaizhuang:
<img width="1057" height="640" alt="Image" src="https://github.com/user-attachments/assets/d4695f87-cdc1-4d21-ae7e-f1e78e5c2f2d" />
After committing to Git, I found my computer freezing. Upon checking the memory usage, I suspected that it was caused by an infinite loop in Git-AI

wangkaizhuang:
<img width="1057" height="640" alt="Image" src="https://github.com/user-attachments/assets/df5dc47c-7399-4225-bef2-29a919604364" />
The memory usage issue is extremely unreasonable, and I am skeptical about it

svarlamov:
Hi @wangkaizhuang, I'm sorry to hear you're running into this issue. It's indeed quite out of the ordinary to see such memory usage and you're absolutely right to be frustrated. To help us investigate, could you share some more background on the commit? Was it an ordinary commit? How many files were changed? Did you have very long threads with the agents? We are shipping some major optimizations to this logic this week as we have heard that some users have ended up with extremely large checkpoints from long/loop-based agent threads. The issue, as I've seen it so far, is not the actual git ai processing logic, but simply the amount of data that the CLI is trying to load in at once.

If you're open to it, please send over your .git/ai folder from this repo to sasha at usegitai.com (only if these files and conversations you had are not confidential)

wangkaizhuang:
I submit more files, but I understand that this is a regular iterative task. I didn't feel any sluggishness when using Git before, and the memory usage of over 50G is outrageous to me. I have sent you my AI files

svarlamov:
This should be fixed as of the `1.2.x` release, but definitely in `1.4.0`

PR #571, opened 2026-02-21
fix: reduce memory overflow from checkpoint reads and writes (#344)

svarlamov:
# fix: reduce memory overflow from checkpoint reads and writes (#344)

## Summary

Addresses the runaway memory usage (30-60GB) reported in #344 by fixing the highest-impact patterns in checkpoint I/O:

**1. Streaming prune-and-rewrite on append** (`repo_storage.rs`): `append_checkpoint` previously read ALL checkpoints into memory, appended one, pruned, then wrote ALL back — O(N) memory per append. Now it streams existing checkpoints one-at-a-time through a `BufReader`, prunes char-level attributions for files superseded by the new checkpoint, writes to a temp file via `BufWriter`, appends the new checkpoint, then atomically renames. Peak memory is one checkpoint + the new checkpoint, rather than all checkpoints. The file stays small throughout long agent loops because pruning happens on every append.

**2. Eliminate redundant full reads** (`checkpoint.rs`): A single `checkpoint::run()` call previously triggered 4+ independent `read_all_checkpoints()` deserializations of the entire JSONL file. Now checkpoints are read once at the top of `run()` and passed through to `get_all_tracked_files` via a new `preloaded_checkpoints` parameter.

**3. Streaming reads** (`repo_storage.rs`): `read_all_checkpoints` now uses `BufReader` line-by-line instead of `fs::read_to_string`, avoiding holding the full file string and parsed structs in memory simultaneously.

**4. BufWriter for writes** (`repo_storage.rs`): `write_all_checkpoints` now streams serialization through `BufWriter` instead of building a full string in memory. An explicit `flush()` call ensures write errors are propagated rather than silently dropped on `BufWriter::drop`.

All 38 checkpoint-related unit tests pass (31 checkpoint tests + 7 repo_storage tests). No new dependencies added.

## Updates since last revision

**Major rework**: The initial approach deferred char-level attribution pruning to `write_all_checkpoints`, but this left un-pruned data in the file during the intra-commit loop where memory issues are worst. The new approach prunes on every `append_checkpoint` call using a streaming read-modify-write pattern that keeps only one checkpoint in memory at a time. The `prune_old_char_attributions` method was removed entirely (logic inlined into `append_checkpoint`).

**Reset ordering fix** (ef4a4d70): Fixed a data loss bug where `reset=true` + `is_pre_commit=true` would destroy the working log before the early-exit check, causing the check to always pass (empty checkpoints → no AI edits) and skip checkpoint creation. Now checkpoints are always read first for the early-exit check, then reset happens afterward if needed.

Other changes:
- `write_all_checkpoints` signature reverted to `&[Checkpoint]` (no longer needs `&mut` since pruning moved to append)
- Added explicit `writer.flush()?;` in both `append_checkpoint` and `write_all_checkpoints` to ensure I/O errors are propagated

- [ ] **Reset + pre_commit interaction**: When `reset=true` and `is_pre_commit=true`, verify that the early-exit check correctly inspects the old data before reset, and that the subsequent reset properly clears both disk state and in-memory checkpoints. Test with a repo that has AI checkpoints, run `checkpoint --reset` during pre-commit, and verify a fresh checkpoint is created (not skipped).
- [ ] **Streaming prune correctness**: The new `append_checkpoint` assumes the checkpoint being appended is the newest for its files (clears attributions from older entries with matching files). This should always be true since we append chronologically, but verify no code path appends out-of-order or writes un-pruned checkpoints via `write_all_checkpoints` directly.
- [ ] **Real-world validation**: Test with a repo that has a large checkpoint file (>100MB) and multiple agent sessions. Verify memory usage stays reasonable during `git commit` and that attributions are correctly preserved end-to-end. Unit tests validate correctness but not the memory improvement.
- [ ] **Temp file cleanup**: `append_checkpoint` writes to `checkpoints.jsonl.tmp` then renames. If the process crashes mid-write, the temp file is left behind (harmless but clutters `.git/ai/`). Consider if cleanup is needed.

### Notes

- `get_all_tracked_files` gained an optional `preloaded_checkpoints` parameter. Existing callers that don't pass it will still work (reads from disk as before).
- No changes to checkpoint format or serialization — purely I/O optimization.
- The streaming prune approach means `write_all_checkpoints` no longer prunes. If any code path writes un-pruned data via `write_all_checkpoints`, it won't be pruned until the next `append_checkpoint`.
- The `has_no_ai_edits` logic was rewritten from `all_ai_touched_files().is_empty()` to `checkpoints.iter().all(|cp| cp.entries.is_empty() || cp.kind != AiAgent/AiTab)`. These are logically equivalent but worth a careful trace through both code paths.

Link to Devin run: https://app.devin.ai/sessions/2a46b6eaa71f4f46913488bef2ff52a1
Requested by: @svarlamov

---

<a href="https://app.devin.ai/review/git-ai-project/git-ai/pull/571" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/src/authorship/post_commit.rs
+++ b/src/authorship/post_commit.rs
@@ -92,7 +92,7 @@ pub fn post_commit(
         );
     }
 
-    working_log.write_all_checkpoints(&mut parent_working_log)?;
+    working_log.write_all_checkpoints(&parent_working_log)?;
 
     // Create VirtualAttributions from working log (fast path - no blame)
     // We don't need to run blame because we only care about the working log data
--- a/src/authorship/post_commit.rs
+++ b/src/authorship/post_commit.rs
@@ -92,7 +92,7 @@ pub fn post_commit(
         );
     }
 
-    working_log.write_all_checkpoints(&parent_working_log)?;
+    working_log.write_all_checkpoints(&mut parent_working_log)?;
 
     // Create VirtualAttributions from working log (fast path - no blame)
     // We don't need to run blame because we only care about the working log data
--- a/src/commands/checkpoint.rs
+++ b/src/commands/checkpoint.rs
@@ -156,12 +156,29 @@ pub fn run(
         storage_start.elapsed()
     ));
 
+    // Read checkpoints once and reuse throughout the function.
+    // This eliminates multiple redundant read_all_checkpoints() calls that were
+    // previously the #1 cause of memory overflow (each read deserializes the entire
+    // JSONL file into memory).
+    let read_checkpoints_start = Instant::now();
+    let mut checkpoints = if reset {
+        working_log.reset_working_log()?;
+        Vec::new()
+    } else {
+        working_log.read_all_checkpoints()?
+    };
+    debug_log(&format!(
+        "[BENCHMARK] Reading {} checkpoints took {:?}",
+        checkpoints.len(),
+        read_checkpoints_start.elapsed()
+    ));
+
     // Early exit for human only
     if is_pre_commit {
-        let has_no_ai_edits = working_log
-            .all_ai_touched_files()
-            .map(|files| files.is_empty())
-            .unwrap_or(true);
+        let has_no_ai_edits = checkpoints.iter().all(|cp| {
+            cp.entries.is_empty()
+                || (cp.kind != CheckpointKind::AiAgent && cp.kind != CheckpointKind::AiTab)
+        });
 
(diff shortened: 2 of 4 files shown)


svarlamov:
devin review devin review's feedback

svarlamov:
Devin address the feedback

PR #644, opened 2026-03-06
fix: use ~/.git-ai/tmp/ for install script instead of /tmp

### LINKED ISSUE #643: Auto-upgrade silently fails when /tmp is noexec or permission-restricted
## Problem

The background auto-upgrade mechanism fails silently on systems where `/tmp` is
mounted with `noexec` or has restricted write permissions (common in corporate
and hardened Linux/macOS environments).

In `run_install_script` (Unix path), the install script is written to
`std::env::temp_dir()` (typically `/tmp`) before execution. If this write fails
due to permissions, the background process calls `process::exit(1)` with all
output suppressed (`stdout/stderr → /dev/null`), so the user sees nothing and
the auto-update silently stops working.

## Steps to reproduce

1. Mount `/tmp` as `noexec`: `mount -o remount,noexec /tmp`
2. Run any git command (which triggers `maybe_schedule_background_update_check`)
3. Observe: no update is applied, no error is shown

## Expected behavior

The upgrade should use a directory that git-ai already owns and has guaranteed
write access to (e.g., `~/.git-ai/tmp/`), rather than relying on `/tmp`.

## Proposed fix

Replace `std::env::temp_dir()` with `~/.config/git-ai/tmp/` (or platform
equivalent), creating the directory if needed, with fallback to `temp_dir()` if
the home path is unavailable.

txf0096:
## Problem

The background auto-upgrade silently fails on systems where `/tmp` is mounted with `noexec` or has restricted write permissions (common in corporate/hardened environments).

In `run_install_script` (Unix path), the install script is written to `std::env::temp_dir()` (typically `/tmp`) before execution. If `File::create` fails, the background process exits with code 1 — but since stdout/stderr are suppressed in the background worker, the user sees nothing and stays on the old version indefinitely.

Closes #643

## Fix

Use `~/.git-ai/tmp/` (via `config::git_ai_dir_path()`) as the temp dir for the install script. git-ai already has guaranteed write access to this directory (it stores config, internal state, etc. there). The directory is created if it doesn't exist. Falls back to `std::env::temp_dir()` if the home path is unavailable.

## Changes

- `src/commands/upgrade.rs`: replace `std::env::temp_dir()` with `~/.git-ai/tmp/` for the Unix install script temp path

---

<a href="https://app.devin.ai/review/git-ai-project/git-ai/pull/644" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/src/commands/upgrade.rs
+++ b/src/commands/upgrade.rs
@@ -442,8 +442,13 @@ fn run_install_script(script_content: &str, tag: &str, silent: bool) -> Result<(
         use std::io::Write;
         use std::os::unix::fs::PermissionsExt;
 
-        // Write script to a temp file
-        let temp_dir = std::env::temp_dir();
+        // Write script to ~/.git-ai/tmp/ to avoid /tmp noexec or permission issues.
+        // Fall back to the system temp dir if the home-based path is unavailable.
+        let temp_dir = crate::config::git_ai_dir_path()
+            .map(|p| p.join("tmp"))
+            .unwrap_or_else(std::env::temp_dir);
+        fs::create_dir_all(&temp_dir)
+            .map_err(|e| format!("Failed to create temp directory: {}", e))?;
         let script_path = temp_dir.join(format!("git-ai-install-{}.sh", std::process::id()));
 
         // Write and make executable


svarlamov:
Thank you for the contribution! Will make sure to test throughly soon and excited to merge this

PR #742, opened 2026-03-17
Rename cloud agents

acunniffe:
Rename background agents so it's possible to distinguish and compare differences between background agents and local agents. 

---

<a href="https://app.devin.ai/review/git-ai-project/git-ai/pull/742" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


src/authorship/agent_detection.rs
```diff
@@ -11,46 +11,53 @@ use crate::authorship::authorship_log_serialization::{
...
 const AGENT_EMAIL_MAPPINGS: &[(&str, &str)] = &[
-    ("cursoragent@cursor.com", "cursor"),
+    ("cursoragent@cursor.com", "cursor-agent"),
+    ("+copilot@users.noreply.github.com", "github-copilot"),
     (
-        "198982749+Copilot@users.noreply.github.com",
-        "github-copilot",
-    ),
-    (
-        "158243242+devin-ai-integration[bot]@users.noreply.github.com",
```
acunniffe, line 22: @svarlamov these old patterns looked wrong to me. Is there a reason to match exactly? I think those are app IDs and won't match for self-hosted devin + copilot?

svarlamov: Yeah good point -- we can remove the ID from the check so that it work on self-hosted instances

PR #743, opened 2026-03-17
Async

svarlamov:
---

<a href="https://app.devin.ai/review/git-ai-project/git-ai/pull/743" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/src/commands/install_hooks.rs
+++ b/src/commands/install_hooks.rs
@@ -203,8 +203,41 @@ fn maybe_configure_async_mode_daemon_trace2(dry_run: bool) -> Result<(), GitAiEr
     Ok(())
 }
 
+fn maybe_teardown_async_mode(dry_run: bool) {
+    if dry_run {
+        return;
+    }
+
+    let runtime_config = config::Config::fresh();
+    if runtime_config.feature_flags().async_mode {
+        return;
+    }
+
+    // Don't touch daemon inside test harnesses
+    if std::env::var_os("GIT_AI_TEST_DB_PATH").is_some()
+        || std::env::var_os("GITAI_TEST_DB_PATH").is_some()
+    {
+        return;
+    }
+
+    // Shut down any leftover daemon from when async_mode was enabled.
+    if let Ok(daemon_config) = DaemonConfig::from_env_or_default_paths()
+        && crate::commands::daemon::daemon_is_up(&daemon_config)
+    {
+        let _ = crate::daemon::send_control_request(
+            &daemon_config.control_socket_path,
+            &crate::daemon::ControlRequest::Shutdown,
+        );
+    }
+}
+
 fn maybe_ensure_daemon(dry_run: bool) {
     if dry_run {
+        return;
+    }
+
+    let runtime_config = config::Config::fresh();
+    if !runtime_config.feature_flags().async_mode {
         return;
     }
 
@@ -218,19 +251,6 @@ fn maybe_ensure_daemon(dry_run: bool) {
     let Ok(daemon_config) = DaemonConfig::from_env_or_default_paths() else {
         return;
     };
-
-    let runtime_config = config::Config::fresh();
-
-    if !runtime_config.feature_flags().async_mode {
-        // Async mode is off — shut down any leftover daemon.
-        if crate::commands::daemon::daemon_is_up(&daemon_config) {
-            let _ = crate::daemon::send_control_request(
-                &daemon_config.control_socket_path,
-                &crate::daemon::ControlRequest::Shutdown,
-            );
-        }
-        return;
-    }
 
     // If daemon is already running, shut it down first so it restarts
     // with the freshly-written trace2 config.
@@ -284,7 +304,9 @@ pub fn run(args: &[String]) -> Result<HashMap<String, String>, GitAiError> {
     }
 
     // In async mode, daemon trace2 config must be in place before any install work starts.
+    // If async mode was disabled, tear down any leftover daemon and trace2 config.
     maybe_configure_async_mode_daemon_trace2(dry_run)?;
+    maybe_teardown_async_mode(dry_run);
     maybe_ensure_daemon(dry_run);
 
     // Now that the daemon is (re)started, initialize the telemetry handle so


src/daemon.rs
```diff
@@ -0,0 +1,4364 @@
...
+    // If trace2.eventTarget points at this daemon socket globally, those internal git
+    // commands can recursively feed trace2 events back into the daemon and starve progress.
+    // Force-disable trace2 emission for the daemon process and all of its child git commands.
+    unsafe {
+        std::env::set_var("GIT_TRACE2_EVENT", "0");
+    }
+}
+
+pub async fn run_daemon(config: DaemonConfig) -> Result<(), GitAiError> {
+    disable_trace2_for_daemon_process();
```
jwiegley, line 6633: 🔴 **CRITICAL — Unsafe Code / Soundness** (Confidence: 97)

`disable_trace2_for_daemon_process()` calls `unsafe { std::env::set_var(...) }` *after* the Tokio multi-thread runtime is already running. Rust stdlib documents that `env::set_var` is not thread-safe when other threads exist; concurrent `getenv` on Tokio workers can race, causing heap corruption on glibc. Rust 2024 makes this a compile error.

**Fix:** Move the call to `handle_run` in `src/commands/daemon.rs` *before* `tokio::runtime::Builder::new_multi_thread().build()`:
```rust
// SAFETY: no other threads exist yet; Tokio runtime has not been built.
unsafe { std::env::set_var("GIT_TRACE2_EVENT", "0"); }
let runtime = tokio::runtime::Builder::new_multi_thread()...
```
Then remove `disable_trace2_for_daemon_process()` from `run_daemon` entirely.

src/daemon.rs
```diff
@@ -0,0 +1,4364 @@
...
+
+        #[cfg(unix)]
+        {
+            let rc = unsafe { libc::flock(file.as_raw_fd(), libc::LOCK_EX | libc::LOCK_NB) };
+            if rc != 0 {
+                return Err(GitAiError::Generic(
+                    "git-ai daemon is already running (lock held)".to_string(),
+                ));
+            }
+        }
```
jwiegley: 🔴 **CRITICAL — Unsafe Code Audit** (Confidence: 95)

Both `unsafe { libc::flock(...) }` blocks — here in `DaemonLock::acquire` and in its `Drop` impl (L236–240) — lack `// SAFETY:` comments. The invariants are: fd is valid (opened above), opened with read+write access, and Rust's field-drop ordering guarantees `_file` is dropped after the `Drop` body (so LOCK_UN fires before fd close). These are correct today but completely undocumented, making any future refactor silently unsafe.

**Fix:** Add a `// SAFETY:` comment to both `unsafe` blocks explaining the fd validity and drop ordering guarantee.

tests/async_mode.rs
```diff
@@ -0,0 +1,497 @@
...
+    for _ in 0..100 {
+        match daemon.try_wait() {
+            Ok(Some(_)) => return,
+            Ok(None) => thread::sleep(Duration::from_millis(20)),
+            Err(_) => break,
+        }
+    }
+    let _ = daemon.kill();
+    let _ = daemon.wait();
+}
```
jwiegley, line 432: 🔴 **CRITICAL — Test Resource Lifecycle** (Confidence: 97)

`daemon_status_does_not_self_emit_trace2_events` stores the daemon `Child` in a plain local variable with no RAII guard. If any assertion from L409–L444 panics, the `kill()` + `wait()` at L446–450 are never reached. The leaked daemon holds the control socket path, causing all subsequent daemon tests against this repo to hang at `wait_for_daemon_sockets`.

**Fix:** Wrap in a `scopeguard::defer!` (already a dependency) or the `DaemonGuard` pattern from `daemon_mode.rs`:
```rust
let mut daemon_process = Command::new(...).spawn().expect("...");
let _guard = scopeguard::defer(|| {
    let _ = daemon_process.kill();
    let _ = daemon_process.wait();
});
```

tests/async_mode.rs
```diff
@@ -0,0 +1,497 @@
...
+        target.is_none(),
+        "install-hooks dry-run should not set trace2.eventTarget"
+    );
+    assert!(
+        nesting.is_none(),
+        "install-hooks dry-run should not set trace2.eventNesting"
+    );
+}
+
+#[test]
```
jwiegley, line 289: 🔴 **CRITICAL — Test Isolation / Race Condition** (Confidence: 95)

`install_hooks_creates_configuration`, `install_hooks_enables_daemon_mode`, `checkpoint_starts_daemon`, and `checkpoint_increments_sequence` interact with daemon sockets or state but carry no `#[serial]` annotation. The test runner is free to execute them concurrently with the `#[serial]` tests in `daemon_mode.rs`, racing on the shared daemon control socket, trace socket, or repo state.

**Impact**: Intermittent non-deterministic failures under `cargo test`, worst on CI machines with high parallelism.

**Suggested fix** — add `#[serial]` to every daemon-touching test (applies here and to the other three functions in this file that start a daemon or call `wait_for_daemon_sockets` / `wait_for_daemon_latest_seq`):

```suggestion
#[serial]
#[test]
```

src/daemon.rs
```diff
@@ -0,0 +1,4364 @@
...
+
+fn control_listener_loop_actor(
+    control_socket_path: PathBuf,
+    coordinator: Arc<ActorDaemonCoordinator>,
+    runtime_handle: tokio::runtime::Handle,
+) -> Result<(), GitAiError> {
+    remove_socket_if_exists(&control_socket_path)?;
+    let listener = LocalSocketListener::bind(control_socket_path.to_string_lossy().as_ref())
+        .map_err(|e| GitAiError::Generic(format!("failed binding control socket: {}", e)))?;
+    set_socket_owner_only(&control_socket_path)?;
```
jwiegley: 🟠 **HIGH — Security / TOCTOU** (Confidence: 95)

`LocalSocketListener::bind()` creates the socket with umask-derived permissions (commonly `0o666`). `set_socket_owner_only()` — which applies `chmod 0600` — is called only *after* `bind()` returns. Any local process can connect in this window and send arbitrary `ControlRequest` messages, including `Shutdown` or `CheckpointRun` with attacker-controlled paths.

**Fix:** Set `umask(0o177)` around the `bind()` syscall so the socket is created `0o600` from the start:
```rust
// SAFETY: single-threaded daemon startup; umask is per-process.
let old_umask = unsafe { libc::umask(0o177) };
let listener = LocalSocketListener::bind(&config.control_socket_path)?;
unsafe { libc::umask(old_umask) };
```

src/daemon.rs
```diff
@@ -0,0 +1,4364 @@
...
+}
+
+fn read_json_line(reader: &mut BufReader<LocalSocketStream>) -> Result<Option<String>, GitAiError> {
+    let mut line = String::new();
+    let read = reader.read_line(&mut line)?;
+    if read == 0 {
+        return Ok(None);
+    }
+    Ok(Some(line))
+}
```
jwiegley: 🟠 **HIGH — Security / DoS** (Confidence: 92)

`read_json_line` calls `BufReader::read_line` with no size limit. Any local process that can connect to the daemon socket can send a single 4 GB line and exhaust the daemon's heap, killing it for all users.

**Fix:**
```rust
const MAX_LINE_BYTES: u64 = 4 * 1024 * 1024; // 4 MiB
fn read_json_line(reader: &mut BufReader<LocalSocketStream>) -> Result<Option<String>, GitAiError> {
    let mut line = String::new();
    let read = reader.get_mut().by_ref().take(MAX_LINE_BYTES).read_line(&mut line)?;
    if read == 0 { return Ok(None); }
    if line.len() as u64 >= MAX_LINE_BYTES {
        return Err(GitAiError::Generic("trace message too large".into()));
    }
    Ok(Some(line))
}
```

src/daemon/git_backend.rs
```diff
@@ -0,0 +1,472 @@
...
+                return Err(GitAiError::Generic(format!(
+                    "reflog cut regressed for {} ({} < {})",
+                    reference, end_offset, start_offset
+                )));
+            }
+            if end_offset == start_offset {
+                continue;
+            }
+
+            let reflog_path = common_dir.join("logs").join(&reference);
```
jwiegley, line 169: 🟠 **HIGH — Security / Path Traversal** (Confidence: 88)

Reference names from `ReflogCut` keys originating in the trace socket's reflog augmentation are joined to `logs/` as filesystem paths without validation for `..` sequences or null bytes. A ref name like `refs/heads/../../../../etc/shadow` causes the daemon to open that path.

**Fix:** Validate all reference names before using them as path components:
```rust
fn is_safe_reference(r: &str) -> bool {
    matches!(r, "HEAD" | "ORIG_HEAD")
        || (r.starts_with("refs/")
            && !r.contains("..")
            && !r.contains('')
            && r.split('/').all(|s| !s.is_empty()))
}
```

src/daemon.rs
```diff
@@ -0,0 +1,4364 @@
...
+        };
+        let coord = coordinator.clone();
+        let handle = runtime_handle.clone();
+        std::thread::spawn(move || {
+            if let Err(e) = handle_control_connection_actor(stream, coord, handle) {
+                debug_log(&format!("daemon control connection error: {}", e));
+            }
+        });
+    }
+    Ok(())
```
jwiegley: 🟠 **HIGH — Resource Exhaustion** (Confidence: 90)

`control_listener_loop_actor` calls `std::thread::spawn` for every accepted connection with no bound. A misbehaved client or parallel test run can exhaust OS thread limits (4096–32768 on Linux), after which connections are silently dropped, causing `barrier` / `wait_family_idle` to hang.

**Fix:** Use the `runtime_handle` already in scope to bound work as Tokio tasks, or add a semaphore:
```rust
let sem = Arc::new(tokio::sync::Semaphore::new(64));
// inside accept loop:
let permit = sem.clone().acquire_owned().await.ok()?;
runtime_handle.spawn(async move {
    let _permit = permit;
    handle_control_connection(stream, ...).await;
});
```

src/daemon.rs
```diff
@@ -0,0 +1,4364 @@
...
+            }
+            stale
+        };
+
+        for root_sid in stale_roots {
+            let mut payload = json!({
+                "event": "atexit",
+                "sid": root_sid,
+                "code": 0,
+                "time_ns": now_unix_nanos() as u64,
```
jwiegley, line 3703: 🟠 **HIGH — Correctness** (Confidence: 85)

`now_unix_nanos() as u64` silently truncates a `u128` value. Values fit today (current epoch ≈ 1.74×10¹⁸, `u64::MAX` ≈ 1.84×10¹⁹), but the cast is lossy at the type level and would overflow in year 2554 — or immediately if the function ever returns a value relative to a different epoch. Same pattern at L1917 and L1954.

**Suggested fix** — make the assumption explicit with a clear failure point:

```suggestion
                "time_ns": u64::try_from(now_unix_nanos()).expect("timestamp fits u64 until year 2554"),
```

scripts/benchmarks/git/benchmark_modes_vs_main.py
```diff
@@ -253,6 +257,168 @@ def __init__(
...
+        if self.variant.mode == "daemon" and self.daemon_started:
+            try:
+                run_cmd(
+                    [str(self.variant.binary), "daemon", "shutdown"],
+                    cwd=self.run_root,
+                    env=self.base_env,
+                    timeout_s=20,
+                )
+            except Exception:
+                pass
```
jwiegley, line 322: 🟠 **HIGH — Bug** (Confidence: 82)

`except Exception: pass` silently swallows all errors from the graceful daemon shutdown command — not just "daemon already gone". Orphaned daemon processes accumulate across benchmark runs with no diagnostic.

**Suggested fix** — at minimum log the failure to stderr:

```suggestion
            except Exception as err:
                print(f"warning: daemon shutdown failed: {err}", file=sys.stderr)
```

scripts/benchmarks/git/benchmark_nasty_rebases.sh
```diff
@@ -144,10 +174,20 @@ PY
...
+    (
+      cd "$REPO_DIR"
+      GIT_AI_DEBUG=0 GIT_AI_DEBUG_PERFORMANCE=0 \
+      GIT_AI_DAEMON_CHECKPOINT_DELEGATE=true \
+      GIT_AI_DAEMON_CONTROL_SOCKET="$GIT_AI_DAEMON_CONTROL_SOCKET" \
+      "$GIT_AI_BIN" checkpoint mock_ai >/dev/null
+    )
+  else
+    (
+      cd "$REPO_DIR"
```
jwiegley, line 187: 🟠 **HIGH — Bug** (Confidence: 92)

Both `cd "$REPO_DIR"` calls in `run_ai_checkpoint()` run inside a subshell without checking the return code. Under `set -euo pipefail`, a failing `cd` aborts the subshell with a cryptic error rather than a clear diagnostic.

**Fix:** Add an explicit error guard in each subshell:
```bash
cd "$REPO_DIR" || { echo "error: cannot cd to $REPO_DIR" >&2; exit 1; }
```

tests/daemon_mode.rs
```diff
@@ -0,0 +1,2462 @@
...
+fn unique_worktree_path(repo: &TestRepo, prefix: &str) -> PathBuf {
+    repo.path().parent().unwrap_or(repo.path()).join(format!(
+        "{}-{}",
+        prefix,
+        std::time::SystemTime::now()
+            .duration_since(std::time::UNIX_EPOCH)
+            .unwrap_or_default()
+            .as_nanos()
+    ))
+}
```
jwiegley, line 469: 🟠 **HIGH — Test Isolation / Flakiness** (Confidence: 83)

`unique_worktree_path` uses a timestamp-only suffix. Concurrent tests can call this within the same nanosecond (the OS batches `gettimeofday`), producing the same path and causing one thread to fail with "already exists".

**Fix:** Combine timestamp with a per-process atomic counter (cf. the pattern in `notes_sync_regression.rs`):
```rust
static COUNTER: AtomicU64 = AtomicU64::new(0);
fn unique_worktree_path(base: &Path) -> PathBuf {
    let nanos = SystemTime::now()...;
    let n = COUNTER.fetch_add(1, Ordering::Relaxed);
    base.join(format!("worktree-{}-{}", nanos, n))
}
```

tests/daemon_mode.rs
```diff
@@ -0,0 +1,2462 @@
...
+            .output()
+            .expect("git command should execute");
+        assert!(
+            output.status.success(),
+            "git {} failed: {}",
+            args.join(" "),
+            String::from_utf8_lossy(&output.stderr)
+        );
+        String::from_utf8_lossy(&output.stdout).trim().to_string()
+    };
```
jwiegley, line 2695: 🟠 **HIGH — Test Isolation** (Confidence: 86)

Several inline `run_git` closures here construct `Command::new(git)` without overriding `HOME`. Git reads `~/.gitconfig`, picking up credential helpers, proxy settings, `core.hooksPath`, and `init.defaultBranch` from the developer's real home directory. Tests pass locally but fail on CI with different global git configuration.

**Fix:** Set `HOME` and `GIT_CONFIG_GLOBAL` on every git `Command` inside a test:
```rust
cmd.env("HOME", repo.test_home_path())
   .env("GIT_CONFIG_GLOBAL", repo.test_home_path().join(".gitconfig"));
```
The `TestRepo::git` method already does this — inline closures should route through it or replicate its env setup.

tests/daemon_mode.rs
```diff
@@ -0,0 +1,2462 @@
...
+            &worker_b_dir,
+            file_b.as_str(),
+            format!("b-ai-line-{idx}").as_str(),
+        );
+    }
+
+    let _ = repo.git_og_with_env(
+        &["worktree", "remove", "--force", worker_b_dir_str.as_str()],
+        &env_refs,
+    );
```
jwiegley, line 3154: 🟠 **HIGH — Test Isolation / Cleanup** (Confidence: 89)

`let _ = repo.git_og_with_env(["worktree", "remove", ...])` discards the `Result`. Failed cleanup means stale worktrees accumulate on disk, and subsequent tests that create a worktree at the same path fail and attribute the error to the wrong test.

**Suggested fix** — assert cleanup so failures are attributed correctly:

```suggestion
    repo.git_og_with_env(
        &["worktree", "remove", "--force", worker_b_dir_str.as_str()],
        &env_refs,
    )
    .expect("worktree remove failed during test cleanup");
```

src/daemon.rs
```diff
@@ -0,0 +1,4364 @@
...
+pub mod control_api;
+pub mod coordinator;
+pub mod domain;
+pub mod family_actor;
+pub mod git_backend;
+pub mod global_actor;
+pub mod reducer;
+pub mod trace_normalizer;
+
+pub use control_api::{CheckpointRunRequest, ControlRequest, ControlResponse, FamilyStatus};
```
jwiegley: 🟡 **MEDIUM — API Design / Maintainability** (Confidence: 93)

Two distinct structs are both named `FamilyStatus` — one in `daemon/control_api.rs` and one in `daemon/domain.rs`. The `pub use control_api::FamilyStatus` re-export here means `use crate::daemon::FamilyStatus` silently resolves to the control-api metrics struct, not the domain type.

**Fix:** Rename the control-api struct to `DaemonFamilyStatus` (or `FamilyStatusReport`) to disambiguate.

src/daemon/reducer.rs
```diff
@@ -0,0 +1,235 @@
...
+    );
+}
+
+pub fn reduce_env_override(state: &mut FamilyState, env: EnvOverrideSet) {
+    state.applied_seq = state.applied_seq.saturating_add(1);
+    let key = env
+        .repo_working_dir
+        .canonicalize()
+        .unwrap_or(env.repo_working_dir);
+    state.env_overrides.insert(key, env.overrides);
```
jwiegley: 🟡 **MEDIUM — Correctness / Error Handling** (Confidence: 86)

`env.repo_working_dir.canonicalize().unwrap_or(env.repo_working_dir)` inserts with a non-canonical key when `canonicalize` fails. Since lookups use canonicalized paths, the override is silently dropped — `git ai daemon env-override` appears to succeed but has no effect.

**Fix:** Return an error on canonicalization failure instead of silently falling back.

src/commands/daemon.rs
```diff
@@ -0,0 +1,211 @@
...
+    child
+        .arg("daemon")
+        .arg("run")
+        .current_dir(&runtime_dir)
+        .stdin(Stdio::null())
+        .stdout(Stdio::null())
+        .stderr(Stdio::null());
+
+    child.spawn().map(|_| ()).map_err(|e| e.to_string())
+}
```
jwiegley, line 316: 🟡 **MEDIUM — Resource Management** (Confidence: 88)

`child.spawn().map(|_| ())` discards the `Child` handle immediately. If the daemon binary fails to start (permission error, missing file), the caller sees only "timed out waiting for daemon socket" with no diagnostic about why the child failed.

**Fix:** Log the PID before dropping, and check for early exit in the polling loop:
```rust
let child = Command::new(...).spawn()?;
eprintln!("git-ai daemon starting (pid={})", child.id());
// In the polling loop:
if let Ok(Some(status)) = child.try_wait() {
    return Err(GitAiError::Generic(format!("daemon exited early: {}", status)));
}
```

src/daemon.rs
```diff
@@ -0,0 +1,4364 @@
...
+        ))
+    }
+
+    async fn wait_for_family_idle(
+        &self,
+        repo_working_dir: String,
+        timeout_ms: Option<u64>,
+    ) -> Result<ControlResponse, GitAiError> {
+        let timeout_polls = timeout_ms.unwrap_or(8_000) / 10;
+        let mut polls_remaining = timeout_polls.max(1);
```
jwiegley: 🟡 **MEDIUM — Correctness** (Confidence: 87)

`let timeout_polls = timeout_ms.unwrap_or(8_000) / 10` with `stable_idle_polls >= 2` requires at least 20ms to ever succeed, but the API gives no indication of this minimum. Callers passing `Some(15)` always get a spurious timeout error.

**Fix:** Document the minimum in the function's doc comment and add a debug assertion:
```rust
debug_assert!(timeout_ms.unwrap_or(8_000) >= 20,
    "timeout_ms must be >= 20ms (requires 2 poll cycles)");
```

src/daemon/family_actor.rs
```diff
@@ -0,0 +1,274 @@
...
+    let mut idx = 0;
+    while idx < waiters.len() {
+        if applied_seq >= waiters[idx].0 {
+            let (_, waiter) = waiters.remove(idx);
+            let _ = waiter.send(Ok(()));
+        } else {
+            idx += 1;
+        }
+    }
+}
```
jwiegley: 🟡 **MEDIUM — Performance** (Confidence: 82)

`Vec::remove(idx)` inside the scan loop is O(n) per removal, making the loop O(n²) when many barriers are satisfied at once (common under test-suite burst load).

**Suggested fix** — replaces the full function with `retain_mut`:

```suggestion
fn satisfy_barriers(
    applied_seq: u64,
    waiters: &mut Vec<(u64, oneshot::Sender<Result<(), GitAiError>>)>,
) {
    waiters.retain_mut(|(seq, sender)| {
        if applied_seq >= *seq {
            let _ = sender.send(Ok(()));
            false
        } else {
            true
        }
    });
}
```

src/daemon/analyzers/history.rs
```diff
@@ -0,0 +1,902 @@
...
+fn normalized_args(argv: &[String]) -> Vec<String> {
+    let start = argv
+        .first()
+        .and_then(|arg| Path::new(arg).file_name().and_then(|name| name.to_str()))
+        .is_some_and(|name| name == "git" || name == "git.exe");
+    if start {
+        argv[1..].to_vec()
+    } else {
+        argv.to_vec()
+    }
```
jwiegley, line 173: 🟡 **MEDIUM — Maintainability** (Confidence: 90)

`command_args` and `normalized_args` are verbatim-copied across `analyzers/history.rs`, `analyzers/transport.rs`, and `analyzers/workspace.rs`. A bug fix must be applied three times.

**Fix:** Move both helpers to `analyzers/mod.rs` (or a new `analyzers/util.rs`) and import them in each analyzer module.

src/daemon/control_api.rs
```diff
@@ -0,0 +1,102 @@
...
+    CheckpointRun {
+        request: Box<CheckpointRunRequest>,
+        wait: Option<bool>,
+    },
+    #[serde(rename = "env.override")]
+    EnvOverride {
+        repo_working_dir: String,
+        env: HashMap<String, String>,
+        wait: Option<bool>,
+    },
```
jwiegley: 🟡 **MEDIUM — Security** (Confidence: 90)

`EnvOverride` accepts an unbounded `HashMap<String, String>` with no restriction on keys and no cap on the number of distinct `repo_working_dir` entries in `state.env_overrides`. An attacker or buggy tool can grow this map without bound.

**Fix:** Define an explicit allowlist of accepted override keys, reject unknown ones at ingestion, and cap the number of tracked worktrees.

src/daemon/reducer.rs
```diff
@@ -0,0 +1,235 @@
...
+    state.worktrees.insert(
+        canonicalize_path(worktree),
+        WorktreeState {
+            head: post_repo.head.clone(),
+            branch: post_repo.branch.clone(),
+            detached: post_repo.detached,
+            last_updated_ns: cmd.finished_at_ns,
+        },
+    );
+}
```
jwiegley, line 141: 🟡 **MEDIUM — Security / Resource Exhaustion** (Confidence: 85)

`FamilyState.checkpoints` grows proportionally to unique checkpoint IDs ever processed with no eviction — unlike `recent_commands` which has `RECENT_COMMAND_CAP`. Over a long benchmark session with 220+ feature commits this is a memory leak. Same issue with `.worktrees`.

**Fix:** Apply LRU eviction or a fixed-size ring to both maps.

src/daemon.rs
```diff
@@ -0,0 +1,4364 @@
...
+            config.control_socket_path = PathBuf::from(path);
+        }
+
+        if let Ok(path) = std::env::var("GIT_AI_DAEMON_TRACE_SOCKET")
+            && !path.trim().is_empty()
+        {
+            config.trace_socket_path = PathBuf::from(path);
+        }
+
+        Ok(config)
```
jwiegley, line 232: 🟡 **MEDIUM — Security** (Confidence: 87)

`GIT_AI_DAEMON_CONTROL_SOCKET` and `GIT_AI_DAEMON_TRACE_SOCKET` are accepted from the environment and used as-is for `LocalSocketListener::bind`. A compromised parent process can redirect the daemon to bind in a world-writable location, undermining the `0o600` permission hardening.

**Fix:** Validate that env-overridden socket paths resolve to a directory owned by the current user before binding.

src/daemon.rs
```diff
@@ -0,0 +1,4364 @@
...
+                            "phase": "trace_ingest_worker",
+                            "reason": "missing_ingest_seq",
+                            "payload": payload,
+                        })),
+                    );
+                    coordinator.request_shutdown();
+                    break;
+                };
+
+                if pending_by_seq.insert(seq, payload).is_some() {
```
jwiegley, line 3983: 🟡 **MEDIUM — Security / DoS** (Confidence: 85)

`pending_by_seq` has no size cap. If a git process sends frames with `seq > next_seq` and then hangs, all frames accumulate in the `BTreeMap` indefinitely. Multiplied across concurrent operations this is a DoS vector.

**Fix:** Add a maximum pending buffer size (e.g., 4096 entries); if exceeded, drop the connection and log an error.

scripts/benchmarks/git/benchmark_modes_vs_main.py
```diff
@@ -235,7 +236,10 @@ def __init__(
         self.run_root = run_root
         self.timeout_s = timeout_s
 
-        self.home_dir = run_root / "home"
+        tmp_root = Path("/tmp") if os.name != "nt" else Path(tempfile.gettempdir())
```
jwiegley, line 239: 🟡 **MEDIUM — Convention** (Confidence: 88)

Hardcoded `Path("/tmp")` ignores `TMPDIR` system configuration. On macOS, `tempfile.gettempdir()` returns a user-scoped path under `/var/folders/...`. On CI machines `/tmp` may be a smaller or slower volume.

**Fix:** Use `Path(tempfile.gettempdir())` consistently.

scripts/benchmarks/git/benchmark_nasty_modes_vs_main.py
```diff
@@ -332,7 +402,13 @@ def render_report(
...
 ) -> None:
     scenarios = ["linear", "onto", "rebase_merges"]
-    variants = ["main_wrapper", "current_wrapper", "current_hooks", "current_both"]
+    variants = [
+        "main_wrapper",
+        "current_wrapper",
+        "current_hooks",
+        "current_both",
+        "current_daemon",
+    ]
```
jwiegley, line 411: 🔵 **LOW — Style** (Confidence: 100)

`variants` is assigned but never referenced in `render_report`. Editing this list has no effect on rendering.

**Fix:** Remove the unused assignment.

scripts/nightly/verify-hook-wiring.sh
```diff
@@ -52,7 +52,7 @@ case "$AGENT" in
     ;;
 
   opencode)
-    PLUGIN="$HOME/.config/opencode/plugin/git-ai.ts"
+    PLUGIN="$HOME/.config/opencode/plugins/git-ai.ts"
     [ -f "$PLUGIN" ] || fail "OpenCode plugin not found at $PLUGIN"
     pass "OpenCode git-ai plugin installed at $PLUGIN"
     ;;
```
jwiegley, line 58: 🔵 **LOW — Test Coverage** (Confidence: 82)

Every other agent branch greps for the `checkpoint` invocation; the `opencode` branch only checks `[ -f "$PLUGIN" ]`. A zero-byte or placeholder file would pass.

```suggestion
  if [[ -f "$PLUGIN" ]] && grep -q 'checkpoint' "$PLUGIN"; then
    pass "opencode: checkpoint hook found at $PLUGIN"
  else
    fail "opencode: checkpoint hook not found or missing invocation in $PLUGIN"
  fi
```

src/daemon/family_actor.rs
```diff
@@ -0,0 +1,274 @@
...
+
+pub fn spawn_family_actor(family_key: FamilyKey) -> FamilyActorHandle {
+    let (tx, mut rx) = mpsc::channel::<FamilyMsg>(1024);
+    let handle = FamilyActorHandle {
+        family_key: family_key.clone(),
+        tx,
+    };
+
+    tokio::spawn(async move {
```
jwiegley, line 73: 🔵 **LOW — Async Correctness** (Confidence: 82)

`spawn_family_actor` discards the `JoinHandle` from `tokio::spawn`. Actor panics are absorbed silently, causing subsequent `barrier` waits to hang or fail with cryptic errors. Same issue in `src/daemon/global_actor.rs:43`.

**Fix:** Store the `JoinHandle` in the actor handle struct and inspect it on subsequent barrier/message operations.

src/daemon.rs
```diff
@@ -0,0 +1,4364 @@
...
+                self.wait_for_family_idle(repo_working_dir, timeout_ms)
+                    .await
+            }
+            ControlRequest::Shutdown => Ok(ControlResponse::ok(None, None, None)),
+        };
+
+        match result {
+            Ok(response) => response,
+            Err(error) => ControlResponse::err(error.to_string()),
+        }
```
jwiegley, line 6050: 🔵 **LOW — Security** (Confidence: 82)

All `Err(error)` results from `handle_control_request` are forwarded to the client as `ControlResponse::err(error.to_string())`. Many `GitAiError` messages include filesystem paths and internal state details, providing reconnaissance to clients that intentionally trigger errors.

**Fix:** Return a generic `"internal error"` to clients while logging detailed errors internally.

tests/async_mode.rs
```diff
@@ -0,0 +1,497 @@
...
+        if latest_seq >= min_seq {
+            return;
+        }
+        thread::sleep(Duration::from_millis(25));
+    }
+    panic!(
+        "daemon did not observe traced command for {}",
+        repo.canonical_path().display()
+    );
+}
```
jwiegley, line 159: 🟡 **MEDIUM — Flakiness** (Confidence: 81)

Both `wait_for_daemon_sockets` (200 × 25ms = 5s) and `wait_for_daemon_latest_seq` use hard-coded iteration counts. The 30-second limit in `DaemonProcess::wait_until_ready` shows slow startup is known to be possible on loaded CI.

**Fix:** Extract iteration counts into named constants defaulting to match the 30-second bound. Respect a `GIT_AI_TEST_DAEMON_TIMEOUT_SECS` env var for CI overrides.

tests/daemon_mode.rs
```diff
@@ -0,0 +1,2462 @@
...
+        stream
+            .write_all(b"\n")
+            .expect("failed to write trace newline");
+    }
+    stream.flush().expect("failed to flush trace payloads");
+}
+
+fn repo_workdir_string(repo: &TestRepo) -> String {
+    repo.path().to_string_lossy().to_string()
+}
```
jwiegley, line 115: 🟡 **MEDIUM — Test Correctness / Flakiness** (Confidence: 80)

`repo.path().to_string_lossy()` returns `/tmp/...` on macOS, but the daemon keys families on the canonical path `/private/tmp/...`. All `StatusFamily` and `WaitFamilyIdle` requests miss the daemon's family record, causing every `wait_for_daemon_latest_seq` to time out and panic on macOS.

**Fix:**
```rust
repo.path()
    .canonicalize()
    .unwrap_or_else(|_| repo.path().to_path_buf())
    .to_string_lossy()
    .into_owned()
```

tests/daemon_mode.rs
```diff
@@ -0,0 +1,2462 @@
...
+    let mut observed = 0usize;
+    for _ in 0..200 {
+        observed = rewrite_event_count(repo, marker);
+        if observed >= expected_count {
+            return observed;
+        }
+        thread::sleep(Duration::from_millis(25));
+    }
+    observed
+}
```
jwiegley, line 581: 🟡 **MEDIUM — Assertion Quality** (Confidence: 82)

`wait_for_rewrite_event_count` polls until it observes *at least* `expected_count` events. A caller using `>=` silently misses over-emission bugs. The "at least" semantics are not communicated by the function name or signature.

**Fix:** Rename to `wait_for_at_least_rewrite_event_count`, or add an `exact: bool` parameter that asserts `== expected_count` when true.

jwiegley:
## Code Review — Companion Findings (PR #743)

The inline review comment posted above covers the majority of findings. This companion comment covers additional findings where the referenced lines are outside the PR's diff hunks, plus one coverage gap that applies to the entire file.

---

### 🔴 Critical

#### PID reuse hazard in `reap_child_if_exited` after `Child` is dropped
**File**: `tests/integration/repos/test_repo.rs#L2202–L2224` | **Confidence**: 92

`reap_child_if_exited` extracts the raw PID integer from the `Arc<Mutex<Option<Child>>>` — taking the `Child` out and setting the `Option` to `None` — then calls `libc::waitpid` on that integer. Dropping `Child` does **not** call `waitpid`. Between the drop and the manual `waitpid`, the OS can recycle the PID to an unrelated process. `is_process_alive` then uses `libc::kill(pid, 0)` on the same potentially-recycled PID.

**Impact**: On loaded CI this produces a spurious "daemon is alive" result, causing `wait_until_ready` to loop 30 seconds before panicking with a misleading error.

**Fix**: Keep the `Child` alive (do not `take()` it from the `Option`) until after `waitpid` returns. Alternatively, use `Child::try_wait` throughout and fall back to raw `libc` calls only with the `Child` still in scope.

---

### 🟠 High

#### Daemon process leaked when `VariantRunner.__init__` raises
**File**: `scripts/benchmarks/git/benchmark_modes_vs_main.py#L1236–L1279` | **Confidence**: 97

`VariantRunner(...)` is constructed before the `try` block, so if `start_daemon()` (called from `__init__`) raises `BenchmarkError` after setting `self.daemon_process`, the `finally: runner.close()` is never reached. The daemon process runs indefinitely.

**Fix**: Move `runner = VariantRunner(...)` inside the `try` block, or add a try/finally in `__init__` that kills `self.daemon_process` on construction failure.

---

#### Unquoted `$*` in diagnostic `echo` in benchmark
**File**: `scripts/benchmarks/git/benchmark_nasty_rebases.sh#L211` | **Confidence**: 90

`echo "command: $GIT_BIN -C $REPO_DIR $*"` uses unquoted `$*`, which undergoes word splitting and pathname expansion. Arguments with spaces produce misleading diagnostic output.

**Fix**:
```bash
echo "command: $GIT_BIN -C $REPO_DIR $(printf '%q ' "$@")"
```

---

#### `status` variable not declared `local` inside `run_rebase_scenario`
**File**: `scripts/benchmarks/git/benchmark_nasty_rebases.sh#L221` | **Confidence**: 88

`status` is assigned inside `run_rebase_scenario()` but never declared `local`, so it bleeds into the global scope. A stale value from a previous call can be reported for the current scenario.

**Fix**: Add `local status` to the local declarations at the top of `run_rebase_scenario()`.

---

#### `--burst-every 0` causes modulo divide-by-zero in benchmark
**File**: `scripts/benchmarks/git/benchmark_nasty_rebases.sh#L375–L388` | **Confidence**: 85

`(( i % BURST_EVERY == 0 ))` with `BURST_EVERY=0` causes a divide-by-zero. Bash exits the arithmetic context with non-zero status, and under `set -e` the script aborts mid-benchmark with no diagnostic.

**Fix**: Validate all numeric CLI arguments after parsing:
```bash
[[ "$BURST_EVERY" =~ ^[1-9][0-9]*$ ]] || {
  echo "error: --burst-every must be a positive integer" >&2; exit 1
}
```
Apply the same to `--files`, `--feature-commits`, `--main-commits`, `--side-commits`, `--lines-per-file`.

---

### 🟡 Medium

#### `geometric_mean` crashes with `ValueError` when any ratio is zero
**File**: `scripts/benchmarks/git/benchmark_modes_vs_main.py#L930–L933` | **Confidence**: 85

Ratios are guarded for `base > 0` but not `med > 0`. A zero median causes `math.log(0)` to raise `ValueError`, crashing report generation after a successful benchmark run. The nasty benchmark correctly guards with `base > 0 and med > 0`.

**Fix**: `if base > 0 and med > 0: ratios[key].append(med / base)`

---

#### `prepare_main_worktree` missing parent directory creation in nasty benchmark
**File**: `scripts/benchmarks/git/benchmark_nasty_modes_vs_main.py#L105–L113` | **Confidence**: 92

Unlike the modes benchmark equivalent (which calls `worktree_dir.parent.mkdir(parents=True, exist_ok=True)` first), the nasty benchmark skips this and fails with a confusing `git` error on a clean `--work-root`.

**Fix**: Add `worktree_dir.parent.mkdir(parents=True, exist_ok=True)` before `shutil.rmtree`.

---

#### `grep -c` with `|| true` masks unreadable log files
**File**: `scripts/benchmarks/git/benchmark_nasty_rebases.sh#L242` | **Confidence**: 95

`grep -c` returns exit code 1 both when the count is zero AND when the file is missing/unreadable. `|| true` masks the file-not-found case, setting `saved_count` to an empty string and silently corrupting the TSV output.

**Fix**:
```bash
[[ -f "$clean_log" ]] || { echo "error: log file not found: $clean_log" >&2; exit 1; }
saved_count="$(grep -c 'Saved authorship log' "$clean_log" 2>/dev/null || echo 0)"
```

---

#### `echo -e` is non-portable in benchmark script
**File**: `scripts/benchmarks/git/benchmark_nasty_rebases.sh#L449` | **Confidence**: 88

`echo -e` behavior is implementation-defined; on `/bin/sh` on macOS, `-e` is printed literally. Even in Bash, `xpg_echo` can change behavior.

**Fix**: `printf 'scenario\tstatus\tduration_s\tsaved_logs\thead_note\n'`

---

#### `RESULTS_DIR` defaults to a predictable `/tmp` path in nightly script
**File**: `scripts/nightly/verify-hook-wiring.sh#L8` | **Confidence**: 83

`RESULTS_DIR="${RESULTS_DIR:-/tmp/test-results}"` is a fixed, predictable path. A local attacker can pre-create `/tmp/test-results` with forged PASS log files, or symlink it to redirect writes.

**Fix**: Use `mktemp -d` when `RESULTS_DIR` is not provided, or validate that the directory is owned by the current user before writing to it.

---

#### `std::mem::forget` on base `TestRepo` is undocumented
**File**: `tests/integration/repos/test_repo.rs#L796` | **Confidence**: 84

`std::mem::forget(base)` is called without explanation. If construction panics between the `forget` call and resource transfer, the base daemon and temp directory are permanently leaked.

**Fix**: Add a `// SAFETY:` comment explaining the ownership transfer. Better: use `ManuallyDrop<TestRepo>` to make the intent explicit at the type level.

---

#### No test for daemon crash-and-restart recovery
**File**: `tests/daemon_mode.rs` (coverage gap) | **Confidence**: 85

The `DaemonGuard::shutdown` kill-fallback path has never been exercised. Behavior after `SIGKILL` — do subsequent git commands degrade gracefully or block indefinitely? — is unverified.

**Fix**: Add a `daemon_crash_recovery` test that starts a daemon, establishes a baseline, sends `SIGKILL`, issues another git command, and asserts graceful degradation or a clear error.

---

### 🔵 Low

#### `FeatureFlags::from_env` re-reads all env vars on every call
**File**: `src/feature_flags.rs#L85–L111` | **Confidence**: 80

`from_env_and_file` calls `envy::prefixed("GIT_AI_").from_env()` on every invocation. If called in per-command or per-trace paths this re-reads all `GIT_AI_*` env vars each time.

**Fix**: Document that `FeatureFlags` should be constructed once and stored, or use `OnceLock<FeatureFlags>` for global access.

---

#### Benchmark script passes unvalidated `--repo-url` to `git clone`
**File**: `scripts/benchmarks/git/benchmark_nasty_rebases.sh#L290` | **Confidence**: 80

`REPO_URL` is passed directly to `git clone` without scheme validation. A user could pass `ext::sh -c 'malicious' %s` or `file:///system/path` as the URL.

**Fix**:
```bash
[[ "$REPO_URL" =~ ^https?:// ]] || [[ "$REPO_URL" =~ ^git@ ]] || {
  echo "error: --repo-url must be https:// or git@" >&2; exit 1
}
```

---

#### Missing `trap` for `EXIT`/`ERR` in `verify-hook-wiring.sh`
**File**: `scripts/nightly/verify-hook-wiring.sh#L1` | **Confidence**: 80

If the script is killed mid-run (SIGTERM, SIGINT), the log file contains a partial record with no final PASS/FAIL line. A CI consumer may treat an incomplete log as passing.

**Fix**: `trap 'echo "INTERRUPTED: hook wiring check did not complete" >> "$LOG"' EXIT INT TERM`

---

#### `[ ]` used instead of `[[ ]]` in Bash script
**File**: `scripts/nightly/verify-hook-wiring.sh#L22`, `#L30`, `#L40`, `#L48`, `#L56` | **Confidence**: 90

The script uses POSIX `[ ]` despite a `#!/usr/bin/env bash` shebang. While correct today, it's a footgun for future contributors who add unquoted expansions inside `[ ]` tests.

**Fix**: Replace all `[ -f "$VAR" ]` with `[[ -f "$VAR" ]]`.

---

#### `expect` with eager `format!` swallows the OS error on test command failure
**File**: `tests/integration/repos/mod.rs#L103–L105`, `#L197` | **Confidence**: 80

`.expect(&format!("Failed to execute git command: {:?}", args))` only includes the args, not the `io::Error` reason (e.g., `ENOENT`, `EMFILE`).

**Fix**: `.unwrap_or_else(|e| panic!("Failed to execute git command: {:?} — OS error: {}", args, e))`

---

#### `#[allow(dead_code)]` blanket suppression on generated test structs
**File**: `tests/integration/repos/mod.rs#L31` | **Confidence**: 80

`#[allow(dead_code)]` is applied to entire `impl` blocks, suppressing dead-code warnings for all methods. A future change that removes a call site leaves methods silently unreported as unused.

**Fix**: Apply the suppression only to specific methods that are genuinely not called, with a comment explaining why.

tests/integration/repos/test_repo.rs
```diff
@@ -589,6 +1146,256 @@ impl TestRepo {
...
+    }
+
+    pub(crate) fn mark_daemon_family_dirty(&self) {
+        if !self.git_mode.uses_daemon() {
+            return;
+        }
+        let family_key = self.daemon_family_key();
+        let mut registry = daemon_sync_registry()
+            .lock()
+            .unwrap_or_else(|poisoned| poisoned.into_inner());
```
jwiegley, line 1667: 🟠 **HIGH — Test Isolation / Error Handling** (Confidence: 91)

`DaemonSyncRegistry::lock` calls `.unwrap_or_else(|e| e.into_inner())` on a poisoned mutex. When a test panics while holding the lock, subsequent tests silently continue with a registry that may be in an inconsistent mid-update state.

**Suggested fix** — fail fast so the true cause is visible:

```suggestion
            .expect("DaemonSyncRegistry lock poisoned — a prior test panicked");
```

src/daemon/coordinator.rs
```diff
@@ -0,0 +1,265 @@
...
+
+    pub async fn barrier_family(
+        &self,
+        repo_working_dir: &Path,
+        seq: u64,
+    ) -> Result<(), GitAiError> {
+        let family = self.backend.resolve_family(repo_working_dir)?;
+        let actor = self.get_or_create_family_actor(family).await;
+        actor.barrier(seq).await
+    }
```
jwiegley: 🟠 **HIGH — Async Correctness / Deadlock** (Confidence: 92)

`barrier_family` `await`s the receiver with no timeout. If the family actor task panics (its `JoinHandle` is dropped), the caller hangs indefinitely.

**Suggested fix** — wrap in a 30-second timeout (requires `use tokio::time::Duration`):

```suggestion
        tokio::time::timeout(
            std::time::Duration::from_secs(30),
            actor.barrier(seq),
        )
        .await
        .map_err(|_| GitAiError::Generic("barrier_family timed out waiting for seq".to_string()))?
    }
```

src/daemon.rs
```diff
@@ -0,0 +1,4364 @@
...
+                        Some(serde_json::json!({
+                            "component": "daemon",
+                            "phase": "trace_ingest_worker",
+                            "reason": "duplicate_ingest_seq",
+                            "sequence": seq,
+                        })),
+                    );
+                    coordinator.request_shutdown();
+                    break;
+                }
```
jwiegley, line 3999: 🟠 **HIGH — Error Handling / Reliability** (Confidence: 88)

Receiving a duplicate sequence number calls `coordinator.request_shutdown()` and exits the ingest worker. While duplicates shouldn't happen normally, a race or integer overflow could trigger this, and the result is full daemon death with no recovery path.

**Suggested fix** — log and skip, letting the daemon continue:

```suggestion
                    // Skip duplicate; earlier payload wins, daemon continues.
                    continue;
                }
```

scripts/benchmarks/git/benchmark_modes_vs_main.py
```diff
@@ -253,6 +257,168 @@ def __init__(
...
+        if self.daemon_process is not None:
+            deadline = time.time() + 5.0
+            while time.time() < deadline:
+                if self.daemon_process.poll() is not None:
+                    break
+                time.sleep(0.05)
+
+            if self.daemon_process.poll() is None:
+                self.daemon_process.kill()
+                self.daemon_process.wait(timeout=5)
```
jwiegley, line 333: 🟠 **HIGH — Bug / Resource Leak** (Confidence: 90)

After `kill()`, `wait(timeout=5)` can still raise `subprocess.TimeoutExpired` on certain OS conditions. In `VariantRunner.close()` this exception is uncaught, preventing `shutil.rmtree(self.home_dir)` from running.

**Suggested fix** — catch `TimeoutExpired` and continue cleanup:

```suggestion
                try:
                    self.daemon_process.wait(timeout=5)
                except subprocess.TimeoutExpired:
                    pass  # process is unkillable; continue cleanup
```

src/daemon/analyzers/history.rs
```diff
@@ -0,0 +1,902 @@
...
+                        old: old_head,
+                        new: new_head,
+                    });
+                }
+            }
+            _ => {
+                return Err(GitAiError::Generic(format!(
+                    "history analyzer does not support command '{}'",
+                    name
+                )));
```
jwiegley, line 137: 🟡 **MEDIUM — Error Handling / API Contract** (Confidence: 85)

The `_ =>` arm returns `Err(GitAiError::Generic(...))` for commands the registry should never route here. A future maintainer who adds a command without updating this analyzer will see silent discards rather than a test panic. Same pattern in `WorkspaceAnalyzer` and `TransportAnalyzer`.

**Suggested fix** — use `unreachable!` so invariant violations surface immediately:

```suggestion
                unreachable!(
                    "history analyzer invoked for unsupported command '{}'",
                    name
                );
```

scripts/benchmarks/git/benchmark_modes_vs_main.py
```diff
@@ -253,6 +257,168 @@ def __init__(
...
+        exit_code: int | None = None
+        for _ in range(300):
+            if self.daemon_control_socket.exists() and self.daemon_trace_socket.exists():
+                self.daemon_started = True
+                if proc.poll() is not None:
+                    self.daemon_process = None
+                return
+            if exit_code is None:
+                exit_code = proc.poll()
+            time.sleep(0.01)
```
jwiegley, line 305: 🟠 **HIGH — Bug** (Confidence: 95)

`exit_code` is assigned inside the poll loop but never actually checked — the `if exit_code is None:` guard on L303 means `proc.poll()` is called once and the result is stashed but the non-None branch never raises. If the daemon crashes at startup the user sees "Timed out waiting for daemon sockets" rather than the actual exit code.

**Suggested fix** — remove the dead variable and raise immediately on early exit:

```suggestion
        for _ in range(300):
            if self.daemon_control_socket.exists() and self.daemon_trace_socket.exists():
                self.daemon_started = True
                if proc.poll() is not None:
                    self.daemon_process = None
                return
            exit_code = proc.poll()
            if exit_code is not None:
                raise BenchmarkError(
                    f"Daemon exited early with code {exit_code} before sockets appeared"
                )
            time.sleep(0.01)
```

PR #825, opened 2026-03-27
fix: recover from stale daemon by killing unresponsive process

svarlamov:
## Summary
- When the daemon holds its lock but sockets are unreachable (stale/zombie state), every git command floods stderr with 3 error lines
- Instead of failing immediately, the wrapper now reads the daemon PID from metadata, sends SIGTERM (escalating to SIGKILL after 1s), waits for the lock to release, and restarts a fresh daemon
- Applied to both `ensure_daemon_running` and `ensure_daemon_running_attached` code paths

## Test plan
- [ ] Build a debug binary, start the daemon, delete its sockets manually, run a git command — should recover and restart the daemon instead of flooding errors
- [ ] Normal daemon operation (healthy daemon) is unaffected — the recovery path is only entered when `daemon_is_up()` is false AND lock is held
- [ ] If PID metadata is missing/corrupt, falls back to the existing error behavior
- [ ] On Windows, the new function is a no-op — existing behavior preserved

---

<a href="https://app.devin.ai/review/git-ai-project/git-ai/pull/825" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/src/commands/daemon.rs
+++ b/src/commands/daemon.rs
@@ -107,10 +107,15 @@ fn ensure_daemon_running_attached(timeout: Duration) -> Result<DaemonConfig, Str
     }
 
     if daemon_startup_is_blocked(&config) {
-        return Err(format!(
-            "daemon startup blocked: lock held at {}",
-            config.lock_path.display()
-        ));
+        // Daemon holds the lock but its sockets are unreachable — it is stale.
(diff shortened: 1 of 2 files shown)


svarlamov:
Wrong root cause — the daemon was actually running fine. Replacing with correct fix.

PR #851, opened 2026-03-28
Add Zed agent support via ACP proxy

### LINKED ISSUE #675: Add support to Zed Agent
The Zed editor has [an agent](https://zed.dev/agentic), it would be rad if git-ai supported it out of box. I don't know, however, if Zed currently offers any API to make this possible, like hooks and whatnot.

I previously created a discussion about Zed itself adding support to save AI threads in a repo, zed-industries/zed#45334 (but it didn't gain any traction). But I think something like git-ai receiving threads from Zed would work better.

Krishnachaitanyakc:
## Summary

Closes #675

- Implements automatic checkpoint tracking for Zed via ACP (Agent Client Protocol) proxy
- The proxy wraps a real agent (e.g., Claude Code), intercepting `tool_call`/`tool_call_update` notifications to create checkpoints automatically — no model cooperation needed
- Architecture: `Zed (ACP Client) ←stdio→ git-ai acp-proxy ←stdio→ Real Agent`

### New files
- `src/commands/acp_proxy.rs` — Core ACP proxy with Content-Length framed I/O, two-thread forwarding, and background checkpoint spawning
- `src/commands/checkpoint_agent/zed_preset.rs` — Preset for parsing hook input JSON from the proxy
- `src/mdm/agents/zed.rs` — `ZedInstaller` that detects Zed and prints `agent_servers` setup instructions

### Modified files
- `src/commands/git_ai_handlers.rs` — Wire up `acp-proxy` command and `zed` checkpoint preset
- Module declarations in `mod.rs` files

## Test plan

- [x] `cargo fmt -- --check` passes
- [x] `cargo build` compiles with no warnings
- [x] 16 unit tests for `acp_proxy.rs` (message framing, file path extraction, tool call tracking, arg parsing)
- [x] 5 unit tests for `zed_preset.rs` (pre/post tool use, error cases)
- [x] 4 unit tests for `zed.rs` (name, id, config hooks flag, install instructions)
- [ ] Manual: `echo '{"hook_event_name":"PostToolUse","session_id":"s1","cwd":"/tmp","file_paths":["a.rs"]}' | git-ai checkpoint zed --hook-input stdin`
- [ ] Manual: `git-ai acp-proxy -- echo "test"` (verifies arg parsing and child spawn)

---

<a href="https://app.devin.ai/review/git-ai-project/git-ai/pull/851" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/src/commands/acp_proxy.rs
+++ b/src/commands/acp_proxy.rs
@@ -410,7 +410,7 @@ pub fn handle_acp_proxy(args: &[String]) {
 
     // Thread 1: Zed (our stdin) → Agent (child stdin)
     // Forward messages without inspection
-    let zed_to_agent = thread::spawn(move || {
+    let _zed_to_agent = thread::spawn(move || {
         let mut reader = BufReader::new(io::stdin().lock());
         let mut writer = child_stdin;
 
@@ -464,12 +464,14 @@ pub fn handle_acp_proxy(args: &[String]) {
         std::process::exit(1);
     });
 
-    // Wait for forwarding threads
-    let _ = zed_to_agent.join();
-    let _ = agent_to_zed.join();
-
-    // Exit with the same code as the agent
-    std::process::exit(exit_status.code().unwrap_or(1));
+    // Exit immediately with the agent's exit code.
+    // The zed_to_agent thread is blocked on stdin read and cannot be unblocked
+    // without closing the process's stdin (which we don't own). Calling
+    // process::exit here is safe because neither forwarding thread holds
+    // resources that require explicit cleanup.
+    let code = exit_status.code().unwrap_or(1);
+    let _ = agent_to_zed.join(); // agent stdout is closed, so this returns quickly
+    std::process::exit(code);
 }
 
 #[cfg(test)]
--- a/src/commands/acp_proxy.rs
+++ b/src/commands/acp_proxy.rs
@@ -36,10 +36,10 @@ fn read_message(reader: &mut impl BufRead) -> io::Result<Option<String>> {
         }
         // Content-Length header is case-insensitive per LSP/JSON-RPC spec
         let lower = trimmed.to_ascii_lowercase();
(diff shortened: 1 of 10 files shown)


dlight:
Thank you for addressing that issue!

I don't know much about how git-ai works, but I'm surprised this approach is needed. I mean, if I use an external agent in Zed like Claude Code, shouldn't Claude Code hooks like pre-edit fire regardless of whether it is being driven by Zed or not? And most external agents Zed could use probably have git-ai support already.

Also I have a question. This proxy won't track edits of the first-party Zed Agent right? 
I'm not entirely sure, but I think that in Zed ACP is used for [external agents](https://zed.dev/docs/ai/external-agents) only.

(the Zed Agent is the default agent interface when you open the agent panel, to use external agents [you need to select them](https://zed.dev/docs/ai/agent-panel#new-thread) when creating a new thread).

If this doesn't track edits from the first-party Zed Agent, I think this shouldn't close #675.

Krishnachaitanyakc:
Good points! You're right that Claude Code's hooks fire independently — running it through the ACP proxy would double-track edits. I'm adding a fix: the proxy sets `GIT_AI_ACP_PROXY=1` on the child process, and the Claude preset skips when that env var is present (same pattern as the existing VS Code Copilot/Cursor filters).

The proxy is still useful for agents that don't have native git-ai integration.

Re: first-party Zed Agent — correct, ACP `agent_servers` is for external agents only. Zed doesn't expose hooks or events for its built-in agent's tool use, so there's no way to intercept those edits currently. I'll file a feature request with Zed for tool-use hooks. That said, this is a Zed-side limitation rather than a gap in this PR's approach.

svarlamov:
Closing re duplication of hooks for existing agents via ACP etc: https://github.com/git-ai-project/git-ai/issues/1239#issuecomment-4368160167

Thanks again for the contribution @Krishnachaitanyakc and we'll catch up soon

PR #1017, opened 2026-04-07
prompt-event: add tool/model/timing/text fields

shortdiv:
## Summary
- Extends `PromptEventValues` with sparse positional fields needed for the Sessions UI: `tool_call_name`, `model`, `error`, `start_ts`, `reported_duration_ms`, `observed_duration_ms`, `input_hash`, `input_text`.
- Stacked on top of #852 (`feat/prompt-event-metric`) — targets that branch, not main.

## Why
The base PR introduces the `PromptEvent` metric (event id 5) but only carries the core identification fields (kind, event_id, parent). The downstream Sessions dashboard in the monorepo needs:
- `model` on AiMessage / ThinkingMessage to attribute turns
- `tool_call_name` + `error` on ToolCall for tool usage / friction analysis
- `start_ts` / `*_duration_ms` for session timing
- `input_text` (truncated) on HumanMessage / AiMessage for friction regex + session titles
- `input_hash` for dedupe across re-emits

## Test plan
- [ ] cargo check / cargo test on the git-ai crate
- [ ] napi build picks up the new fields and round-trips them through `parseMetricsBatch`
- [ ] End-to-end: emit a real prompt with claude-code, verify all new fields populate in the local ClickHouse `prompt_events` table

---

<a href="https://app.devin.ai/review/git-ai-project/git-ai/pull/1017" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/src/commands/prompt_event.rs
+++ b/src/commands/prompt_event.rs
@@ -19,6 +19,48 @@ pub mod prompt_event_kind {
     pub const THINKING_MESSAGE: &str = "ThinkingMessage";
     pub const TOOL_CALL: &str = "ToolCall";
     pub const FILE_WRITE: &str = "FileWrite";
+    pub const SKILL_INVOCATION: &str = "SkillInvocation";
+    pub const MCP_CALL: &str = "McpCall";
+}
+
+/// Maximum length for `input_text` payloads (HumanMessage/AiMessage only).
+const INPUT_TEXT_MAX_BYTES: usize = 1024;
+
+/// Compute the 16-char content hash always stored on a PromptEvent.
+fn compute_input_hash(content: &str) -> String {
+    let mut hasher = Sha256::new();
+    hasher.update(content.as_bytes());
+    let digest = hasher.finalize();
+    format!("{:x}", digest)[..16].to_string()
+}
+
+/// Truncate text on a UTF-8 boundary, used only for `input_text` payloads.
+fn truncate_for_text(s: &str) -> String {
+    if s.len() <= INPUT_TEXT_MAX_BYTES {
+        return s.to_string();
+    }
+    let safe_end = s.floor_char_boundary(INPUT_TEXT_MAX_BYTES);
+    s[..safe_end].to_string()
+}
+
+/// Parse an ISO-8601 timestamp string into unix milliseconds.
+fn parse_iso_to_unix_ms(ts: &str) -> Option<u64> {
+    chrono::DateTime::parse_from_rfc3339(ts)
+        .ok()
+        .map(|dt| dt.timestamp_millis() as u64)
+}
+
+/// Classify a Claude tool_use into the right PromptEvent kind.
+fn kind_for_tool(tool_name: &str) -> &'static str {
+    if tool_name.starts_with("mcp__") {
+        prompt_event_kind::MCP_CALL
+    } else if matches!(tool_name, "Write" | "Edit" | "MultiEdit") {
+        prompt_event_kind::FILE_WRITE
+    } else if tool_name == "Skill" || tool_name == "SlashCommand" {
+        prompt_event_kind::SKILL_INVOCATION
+    } else {
+        prompt_event_kind::TOOL_CALL
+    }
 }
 
 /// Apply parent_id and parent_id_estimated to a PromptEventValues builder.
@@ -40,9 +82,16 @@ struct TranscriptEvent {
     kind: String,
     /// Content used for stable ID generation.
     content_hash_input: String,
-    /// Timestamp of the event, if available.
-    #[allow(dead_code)]
+    /// Wall-clock timestamp of the event, if available.
     timestamp: Option<String>,
+    /// Tool/skill/mcp server name (for tool_use kinds).
+    tool_call_name: Option<String>,
+    /// Model id (for AiMessage/ThinkingMessage).
+    model: Option<String>,
+    /// SHA256[..16] of the normalized payload — always populated.
+    input_hash: String,
+    /// Truncated message text — only set for HumanMessage/AiMessage.
+    input_text: Option<String>,
 }
 
 /// State file for tracking emitted events per session.
@@ -133,10 +182,15 @@ fn parse_claude_transcript(transcript_path: &str) -> Result<Vec<TranscriptEvent>
                         kind: prompt_event_kind::HUMAN_MESSAGE.to_string(),
                         content_hash_input: truncate_for_hash(&text),
                         timestamp,
+                        tool_call_name: None,
+                        model: None,
+                        input_hash: compute_input_hash(&text),
+                        input_text: Some(truncate_for_text(&text)),
                     });
                 }
             }
             Some("assistant") => {
+                let model = entry["message"]["model"].as_str().map(|s| s.to_string());
                 if let Some(content_array) = entry["message"]["content"].as_array() {
                     for item in content_array {
                         match item["type"].as_str() {
@@ -148,6 +202,10 @@ fn parse_claude_transcript(transcript_path: &str) -> Result<Vec<TranscriptEvent>
                                         kind: prompt_event_kind::THINKING_MESSAGE.to_string(),
                                         content_hash_input: truncate_for_hash(thinking),
                                         timestamp: timestamp.clone(),
+                                        tool_call_name: None,
+                                        model: model.clone(),
+                                        input_hash: compute_input_hash(thinking),
+                                        input_text: None,
                                     });
                                 }
                             }
@@ -159,6 +217,10 @@ fn parse_claude_transcript(transcript_path: &str) -> Result<Vec<TranscriptEvent>
                                         kind: prompt_event_kind::AI_MESSAGE.to_string(),
                                         content_hash_input: truncate_for_hash(text),
                                         timestamp: timestamp.clone(),
+                                        tool_call_name: None,
+                                        model: model.clone(),
+                                        input_hash: compute_input_hash(text),
+                                        input_text: Some(truncate_for_text(text)),
                                     });
                                 }
                             }
@@ -166,16 +228,16 @@ fn parse_claude_transcript(transcript_path: &str) -> Result<Vec<TranscriptEvent>
                                 let tool_name =
                                     item["name"].as_str().unwrap_or("unknown").to_string();
                                 let tool_id = item["id"].as_str().unwrap_or("").to_string();
-                                let is_file_write =
-                                    matches!(tool_name.as_str(), "Write" | "Edit" | "MultiEdit");
+                                let kind = kind_for_tool(&tool_name).to_string();
+                                let hash_input = format!("{}:{}", tool_name, tool_id);
                                 events.push(TranscriptEvent {
-                                    kind: if is_file_write {
-                                        prompt_event_kind::FILE_WRITE.to_string()
-                                    } else {
-                                        prompt_event_kind::TOOL_CALL.to_string()
-                                    },
-                                    content_hash_input: format!("{}:{}", tool_name, tool_id),
+                                    kind,
+                                    content_hash_input: hash_input.clone(),
                                     timestamp: timestamp.clone(),
+                                    tool_call_name: Some(tool_name),
+                                    model: model.clone(),
+                                    input_hash: compute_input_hash(&hash_input),
+                                    input_text: None,
                                 });
                             }
                             _ => {}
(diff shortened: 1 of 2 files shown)


shortdiv:
## Proposal: add skill file metadata to `SkillInvocation` events

Context: we're building a skills leaderboard in the monorepo Sessions UI and the three things we can't derive from invocation data alone are **"when was this skill last edited?"**, **"when was it first authored?"**, and **"did two sessions invoke the same version of the skill?"**. All three are cheap to get if the emitter stats the SKILL.md file at invocation time.

I'd like to add three sparse fields to `PromptEventValues`, populated only when `kind = SkillInvocation`:

```rust
pub mod prompt_event_pos {
    // ... existing 0..11 ...
    pub const SKILL_FILE_PATH: usize = 12;   // String  - absolute path to SKILL.md
    pub const SKILL_FILE_MTIME: usize = 13;  // u64     - file mtime (unix ms)
    pub const SKILL_FILE_HASH: usize = 14;   // String  - sha256[..16] of file contents
}
```

### Why denormalized on the event, not a separate `skills` table

The invocation log already *is* the usage data. The leaderboard query becomes:

```sql
SELECT
  skill_file_path                AS path,
  anyLast(tool_call_name)        AS name,
  MIN(start_ts)                  AS first_seen,
  MAX(skill_file_mtime)          AS last_edited,
  countDistinct(prompt_id)       AS sessions,
  countDistinct(human_author)    AS distinct_users,
  countDistinct(skill_file_hash) AS distinct_versions
FROM prompt_events
WHERE kind = 'SkillInvocation'
  AND org_id = ?
GROUP BY skill_file_path
```

No second table, no upsert logic, no GC. A few hundred redundant bytes per invocation — negligible at our scale. If we later need \`creator\` (from \`git blame\`) or \`description\` (from frontmatter), *that's* when a cached \`skills\` entity table earns its keep. Invocation-level data shouldn't go there.

### Where the path comes from

In \`parse_claude_transcript\`, the skill case currently lands in the \`tool_use\` branch once \`kind_for_tool\` returns \`SKILL_INVOCATION\`. The tool_use payload for Claude's \`Skill\` / \`SlashCommand\` tools contains the skill name in \`item[\"input\"]\`. We resolve that to a filesystem path by searching the standard Claude skill locations in order:

1. \`<cwd>/.claude/skills/<name>/SKILL.md\`
2. \`<cwd>/.claude/skills/<name>.md\`
3. \`\$HOME/.claude/skills/<name>/SKILL.md\`
4. \`\$HOME/.claude/skills/<name>.md\`

First hit wins. If none exist (shouldn't happen in practice, but), we emit the event with the other fields and leave all three new fields null — sparse encoding handles it.

### Wiring sketch

```rust
// new helper in src/commands/prompt_event.rs
struct SkillFileMeta {
    path: String,
    mtime_ms: u64,
    hash: String,
}

fn resolve_skill_file(skill_name: &str, cwd: &Path) -> Option<SkillFileMeta> {
    let candidates = [
        cwd.join(\".claude/skills\").join(skill_name).join(\"SKILL.md\"),
        cwd.join(\".claude/skills\").join(format!(\"{skill_name}.md\")),
        dirs::home_dir()?.join(\".claude/skills\").join(skill_name).join(\"SKILL.md\"),
        dirs::home_dir()?.join(\".claude/skills\").join(format!(\"{skill_name}.md\")),
    ];
    for path in candidates {
        let Ok(meta) = std::fs::metadata(&path) else { continue };
        let Ok(mtime) = meta.modified() else { continue };
        let mtime_ms = mtime
            .duration_since(std::time::UNIX_EPOCH)
            .ok()?
            .as_millis() as u64;
        let bytes = std::fs::read(&path).ok()?;
        let mut hasher = Sha256::new();
        hasher.update(&bytes);
        let hash = format!(\"{:x}\", hasher.finalize())[..16].to_string();
        return Some(SkillFileMeta {
            path: path.to_string_lossy().into_owned(),
            mtime_ms,
            hash,
        });
    }
    None
}
```

Then at the \`tool_use\` arm in \`parse_claude_transcript\` (and symmetrically in \`emit_single_event_from_hook\` for the PreToolUse path):

```rust
let skill_file = if kind == prompt_event_kind::SKILL_INVOCATION {
    item[\"input\"][\"skill_name\"]
        .as_str()
        .or_else(|| item[\"input\"][\"command\"].as_str())
        .and_then(|name| resolve_skill_file(name.trim_start_matches('/'), &cwd))
} else {
    None
};
```

And on the \`PromptEventValues\` builder:

```rust
if let Some(meta) = &evt.skill_file {
    values = values
        .skill_file_path(&meta.path)
        .skill_file_mtime(meta.mtime_ms)
        .skill_file_hash(&meta.hash);
}
```

### Scope note

This is **Claude-only** in the first cut. Cursor / Codex / etc. handle skills (or their equivalents) differently — some don't have a SKILL.md-on-disk model at all. For those agents we'd leave the fields null and the leaderboard would just show them as \"unresolved.\" I think that's the right default: don't pretend we have data we don't, and the UI can render \"—\" for unresolved rows.

Happy to open a follow-up PR on top of this one if you agree with the shape.

shortdiv:
## Proposal: capture human interruptions as a first-class event

Follow-up to the skill_file_* proposal above. While building the sessions UI I kept wanting to answer "did the human stop the model mid-flight?" and realized we have no signal for it today — an interrupted run just looks like a short AI turn followed by another HumanMessage, which is indistinguishable from a normal follow-up.

Interruptions are a high-value friction signal: unlike a correction ("that's wrong, do X instead"), an interruption means the human didn't even let the model finish. It's the strongest "I don't trust where this is going" signal we can capture cheaply.

### Proposed change

Add a new event kind and a couple of fields:

\`\`\`rust
pub mod prompt_event_kind {
    // ...existing...
    pub const INTERRUPTED: &str = "Interrupted";
}
\`\`\`

And extend the sparse positional encoding:

\`\`\`rust
pub mod prompt_event_pos {
    // ...existing 0..11...
    pub const INTERRUPT_REASON: usize = 12; // String - "user_cancel" | "tool_timeout" | "stop_hook" | "unknown"
    pub const INTERRUPTED_EVENT_ID: usize = 13; // String - event_id of the in-flight event that was cut off
}
\`\`\`

The \`Interrupted\` event would always be a child of the event it interrupted (via \`parent_id\`), and \`interrupted_event_id\` would point to the same target — redundant with parent_id in most cases, but explicit enough that queries don't have to infer it.

### Where it gets emitted

For Claude Code, three signals map to this cleanly:

1. **User cancel (Ctrl+C / Esc)** — Claude Code transcripts mark the aborted assistant turn with \`\"stop_reason\": \"user_cancel\"\` (or similar; worth confirming against a real transcript). In \`parse_claude_transcript\`, when we see a tool_use or assistant message with that stop_reason, emit an \`Interrupted\` event as a child of that turn.

2. **Stop hook firing mid-tool** — The \`Stop\` hook in Claude Code fires when the user kills an in-progress session. In \`emit_single_event_from_hook\`, if \`hook_event_name == \"Stop\"\` and there's an in-flight tool_use in state, emit \`Interrupted\` with reason \`stop_hook\`.

3. **Tool timeout** — \`PostToolUse\` with a timeout error. Separate from the error bool because a timeout is a kind of interruption (the runtime killed it), not an application-level failure.

### Queries this unlocks

\`\`\`sql
-- Interruption rate per session
SELECT
  prompt_id,
  countIf(kind = 'Interrupted') AS interruptions,
  countIf(kind = 'AiMessage') AS ai_turns,
  interruptions / nullIf(ai_turns, 0) AS interrupt_rate
FROM prompt_events
GROUP BY prompt_id

-- Tools most likely to get killed mid-flight
SELECT
  (SELECT tool_call_name FROM prompt_events WHERE event_id = i.interrupted_event_id) AS tool,
  count() AS kills
FROM prompt_events i
WHERE kind = 'Interrupted'
GROUP BY tool
ORDER BY kills DESC
\`\`\`

The second query in particular would tell us things like "Bash calls get interrupted 3x more than Edits" which is a useful reliability signal for the agent itself.

### Open questions

1. **Should \`interrupted\` be a bool on the parent event instead of a separate kind?** Pro: simpler schema, no new kind. Con: loses the timestamp of *when* the interrupt happened (useful for "did they wait 2s or 2min?"), and conflates "event" with "event status."  I lean toward separate kind for the timeline clarity, but open to the other framing.

2. **Is it worth capturing \`partial_output_bytes\`?** i.e. how much the model had produced before the kill. Would let us distinguish "killed before it even started" from "killed after 2KB of wrong answer." Probably out of scope for this PR but worth a TODO.

3. **Does Codex / Cursor expose equivalent signals in their transcript formats?** The Claude side is clear, but if other agents don't have a stop_reason we may need to fall back on the Stop hook universally.

Happy to open this as a separate follow-up PR if it's too much for #1017. Mostly wanted to flag it now so the positional encoding has room reserved.

shortdiv:
### Addendum: daemon telemetry context

After digging into the daemon code I wanted to add context I didn't have when writing the proposal above — it answers open question #1 and tightens the server-side reaper story.

**How telemetry flows today:**
- Wrapper hooks are fire-and-forget over a Unix control socket to the long-lived daemon (`ensure_daemon_running` in `src/commands/daemon.rs:195`, `daemon_is_up` at `:237`).
- The daemon buffers envelopes in memory and flushes **every 3 seconds** via `telemetry_flush_loop` (`src/daemon/telemetry_worker.rs:19`, `FLUSH_INTERVAL`).
- `PromptEvent` (metric id 5) rides the same path → `POST /worker/metrics/upload`, batches ≤250, SQLite fallback on failure (`src/api/metrics.rs:48`).
- The only periodic daemon work is `daemon_update_check_loop` at `src/daemon.rs:7409`, which runs **hourly** and does GC + version check + uptime cap (`DAEMON_MAX_UPTIME_SECS = 24h30m`). It emits nothing session-shaped.
- The closest thing to session state today is `WrapperPreState` / `WrapperPostState` (`src/daemon.rs:6826`), keyed by `invocation_id`, scoped to a **single git command invocation** (not an AI prompt session), GC'd after ~1h.

**There is no current daemon heartbeat.** I grepped the same terms and confirmed. Socket reachability is the only liveness notion, and it only gates wrapper → daemon RPC, not session → server.

**What this means for the proposal:**

The good news — **the 3s flush interval means we don't need a dedicated heartbeat**. Any `HumanMessage` / `ToolCall` / `PreToolUse` already reaches the server within ~3s of being emitted, so `max(event_ts) per session_id` on the server side is a tight enough liveness proxy for the 60s / 10min thresholds I proposed. The "any existing event implicitly updates `last_seen_at`" line in the proposal is load-bearing and the 3s flush makes it actually true.

The bad news — **hard wrapper death is invisible to the daemon too**. If the wrapper process is `kill -9`'d, terminal closed, laptop lid slammed, OOM killed, etc., the SIGTERM/SIGINT handler I proposed never runs. Nothing reaches the socket. From the daemon's perspective there's just… silence, and the daemon itself has no per-session state to notice the silence from (remember, `WrapperPreState` is keyed by git invocation, not prompt session). This means:

- **Client-side `session_end(user_quit)` is best-effort** — we should not treat its absence as a signal of anything.
- **The server-side reaper is the only reliable source of truth for termination.** It must always run, even if the daemon is "well-behaved," because the daemon fundamentally can't see its own death.
- The 10min reap threshold in the proposal is doing real work, not just catching edge cases.

**Small refinement to the proposal:**

The daemon could cheaply help the reaper by adding a second in-flight map, parallel to `WrapperPreState`, keyed by `session_id`, updated on every outbound telemetry batch. Then in the hourly GC loop (or a faster 60s one — we'd need to add it), any session_id whose `last_flush_ts` is older than the idle threshold gets an `session_end(idle_timeout)` synthesized and emitted on the next flush. This gives us a *second* source of end events that doesn't require the wrapper to be alive — just the daemon. The server-side reaper stays as the final backstop for when the daemon itself dies.

So the layered story becomes:

1. **Wrapper alive, clean exit** → SIGTERM handler emits `session_end(user_quit)`. Arrives within 3s.
2. **Wrapper dies hard, daemon alive** → daemon's own idle sweep synthesizes `session_end(idle_timeout)` after 30min. Arrives on next flush.
3. **Daemon also dead** → server-side reaper synthesizes `session_end(signal)` after 10min of silence.

Three independent mechanisms, each catching what the layer above can't. The daemon-side sweep is optional for the MVP (rollout step 1 still works with just 1 and 3), but worth designing space for now.

**Refined answer to open question #3** ("what counts as a user-facing event for the idle timer"): given the 3s flush, we can afford to be stricter than I initially thought. `HumanMessage` and `ToolCall` clearly yes. `AiMessage` I'd now say **no** — if the agent is talking to itself with no human input for 30min, that *is* an idle session from the user's perspective, regardless of AI activity. This matches the intuition that "idle" = "human walked away," not "nothing is happening."

svarlamov:
Superseded by #1153

PR #1045, opened 2026-04-10
Add trigger and workdir fields to post_notes_updated hook payload 

cnhnwt:
Summary                                                                                                                                                                                                  
                                                                                                                                                                                                           
  This PR extends the post_notes_updated hook mechanism to provide richer context to hook scripts.

  Changes

  - Added a trigger parameter to post_notes_updated and post_notes_updated_single, allowing callers to identify which code path fired the hook.
  - Added workdir to the RepoHookContext struct, populated from the repository's working directory via repo.workdir().
  - Both trigger and workdir are now included in each entry of the JSON payload sent to configured git_ai_hooks.post_notes_updated shell commands.
  - Updated all three call sites in refs.rs with descriptive trigger values: notes_add, notes_add_batch, and notes_add_blob_batch.

  Motivation

  Hook scripts previously had no way to distinguish which operation triggered the hook, nor did they receive the local repository path. This made it difficult to implement integrations that need to
  report events to external services with full context.

  Payload example

  {
    "commit_sha": "abc123",
    "repo_url": "git@github.com:org/repo.git",
    "repo_name": "repo",
    "branch": "main",
    "is_default_branch": true,
    "note_content": "...",
    "trigger": "notes_add",
    "workdir": "/home/user/projects/my-repo"
  }

<!-- devin-review-badge-begin -->

---

<a href="https://app.devin.ai/review/git-ai-project/git-ai/pull/1045" target="_blank">
  <picture>
    <img src="https://static.devin.ai/assets/gh-open-in-devin-review-light.svg?v=1" alt="Open with Devin">
  </picture>
</a>
<!-- devin-review-badge-end -->

### DIFF
--- a/src/authorship/git_ai_hooks.rs
+++ b/src/authorship/git_ai_hooks.rs
@@ -26,7 +26,7 @@ struct RepoHookContext {
 /// The hook input is always passed through stdin as a JSON array of 1..N note entries.
 /// Commands are started in parallel, and we wait up to 3 seconds for completion before
 /// detaching and continuing so git-ai does not block.
-pub fn post_notes_updated(repo: &Repository, notes: &[(String, String)], trigger: &str) {
+pub fn post_notes_updated(repo: &Repository, notes: &[(String, String)]) {
     if notes.is_empty() {
         return;
     }
@@ -45,7 +45,6 @@ pub fn post_notes_updated(repo: &Repository, notes: &[(String, String)], trigger
     let branch = context.branch;
     let is_default_branch = context.is_default_branch;
     let workdir = context.workdir;
-    let trigger = trigger.to_string();
     let payload = notes
         .iter()
         .map(|(commit_sha, note_content)| {
@@ -56,7 +55,6 @@ pub fn post_notes_updated(repo: &Repository, notes: &[(String, String)], trigger
                 "branch": branch.as_str(),
                 "is_default_branch": is_default_branch,
                 "note_content": note_content,
-                "trigger": trigger.as_str(),
                 "workdir": workdir.as_str(),
             })
         })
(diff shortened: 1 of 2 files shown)


svarlamov:
@cnhnwt I think we could add the workdir, but what's that used for? Would it be more useful to get the git dir?

Also, can you explain the rationale behind receiving the name of the function (ex: `notes_add_blob_batch`)? This is an internal implementation detail of Git AI that is likely to change as Git AI evolves.

cnhnwt:
@svarlamov You are right. The workdir is indeed the local Git project directory. Since the hook is executed via a child thread, external programs can no longer access the raw data for processing. I added workdir and trigger to improve extensibility and provide this missing context. Specifically, passing the trigger makes it easier to identify the exact execution scenario of the hook.

cnhnwt:
> @cnhnwt I think we could add the workdir, but what's that used for? Would it be more useful to get the git dir?
> 
> Also, can you explain the rationale behind receiving the name of the function (ex: `notes_add_blob_batch`)? This is an internal implementation detail of Git AI that is likely to change as Git AI evolves.

@svarlamov Could you please review or implement this related request? We really need this functionality. Thank you so much!

svarlamov:
@cnhnwt I think this looks fine, but can you remove the `trigger` field before I get this through? This isn't something I think should be user-facing because it's just an internal implementation detail. It doesn't actually imply anything about the operations themselves.

cnhnwt:
> @cnhnwt I think this looks fine, but can you remove the `trigger` field before I get this through? This isn't something I think should be user-facing because it's just an internal implementation detail. It doesn't actually imply anything about the operations themselves.

@svarlamov I've removed the trigger field as requested.Please take a look.

PR #1060, opened 2026-04-11
perf(daemon): per-root DashMap sharding, normalizers, and bounded connection pool

svarlamov:
## Summary

Two architectural improvements that make the trace2 ingest pipeline scale with concurrent workloads.

### 1. Shard `TraceIngressState` with DashMap

Replace the single `Mutex<TraceIngressState>` (15 `HashMap`/`HashSet` fields all keyed by root SID) with 15 independent `dashmap::DashMap`/`DashSet` fields on `ActorDaemonCoordinator`:

- Per-root events no longer serialize on a global mutex — different roots run concurrently
- The **readonly fast path** (the 40/sec Zed flood) is now **lock-free**: `DashSet::contains` replaces `Mutex::lock + HashSet::contains`
- Fold `mark_trace_root_activity` (previously an extra Mutex cycle) into the existing lock acquisition for mutating events — saves one lock cycle per `git commit`
- `DashSet::insert` returning `bool` provides atomic check-and-set for the close-fallback deduplication logic

### 2. Per-root normalizer instances

Replace `AsyncMutex<TraceNormalizer>` (one global instance) with `DashMap<String, Arc<AsyncMutex<TraceNormalizer>>>`:

- Each root SID gets its own normalizer, scoped to its lifetime
- Normalizer entries are cleaned up in `clear_trace_root_tracking`
- Completed root SIDs tracked in a bounded `DashSet` (16,384 FIFO) to prevent orphaned normalizer creation from late events
- Early-exit guard for empty root keys prevents orphaned normalizer entries
- Architectural foundation for future parallel ingest workers per root

### Correctness fixes

- **TOCTOU race in `record_trace_connection_close`**: Use `remove_if_mut` to atomically decrement-and-remove within a single shard lock, with explicit `v > 0` guard against double-close edge case
- **TOCTOU race on `ingr_root_mutating` / `ingr_root_target_repo_only`**: Use DashMap entry API for atomic read-modify-write so concurrent threads cannot overwrite `true` with `false`
- **Normalizer leak guard**: `completed_root_sids` DashSet prevents late events from creating orphaned normalizers after root tracking is cleared
- **Documented benign race**: `record_trace_connection_close` can produce duplicate stale roots under concurrent close; downstream `ingr_root_close_fallback_enqueued` DashSet deduplicates

## Test plan

- [x] `cargo test --lib` — all tests pass
- [x] `cargo clippy --all-targets` — clean
- [x] `cargo fmt` — applied
- [x] CI ubuntu checks — all green

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -631,6 +631,20 @@ dependencies = [
 ]
 
 [[package]]
+name = "dashmap"
+version = "6.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5041cc499144891f3790297212f32a74fb938e5136a14943f338ef9e0ae276cf"
+dependencies = [
+ "cfg-if",
+ "crossbeam-utils",
+ "hashbrown 0.14.5",
+ "lock_api",
+ "once_cell",
+ "parking_lot_core",
+]
+
+[[package]]
 name = "dbus"
 version = "0.9.10"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1104,6 +1118,7 @@ dependencies = [
  "chrono",
  "clap",
  "crossterm",
+ "dashmap",
  "dirs",
  "envy",
  "filetime",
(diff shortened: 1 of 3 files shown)


svarlamov:
Just fyi maybe some useful ideas, but this is old so I'm just gonna close it out for now. CC @heapwolf

PR #1064, opened 2026-04-12
fix(daemon): eliminate flakiness in daemon and wrapper-daemon test suites

svarlamov:
## Summary

Three root causes of daemon/wrapper-daemon test flakiness fixed:

- **Snapshot timestamp bug**: `payload_timestamp_ns` in `trace_normalizer.rs` was calling `Value::as_u64` on git trace2's `time` field, which is an RFC3339 string. It fell through to `t_abs` (elapsed seconds since program start, ~2–3ms), making `finished_at_ns` nearly zero. The mtime-cutoff filter then treated every file as "modified after cutoff", producing empty carryover snapshots and silently dropping amend-authorship working-log data. Fix: parse `time` as RFC3339 via chrono.

- **Missing pending-root slot for `current_dir` invocations**: Commands run via `current_dir` (not `-C`) have no `-C` in argv, so the git trace2 `start` event carries `cwd=None` → no worktree hint → no `PendingRoot` slot → command was processed via the no-op "applied" path with no deferred snapshot capture. Subdirectory amend tests failed because the carryover snapshot was never captured. Fix: when `replace_pending_root_entry` finds no slot but the command has a resolved `family_key`, insert it directly as a `ReadyCommand` (late registration).

- **Async drain races captured-checkpoint tests**: The async per-family drain workers made captured checkpoints (delegated with `wait=false`) truly fire-and-forget. Tests that called `git_ai checkpoint mock_ai <files>` and immediately read `current_working_logs()` raced against the daemon. The production code path is correct. Fix in the test harness only: `git_ai_with_env` and `git_ai_with_stdin` now capture the daemon's total completion-log baseline before running a checkpoint, then wait for it to advance if the output contains "Checkpoint queued" (the existing production marker for async-delegated checkpoints).

## Test plan

- [x] All 2937 integration tests pass at 4, 8, 12, and 16 threads with 0 failures
- [x] Verified `status_ignore` and `subdirs` amend tests pass in daemon and wrapper-daemon modes
- [x] No production code paths changed by the test harness fix

---

<a href="https://app.devin.ai/review/git-ai-project/git-ai/pull/1064" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


svarlamov:
dead end

PR #1065, opened 2026-04-12
fix(rebase): prevent prompt session leak via no-trailing-newline Replace hunks

svarlamov:
## Summary

- Fixes a bug where after a slow-path rebase, commit 1′ would contain prompt records from commit 2′s session (prompt leak).
- Root cause: when a file has no trailing newline, git represents the last original line as *replaced* rather than untouched on the next append — turning an Insert into a Replace hunk. This misclassification caused (a) the line's attribution to be dropped by `apply_hunks_to_line_attributions` and (b) the appending commit's note to overwrite the earlier commit's attribution.
- Fix: detect content-identical leading lines in a Replace hunk (`count_content_preserving_lines`), preserve their attribution in `apply_hunks_to_line_attributions`, and skip the overlay in `try_reconstruct_attributions_from_notes_cached` only when the line is already claimed by an earlier commit.

Stacks on top of #1057.

## Test plan

- [ ] `test_rebase_each_commit_note_has_only_its_own_session_prompt` — new test; was failing before the fix, now passes
- [ ] `test_rebase_autosquash` — regression guard; the "already-attributed" check ensures the autosquash case (first commit legitimately owns line 1) continues to work
- [ ] Full integration suite: `cargo test --test integration` — 2945 tests pass, 0 fail

---

<a href="https://app.devin.ai/review/git-ai-project/git-ai/pull/1065" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/src/authorship/rebase_authorship.rs
+++ b/src/authorship/rebase_authorship.rs
@@ -803,12 +803,38 @@ fn try_reconstruct_attributions_from_notes_cached(
         }
 
         // Step 2: Overlay this commit's note attributions.
+        //
+        // Skip content-preserving lines when overlaying: when git's diff shows a Replace
+        // hunk for a line whose content is identical in old and new (e.g., appending to a
+        // file without a trailing newline makes the last original line appear "replaced"),
+        // the note attributes those "replaced" lines to this commit.  But those lines were
+        // actually written by an earlier commit — we must not let this commit overwrite them.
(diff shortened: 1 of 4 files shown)


svarlamov:
Rolling back: the no-trailing-newline Replace hunk behavior is correct git semantics — commit 2 modifies line 6 (adds a newline), so commit 2 owns it. The test was testing incorrect expected behavior.

PR #1120, opened 2026-04-15
fix: prevent extreme memory consumption when reading large OpenCode db

steventohme:
Replace N+1 SQLite query pattern with a single batch query for parts, and cap the SQLite page cache at ~2MB. Without indexes on session_id in the OpenCode DB, each per-message parts query triggered a full table scan — causing 10-18 GB RSS on databases with many sessions.

Bug report from Weave customer:

When running `git-ai checkpoint opencode --hook-input stdin`​, the process reads OpenCode's local SQLite database (​`~/.local/share/opencode/opencode.db`​) to extract conversation transcripts. Because OpenCode doesn't prune old sessions, this database grows continuously over time.
Once the database reaches a certain size (in my case, 318 MB across roughly 353 sessions), the checkpoint process attempts to load the entire database into memory. This causes the process to spiral to 10–18 GB of RSS and completely peg a CPU core.

---

<a href="https://app.devin.ai/review/git-ai-project/git-ai/pull/1120" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/src/commands/checkpoint_agent/opencode_preset.rs
+++ b/src/commands/checkpoint_agent/opencode_preset.rs
@@ -641,9 +641,7 @@ impl OpenCodePreset {
         let mut parts_by_message = Self::read_all_session_parts_from_sqlite(&conn, session_id)?;
 
         Self::build_transcript_from_messages(messages, |message_id| {
-            Ok(parts_by_message
-                .remove(message_id)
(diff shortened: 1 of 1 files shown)


svarlamov:
Thanks for the contribution @steventohme -- I'll test and merge for the next release

PR #1146, opened 2026-04-19
fix(jetbrains): prefer dataDirectoryName for plugin path

peckfly:
## Summary
- prefer `dataDirectoryName` from `product-info.json` when resolving JetBrains plugin directories
- keep the existing product/build-based path derivation as a fallback
- add tests covering both the preferred and fallback behaviors

## Why
Some JetBrains IDEs use a data directory name that does not match the path derived from `productCode + buildNumber`. In those cases, the plugin can be installed into the wrong directory.

## Testing
- `cargo test mdm::jetbrains::detection::tests:: -- --nocapture`
- `cargo fmt -- --check`

### DIFF
--- a/src/mdm/jetbrains/detection.rs
+++ b/src/mdm/jetbrains/detection.rs
@@ -130,11 +130,15 @@ fn detect_macos_ide(ide: &'static JetBrainsIde, app_path: &Path) -> Option<Detec
         return None;
     }
 
-    // Get build number from Info.plist or product-info.json
-    let (build_number, major_build) = get_macos_build_number(app_path);
+    // Get build number and data directory from Info.plist or product-info.json
+    let (build_number, major_build, data_directory_name) = get_macos_build_metadata(app_path);
 
     // Get plugins directory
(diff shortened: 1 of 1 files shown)


svarlamov:
Thanks for the contribution @peckfly! What behavior/bug is this fixing?

peckfly:
> Thanks for the contribution @peckfly! What behavior/bug is this fixing?

This fixes JetBrains plugin installation/detection on IDEs where the actual config/plugins directory is defined by `dataDirectoryName` in `product-info.json`, but the current code derives the directory from `productCode + buildNumber`.

For example, on IntelliJ IDEA 2026.1, `product-info.json` reports `dataDirectoryName=IntelliJIdea2026.1`, while the current logic derives `IU2026.1`. That causes git-ai to look in / install into the wrong plugins directory, so the plugin may appear missing or not get loaded by the IDE.

This change prefers `dataDirectoryName` when available, and falls back to the existing `productCode + buildNumber` behavior otherwise.

svarlamov:
Great context, thank you for the contribution! Looking forward to more! 🙏

PR #1151, opened 2026-04-20
fix(jetbrains): use Google plugin dir for Android Studio on macOS

peckfly:
## Summary
- fix macOS JetBrains plugin path resolution for Android Studio
- use the IDE's Google-scoped plugins directory instead of the generic JetBrains parent
- add a regression test for the Android Studio macOS path

## What bug does this fix?
Android Studio is already listed as a supported JetBrains IDE, but on macOS its plugins live under `~/Library/Application Support/Google/<dataDirectoryName>/plugins`.

The current code always uses `~/Library/Application Support/JetBrains/...`, so git-ai can install/check the plugin in the wrong directory for Android Studio. This means the plugin may appear missing or fail to load even though installation reported success.

This change keeps the existing behavior for other JetBrains IDEs and only switches Android Studio to the Google-scoped parent directory on macOS.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1147,6 +1147,7 @@ dependencies = [
  "ureq",
  "url",
  "uuid",
+ "winreg",
  "zip",
 ]
 
@@ -4232,6 +4233,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5a5364e9d77fcdeeaa6062ced926ee3381faa2ee02d3eb83a5c27a8825540829"
 dependencies = [
  "memchr",
+]
+
+[[package]]
+name = "winreg"
+version = "0.55.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cb5a765337c50e9ec252c2069be9bf91c7df47afb103b642ba3a53bf8101be97"
+dependencies = [
+ "cfg-if",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,6 +48,7 @@ openssl = { version = "0.10", features = ["vendored"] }
 
 [target.'cfg(windows)'.dependencies]
 named_pipe = "0.4.1"
+winreg = "0.55"
 
 [features]
 test-support = ["git2"]
--- a/src/mdm/jetbrains/detection.rs
+++ b/src/mdm/jetbrains/detection.rs
@@ -553,39 +553,39 @@ fn get_plugins_dir(
 
     #[cfg(target_os = "macos")]
     {
-        return plugins_dir_for_platform(
+        plugins_dir_for_platform(
             JetBrainsPlatform::Macos,
             &home,
             None,
             data_directory_name,
             product_code,
             build_number,
-        );
+        )
     }
 
     #[cfg(windows)]
     {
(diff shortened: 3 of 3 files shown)


svarlamov:
Hey @peckfly thanks for the contribution again! Do you know if we can apply the same to Linux and Windows https://github.com/git-ai-project/git-ai/pull/1151#discussion_r3108839131

peckfly:
> Hey @peckfly thanks for the contribution again! Do you know if we can apply the same to Linux and Windows [#1151 (comment)](https://github.com/git-ai-project/git-ai/pull/1151#discussion_r3108839131)

ok, i'll try it

svarlamov:
Thanks! Do we know if the Windows registry calls require admin access?

svarlamov:
## Question: Intentional Linux plugins path change?

The refactoring into `plugins_dir_for_platform` dropped the trailing `/plugins` segment for Linux. On `main`, all three platforms append `.join("plugins")` at the end:

```rust
// main branch — Linux path
home_dir()
    .join(".local").join("share").join("JetBrains")
    .join(&version_suffix)
    .join("plugins")
```

In this PR, the Linux arm no longer has it:

```rust
JetBrainsPlatform::Linux => home_dir
    .join(".local").join("share").join(parent_dir)
    .join(&version_suffix),
    // missing .join("plugins")
```

macOS and Windows both still have `.join("plugins")`. Was this intentional, or did it get lost during the refactor?

peckfly:
> Thanks! Do we know if the Windows registry calls require admin access?

I validated this under a real standard (non-admin) Windows user account on my machine, and the registry lookup worked without admin access.

The code only does read-only probes via `open_subkey()` / `get_value()`. If a key cannot be opened or read, we just skip that candidate and continue with the other detection paths instead of failing the install flow.

peckfly:
> ## Question: Intentional Linux plugins path change?
> The refactoring into `plugins_dir_for_platform` dropped the trailing `/plugins` segment for Linux. On `main`, all three platforms append `.join("plugins")` at the end:
> 
> ```rust
> // main branch — Linux path
> home_dir()
>     .join(".local").join("share").join("JetBrains")
>     .join(&version_suffix)
>     .join("plugins")
> ```
> 
> In this PR, the Linux arm no longer has it:
> 
> ```rust
> JetBrainsPlatform::Linux => home_dir
>     .join(".local").join("share").join(parent_dir)
>     .join(&version_suffix),
>     // missing .join("plugins")
> ```
> 
> macOS and Windows both still have `.join("plugins")`. Was this intentional, or did it get lost during the refactor?

Yes, the Linux path change was intentional; the trailing `/plugins` was not dropped accidentally during the refactor.

I validated this end-to-end in WSL Ubuntu on my Windows machine with a real Android Studio installation. `git-ai install-hooks --dry-run` detected Android Studio, `--dry-run=false` installed the plugin successfully, and a follow-up dry-run reported `Plugin already installed`.

I also verified the resulting path from Android Studio itself: its log shows
`idea.plugins.path=/home/.../.local/share/Google/AndroidStudio2025.3.4`
with no trailing `/plugins`, and the Git AI plugin files were actually written there.

So for Linux, removing the trailing `/plugins` here was intentional and matches the validated install flow.

image

svarlamov:
Got it -- helpful context. I need to test on Mac OS as well and will try to merge asap

PR #1180, opened 2026-04-22
Enforce prompt_storage policy on every notes_add rewrite path

victormier:
## Summary

- `src/authorship/post_commit.rs:164–223` correctly honors `effective_prompt_storage` (strip / redact / CAS) before `notes_add`. The **eight** `notes_add`/`notes_add_batch` callsites in `src/authorship/rebase_authorship.rs` (merge, rebase slow-path, cherry-pick, amend, and the three remap paths) bypassed this entirely — so on amend / rebase / cherry-pick / squash-merge, messages from upstream notes or from working-log checkpoints were serialized straight into git notes, breaking the documented \"default mode strips messages\" contract.
- Extracts the policy block into a new module `src/authorship/prompt_storage_policy.rs` with two helpers and wires it into every rewrite callsite.
- `post_commit.rs` behavior is unchanged; the rewrite paths now match it.
- Incidentally closes a secondary (security) gap: under `prompt_storage = \"notes\"`, the rewrite paths never ran `redact_secrets_from_prompts` either, so secrets could land in notes on any rebase/amend/etc.

## Observed symptom

- Spans every `git_ai_version` from 1.0.37 through 1.3.2 — not a regression in one release.
- **Zero** leaked prompts have `messages_url` set → none traversed the `Default` / CAS branch of the policy block.
- Same squash-merge commit carries a mix of stripped and populated prompts
- We saw this in commits with `accepted_lines > 0` and `total_additions: 0, total_deletions: 0`, which lines up with rebase/squash metadata-template assembly.

Exactly what you'd see if `rebase_authorship.rs` skipped the policy block — which, per the audit, it did.

## Callsite audit (`src/authorship/rebase_authorship.rs`)

| Line | Caller / context | Source of `AuthorshipLog` or note body | strip / redact before this PR? |
|---|---|---|---|
| 631 | Merge (empty-diff branch) → `build_metadata_only_authorship_log_from_source_notes` | Source notes | No |
| 722 | Merge (normal path) → `merged_va.to_authorship_log()` | `new_for_base_commit` → notes + blame | No |
| 1883 | Slow-path rebase batch | Cached per-file attestation text + metadata template | No |
| 2107 | `rewrite_authorship_after_cherry_pick` → `current_va.to_authorship_log()` | `new_for_base_commit` → notes | No |
| 2956 | `rewrite_authorship_after_commit_amend` → `working_va.to_authorship_log_and_initial_working_log()` | Working log (transcripts when present) | No |
| 3449 | `remap_notes_for_commit_pairs` (slow-path fallback) | Source note content, byte-copied | No |
| 3587 | Fast-path rebase note remap | Source note content, byte-copied | No |
| 3682 | Fast-path cherry-pick note remap | Source note blob, byte-copied | No |

The non-test callsite list is exhaustive; nothing else writes notes in this file. Fast-path remap paths need `apply_prompt_storage_policy_to_note` (parse → apply → reserialize) because they operate on raw note bytes. The cost is per-note JSON deserialize+serialize on rebase/cherry-pick — trading raw-byte speed for correctness parity with post_commit.

## Design notes

`apply_prompt_storage_policy(repo, &mut AuthorshipLog)` owns the full policy: custom_attributes injection + match on `PromptStorageMode::{Local, Notes, Default}` + CAS enqueue + strip-on-fallback. `post_commit.rs` now calls it once instead of inlining the block. The amend path also loses its duplicate custom_attributes injection since the helper covers it.

`apply_prompt_storage_policy_to_note(repo, &str)` is the byte-level variant for remap callers. Parse failure falls through unchanged (the downstream reader would fail either way; we don't want to silently lose data).

## Security consideration

Under `prompt_storage = \"notes\"`, `post_commit.rs` calls `redact_secrets_from_prompts` before writing the note. The rewrite paths never did — so any secrets detected at commit time were re-exposed on any subsequent rebase / amend / squash-merge (because that path produced a fresh note without redaction). This PR closes that gap incidentally: the same shared helper enforces redaction on every write.

## Test plan

`tests/integration/prompt_storage_rewrite_paths.rs`:

- [x] Control: `prompt_storage = \"local\"`, initial commit strips messages.
- [x] Amend under local: amended note strips messages (this was the direct reproduction of the leak with agent-v1 / non-refetch preset).
- [x] Rebase / cherry-pick / squash-merge under local, no pre-existing leak.
- [x] **Leaky-upstream propagation tests**: seed a source note with messages (simulating an older git-ai version or an unrelated leak path), then rebase / cherry-pick / squash-merge — new notes must strip. This is the production pattern.
- [x] `prompt_storage = \"notes\"`: initial commit *keeps* messages AND rebased commit *keeps* messages (policy honored both ways — no over-stripping).
- [x] Full suite green locally (`task test`): 1,475 lib + 3,017 integration + 33 + 56 + 17 + 8 + 7 others, zero failures, no snapshot updates.

---

<a href="https://app.devin.ai/review/git-ai-project/git-ai/pull/1180" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/src/authorship/mod.rs
+++ b/src/authorship/mod.rs
@@ -10,6 +10,7 @@ pub mod internal_db;
 pub mod move_detection;
 pub mod post_commit;
 pub mod pre_commit;
+pub mod prompt_storage_policy;
 pub mod prompt_utils;
 pub mod range_authorship;
 pub mod rebase_authorship;
--- a/src/authorship/post_commit.rs
+++ b/src/authorship/post_commit.rs
@@ -1,14 +1,12 @@
-use crate::api::{ApiClient, ApiContext};
 use crate::authorship::authorship_log_serialization::AuthorshipLog;
 use crate::authorship::ignore::{
     build_ignore_matcher, effective_ignore_patterns, should_ignore_file_with_matcher,
 };
 use crate::authorship::prompt_utils::{PromptUpdateResult, update_prompt_from_tool};
-use crate::authorship::secrets::{redact_secrets_from_prompts, strip_prompt_messages};
 use crate::authorship::stats::{stats_for_commit_stats, write_stats_to_terminal};
 use crate::authorship::virtual_attribution::VirtualAttributions;
 use crate::authorship::working_log::{Checkpoint, CheckpointKind, WorkingLogEntry};
-use crate::config::{Config, PromptStorageMode};
+use crate::config::Config;
 use crate::error::GitAiError;
 use crate::git::refs::notes_add;
 use crate::git::repository::Repository;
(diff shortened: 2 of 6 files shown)


svarlamov:
Hey @victormier thank you so much for the contribution and identifying all these issues! I will work to review asap and wanted to let you know that for new notes we're in the midst of an update that will completely remove the `messages` property.

I'm working to determine whether we can also just remove the `messages` field wholesale from all the old notes as well as (AFAIK) no one is really using the messages-in-notes feature given all the foot guns.

svarlamov:
Addressing via #1185 and sunsetting raw messages in prompts and future updates

Issue #237, opened 2025-11-24
VS Code remote/devcontainer/codespaces support for prompt and model data


svarlamov:
Since the copilot agent chat session data is stored in the workspace storage, we need to build special support for remote VSC and codespaces

In codespaces, these chat JSON logs are only available in the browser in the indexed DB-based virtual filesystem. In the remote VSC/local devcontainer environment, the files are stored on the host, but there's no way (as far as I know) to access the host filesystem from the extension. Needs more research to figure out how we could pass this data across all of the environments to the git-ai CLI at checkpoint time.

This is the snippet of code where VS Code stores the chat session:

```
vscode/src/vs/workbench/contrib/chat/common /chatSessionStore.ts

private async writeSession(session: ChatModel | ISerializableChatData): Promise<void> { try { const index = this.internalGetIndex(); const storageLocation = this.getStorageLocation(session.sessionId); const content = JSON.stringify(session, undefined, 2); await this.fileService.writeFile(storageLocation, VSBuffer.fromString(content)); // Write succeeded, update index index.entries[session.sessionId] = getSessionMetadata(session); } catch (e) { this.reportError('sessionWrite', 'Error writing chat session', e); } } I found the code that's responsible for storing this file.
```

PR #1189, opened 2026-04-24
fix(vscode): keep legacy Copilot hooks active in WSL/SSH/remote contexts

### LINKED ISSUE #1188: [Bug]: VS Code extension disables legacy Copilot hooks in WSL Remote, causing no AI attribution
### What broke?

When using VS Code with WSL (Remote - WSL), the git-ai VS Code extension v0.1.18+ stops detecting AI-authored changes entirely. All code produced by Copilot is attributed as 100% human-edited.

This started with the migration to native VS Code/Copilot hooks in v0.1.18 (ref #520, PR #524). The extension checks `vscode.version >= 1.109.3` and disables legacy checkpoint listeners (`onDidSaveTextDocument`, `onDidOpenTextDocument`, `onDidCloseTextDocument`, `onDidChangeTextDocument`) in favor of native `PreToolUse`/`PostToolUse` hooks. However, in a WSL Remote context, the native hooks do not appear to propagate correctly from the Windows host (where the Copilot extension runs) to the WSL Remote Extension Host (where the git-ai extension and binary run).

Downgrading to extension v0.1.17 (which uses the legacy hooks) restores correct attribution, confirming that the issue is specifically the hook migration gate, not the git-ai binary itself.

### Steps to reproduce

1. Install VS Code with the Remote - WSL extension on Windows.
2. Open a folder inside WSL via VS Code Remote.
3. Install git-ai (binary + VS Code extension v0.1.18 or later) inside WSL.
4. Use GitHub Copilot (chat or inline) to generate or modify code.
5. Commit the changes.
6. Run `git-ai blame <file>` or `git-ai status` — all lines show as human-authored.
7. Downgrade to extension v0.1.17 and repeat — attribution works correctly.

### Expected vs actual behavior

**Expected:** AI-authored lines should be correctly attributed to GitHub Copilot regardless of whether the VS Code instance is local or running via Remote - WSL.

**Actual:** On VS Code >= 1.109.3 with WSL Remote, the extension logs `[git-ai] VS Code has native Copilot hooks; skipping legacy extension checkpoint listeners` and no checkpoints are ever fired for Copilot edits, resulting in 0% AI attribution.

### Diagnostics (`git-ai debug`)

```shell
_Not available from the reporter's environment, but the core issue is observable in the extension source code._
```

### Extra context (optional)

**Root cause analysis from source code:**

1. In `src/utils/vscode-hooks.ts`, `shouldSkipLegacyCopilotHooks()` only checks the VS Code version against `MIN_VSCODE_NATIVE_HOOKS_VERSION` (1.109.3). It does **not** check whether the extension is running in a remote context (WSL, SSH, Codespaces, etc.).

2. In `src/extension.ts` (lines ~85-118), when `areLegacyCopilotHooksEnabled()` returns `false`, **none** of the legacy event listeners are registered. The extension relies entirely on native hooks that don't reach the WSL context.

3. There is no check for `vscode.env.remoteName` or `vscode.env.remoteAuthority` to detect the remote scenario.

**Suggested fix:**

```shell
// In src/utils/vscode-hooks.ts
import * as vscode from "vscode";

export function shouldSkipLegacyCopilotHooks(vscodeVersion: string): boolean {
  // In remote contexts (WSL, SSH, etc.), native hooks may not propagate
  // from the host to the remote extension host — keep legacy hooks active
  if (vscode.env.remoteName) {
    return false;
  }
  return isVersionSatisfied(vscodeVersion, MIN_VSCODE_NATIVE_HOOKS_VERSION);
}
```

Alternatively, the extension could detect at runtime that no native hooks have been received within a timeout window and automatically fall back to legacy hooks.

**Related issues:** #520, #524, #544

jorgegarciarey:
Fixes #1188

### DIFF
--- a/agent-support/vscode/src/ai-edit-manager.ts
+++ b/agent-support/vscode/src/ai-edit-manager.ts
@@ -33,9 +33,12 @@ export class AIEditManager {
   private readonly STABLE_CONTENT_DEBOUNCE_MS = 2000;
 
   constructor(context: vscode.ExtensionContext) {
-    this.legacyCopilotHooksEnabled = !shouldSkipLegacyCopilotHooks(vscode.version);
+    this.legacyCopilotHooksEnabled = !shouldSkipLegacyCopilotHooks(vscode.version, vscode.env.remoteName);
     if (!this.legacyCopilotHooksEnabled) {
       console.log(`[git-ai] AIEditManager: VS Code ${vscode.version} has native hooks; skipping legacy extension checkpoints`);
+    }
+    if (vscode.env.remoteName) {
+      console.log(`[git-ai] AIEditManager: Remote context detected (${vscode.env.remoteName}); keeping legacy hooks active`);
     }
 
     if (context.storageUri?.fsPath) {
--- a/agent-support/vscode/src/test/vscode-hooks.test.ts
+++ b/agent-support/vscode/src/test/vscode-hooks.test.ts
@@ -14,4 +14,17 @@ suite("VS Code Hook Gating", () => {
     assert.strictEqual(shouldSkipLegacyCopilotHooks("1.108.0"), false);
     assert.strictEqual(shouldSkipLegacyCopilotHooks("1.109.3-alpha"), false);
   });
+
+  test("keeps legacy hooks in remote contexts (WSL, SSH) regardless of version", () => {
+    assert.strictEqual(shouldSkipLegacyCopilotHooks("1.109.3", "wsl"), false);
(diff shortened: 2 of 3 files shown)


svarlamov:
Thank you for your first contribution @jorgegarciarey -- excited for more!!

Quick question, how has this update been working for you on remove envs/devcontainers? I need to dig deeper, but two things come to mind:

a) Shocked that the copilot hooks would run locally (🤯). Is this documented or are they looking at fixing this? Any way we could configure our hooks to run on the remote?
b) We had open issues re remote containers using the legacy approach too #237 

Ideally, we could figure out a fix that doesn't require using the legacy/extension model, but open to it if that's our only option.

PR #1197, opened 2026-04-24
Fix captured checkpoint dirty_files path normalization on macOS

svarlamov:
## Summary
- Fix macOS symlink mismatch (`/var` vs `/private/var`) in `prepare_captured_checkpoint` that caused dirty_files keys to never match repo-relative file paths, breaking copilot `create_file` attribution in daemon mode
- Rewrite `prepare_captured_checkpoint` to read files directly from disk/dirty_files instead of going through `resolve_live_checkpoint_execution`, removing dependency on git status and working log reads during capture
- Add auto-wait in test infrastructure (`git_ai_with_env`) for async daemon checkpoint completions, preventing race conditions in tests that fire captured checkpoints

## Test plan
- [x] All 5 `github_copilot_create_file` tests pass in daemon mode
- [x] All 59 `daemon_mode` tests pass
- [x] Full test suite passes (3068 tests, 0 failures)
- [x] `task lint` passes
- [x] `task fmt` passes

### DIFF
--- a/src/commands/checkpoint.rs
+++ b/src/commands/checkpoint.rs
@@ -1015,22 +1015,6 @@ pub fn prepare_captured_checkpoint(
         ));
     };
 
-    let Some(resolved) = resolve_live_checkpoint_execution(
-        repo,
-        kind,
-        agent_run_result,
-        is_pre_commit,
-        base_commit_override,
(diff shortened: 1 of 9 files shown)


svarlamov:
no longer valid

PR #1322, opened 2026-05-10
feat: add deterministic installers

heapwolf:
uses makecab.exe, msidb.exe (Windows SDK), and the built-in WindowsInstaller.Installer COM object to produce a valid MSI.

Installs git-ai.exe + git.exe shim to C:\Program Files\git-ai\bin\ and prepends to the system PATH. Supports silent enterprise deployment via msiexec /qn.


Siddhant-K-code:
Addressed Devin's release workflow findings in `aff1ada12`:

- Release artifact copy now includes both `git-ai-*` and Debian-style `git-ai_*` filenames.
- Checksum generation now uses `nullglob` and a bash array, so missing underscore/dash globs cannot force an empty `SHA256SUMS`.
- The amd64 deb artifact upload is narrowed to `release/*_amd64.deb` so it does not also upload the arm64 deb.

Siddhant-K-code:
Thanks for putting this together. This was useful prior art for the release job shape, package artifact naming, and installer smoke tests.

We’re going to close this as superseded by #1714 because the installer direction changed: new package-manager installers must install `git-ai` only and must not install `git`/`git.exe` shims or intercept Git via PATH precedence. Git AI setup should continue through trace2 via `git-ai install-hooks` / package setup.

Appreciate the groundwork here.

PR #1368, opened 2026-05-13
fix: ignore CRLF-only rewrites in attribution stats

peckfly:
## Problem

When a file switches between LF and CRLF line endings at the same time as a real edit, git can represent unchanged lines as delete/add pairs. That inflates `git-ai stats` / post-commit stats / range stats and can make unchanged existing lines look newly AI-authored.

For example, if a file has 100 existing LF lines and an AI edit adds 1 real line while the file is saved as CRLF, the old behavior could count many rewritten lines instead of the single logical addition.

## Fix

- Ignore CR-at-EOL-only delete/add pairs when building line-numbered diff hunks used by stats, post-commit stats, range stats, and diff AI accepted calculations.
- Align old attribution snapshots to matching new line-ending bytes before byte-level attribution transforms, so LF/CRLF-only rewrites do not reattribute unchanged lines.
- Preserve final-newline behavior: adding a missing final newline before appended content is still treated as a real diff, not as CRLF-only churn.

## Tests

- `cargo test --test integration eol_attribution -- --nocapture --test-threads=1`
- `cargo test --test integration stats_unit -- --nocapture`
- `cargo test --test integration range_authorship_unit -- --nocapture`
- `cargo test --test integration checkpoint_unit -- --nocapture`
- `cargo test --test integration post_commit_unit -- --nocapture`
- `cargo test attribution_tracker::tests:: -- --nocapture`
- `cargo test diff_ai_accepted -- --nocapture`
- `cargo clippy --all-targets --all-features -- -D warnings`

### DIFF
--- a/src/authorship/attribution_tracker.rs
+++ b/src/authorship/attribution_tracker.rs
@@ -220,6 +220,192 @@ fn collect_line_metadata(content: &str) -> Vec<LineMetadata> {
 }
 
 #[derive(Clone, Debug)]
+struct LineEndingSpan {
+    start: usize,
+    eol_start: usize,
+    end: usize,
+}
+
+impl LineEndingSpan {
+    fn text<'a>(&self, content: &'a str) -> &'a str {
+        &content[self.start..self.eol_start]
+    }
+
+    fn eol<'a>(&self, content: &'a str) -> &'a str {
+        &content[self.eol_start..self.end]
+    }
+
+    fn logical_key(&self, content: &str) -> LogicalLineKey {
+        LogicalLineKey {
+            text: self.text(content).to_string(),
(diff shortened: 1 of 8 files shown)


svarlamov:
Thank you for the contribution @peckfly! This is definitely a real issue, but I'm split on how to fix this.

Users expect AI changes to whitespace to be attributed to AI *and* they expect git-ai stats to line up with default git stats which treat these as substantive changes. However, I do agree that this unnecessarily inflates counts.

This PR is pretty large and touches a lot of core flows in the code. Is there any way that this could be handled via flags or something like that to avoid changing core behavior/contracts? Ideally in a way that's simpler overall as well?

PR #1378, opened 2026-05-15
Fix async post-commit stats for commit aliases

peckfly:
## Problem

In async/wrapper-daemon mode, commits made through a Git alias could successfully create the commit and authorship note, but the foreground wrapper did not recognize the invocation as a `commit`. For example, with `alias.ci=commit`, running `git ci -m "..."` skipped the inline post-commit stats path, so users would not see the same `you ... ai` stats output they get from `git commit -m "..."`.

## Root Cause

The daemon-side path already resolves Git aliases when processing trace events, but the foreground wrapper classified commands using the raw parsed command. That means `git ci -m ...` was still seen as command `ci` when deciding whether to run post-commit stats, so this check never matched `commit`.

## Fix

- Resolve Git aliases in the wrapper before classifying async invocations.
- Keep executing the original argv through Git, so Git alias behavior and argument handling remain unchanged.
- Preserve the existing built-in read-only fast path before alias resolution to avoid extra repo/config work for normal read-only commands.
- Add a regression test that configures `alias.ci=commit` and verifies `git ci -m ...` shows post-commit AI stats.

This is a general alias handling fix, not a special case for `ci`; aliases and alias chains that resolve to `commit` are classified as commit invocations.

## Tests

- `cargo fmt --check`
- `git diff --check`
- `cargo test --test async_mode async_mode_post_commit -- --nocapture`
- `GIT_AI_TEST_GIT_MODE=wrapper-daemon cargo test --test integration git_alias_resolution -- --nocapture`
- `task build`
- `task lint`

---

<a href="https://app.devin.ai/review/git-ai-project/git-ai/pull/1378" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/src/commands/git_handlers.rs
+++ b/src/commands/git_handlers.rs
@@ -61,7 +61,21 @@ pub fn handle_git(args: &[String]) {
         return;
     }
 
-    let parsed = parse_git_cli_args(args);
+    let raw_parsed = parse_git_cli_args(args);
+
+    // Preserve the built-in read-only fast path before attempting alias
+    // resolution. Aliases need a repository config lookup, but commands that are
+    // already known to be read-only should stay as cheap as they were before.
+    if is_read_only_invocation(&raw_parsed) {
+        let exit_status = proxy_to_git(args, false, None);
+        exit_with_status(exit_status);
+    }
+
+    let repository = find_repository(&raw_parsed.global_args).ok();
+    let parsed = repository
+        .as_ref()
+        .and_then(|repo| resolve_alias_invocation(&raw_parsed, repo))
+        .unwrap_or_else(|| raw_parsed.clone());
 
     // Read-only invocations don't need wrapper state (the daemon fast-paths
     // their trace events and never processes them through the normalizer).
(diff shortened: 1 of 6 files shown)


peckfly:
Added a follow-up fix commit: `10693d0ce fix: expand aliases for daemon history analysis`.

What changed:

- The original PR fixed the wrapper-side async post-commit stats path for commit aliases, for example `alias.ci = commit`.
- This follow-up fixes the daemon trace normalization path as well. The daemon already resolved alias commands to the correct primary command, but it still kept the raw alias arguments in `invoked_args`. That meant an alias like `alias.ca = commit --amend` could be normalized as `commit` without preserving `--amend`, so history analysis could emit a regular commit event instead of a commit-amend event.
- `GitBackend` now exposes an alias-expanded parsed invocation, and `TraceNormalizer` uses that expanded invocation only when it is consistent with the resolved primary command. Otherwise it falls back to the existing canonical invocation behavior.
- This is not a special case for `ca` or `--amend`; it is a general alias-expansion fix for daemon history analysis. Shell aliases and unresolved/cyclic aliases still fall back rather than being guessed.

Validation run locally:

- `cargo fmt --check`
- `git diff --check`
- `cargo test alias_commit_amend_expands_invoked_args_for_history_analysis --lib`
- `cargo test daemon_pure_trace_socket_commit_amend_alias_emits_amend_event --test daemon_mode`
- `cargo test async_mode_post_commit_shows_stats_for_aliased_commit --test async_mode`
- `cargo test daemon_pure_trace_socket_write_mode_applies_amend_rewrite --test daemon_mode`
- `cargo test git_backend --lib`
- `cargo test trace_normalizer --lib`
- `cargo test --lib -- --test-threads=1` passed: 1579 passed, 0 failed

Note: `cargo test --test daemon_mode` still has a few unrelated failures in this branch/environment. I verified the same failures reproduce on the pre-follow-up PR commit `c19b68d50`, so they do not appear to be introduced by this alias-expansion change. The new alias-amend daemon test itself passes.

svarlamov:
Closing as Git AI no longer installs a git wrapper

PR #1383, opened 2026-05-16
fix: support Cursor path field in tool input

peckfly:
## Summary
- read Cursor file edit paths from `tool_input.path` and `tool_input.filePath` in addition to `tool_input.file_path`
- keep `file_path` precedence when multiple fields are present
- add regression coverage for `path` fallback and precedence

## Testing
- `cargo fmt -- --check`
- `cargo test commands::checkpoint_agent::presets::cursor::tests --lib -- --test-threads 12`
- `cargo test cursor_file_edit -- --test-threads 12`

---

<a href="https://app.devin.ai/review/git-ai-project/git-ai/pull/1383" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/src/commands/checkpoint_agent/presets/cursor.rs
+++ b/src/commands/checkpoint_agent/presets/cursor.rs
@@ -77,13 +77,8 @@ impl AgentPreset for CursorPreset {
             )));
         }
 
-        // Extract file_path from tool_input (file-edit tools only).
-        let file_path = data
-            .get("tool_input")
-            .and_then(|ti| ti.get("file_path"))
-            .and_then(|v| v.as_str())
-            .map(normalize_cursor_path)
-            .unwrap_or_default();
+        // Extract the edited path from Cursor file-edit tool input.
+        let file_path = cursor_file_path_from_tool_input(data.get("tool_input"));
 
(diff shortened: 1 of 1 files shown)


svarlamov:
Is this Cursor format something new that you've seen recently?

peckfly:
> Is this Cursor format something new that you've seen recently?

I observed this recently in local Cursor agent transcripts, but I cannot say for sure whether it is a brand new Cursor format or an existing variant that git-ai did not handle before.

In the failing case I debugged, Cursor emitted a file edit tool call like `StrReplace` with the edited file under `tool_input.path` instead of `tool_input.file_path`. Since the Cursor preset only looked at `file_path`, git-ai did not resolve the edited file path and the AI file-edit checkpoint was missed. A later editor/known-human checkpoint then captured the saved file contents, so the commit attribution ended up classified as human instead of AI.

The directly observed field was `path`. I kept `file_path` as the first choice to preserve existing behavior, and added `path` as a fallback. I also added `filePath` as a small compatibility fallback for the same concept, but the concrete format I saw in the transcript was `tool_input.path`.

Issue #1377, opened 2026-05-14
Command to await commit stats


svarlamov:
Create a new git-ai subcommand that will up to X ms for a commit to get an authorship note and then print out stats. This could be used in an optional post-commit hook to print % AI stats without using wrapper

Ameb8:
I gave this a shot, but any feedback would be greatly appreciated

svarlamov:
hey @Ameb8 thank you for the contribution! took a quick glance and the implementation looks pretty good to me! do you know if this would work in a git post-commit hook as it's currently implemented?

my bad for not including that in the issue, but that's the primary use case for this feature (at least on my end)

Ameb8:
Thanks for the response! I went ahead and updated my PR. I implemented await-stats invocation through the `src/commands/git_handlers.rs` file rather than a literal post-commit hook, as from my understanding, that seems to be the direction this repo is heading. However, if that was incorrect, I believe the await-stats command would work as-is if called from a literal hook.

Happy to hear any feedback or make any updates/fixes.

svarlamov:
We've built a new `git ai await` command that will wait for all pending operations 😄

PR #1401, opened 2026-05-20
feat: implement await-stats subcommand

Ameb8:
# Implements #1377: add `git ai await-stats` command

## Summary

This PR adds a new read-only `git ai await-stats` subcommand for the post-commit gap between when `git commit` returns and when the git-ai daemon finishes writing the commit's authorship note. The command resolves a target commit (defaulting to `HEAD`), polls for the note through the existing notes API, validates that the note is parseable authorship data, and then prints the same commit stats users already get from `git ai stats`. It supports the expected CLI surface from the spec and design docs: `--timeout`, `--interval`, `--commit`, `--json`, `--quiet`, and command-specific help, while keeping successful output aligned with the existing terminal renderer and JSON schema instead of introducing a second stats format.

The implementation also makes the failure modes explicit so callers can distinguish "still waiting" from "bad input" and "bad note data". Timeouts return exit code `1` and can be silenced with `--quiet`, unresolved commit revs return exit code `2`, and corrupt note data returns exit code `3`. Alongside the new command module, this PR wires `await-stats` into `git ai` dispatch and top-level help output, and updates the integration test harness so commands in this path daemon-sync consistently with other note-reading subcommands. This gives us a small, reusable CLI primitive now, and it also establishes the user-facing contract needed for future plain Git `post-commit` hook usage without baking hook-specific behavior into the command itself.

## Test Plan

- Added command-module parser/unit coverage for default options, full option parsing, missing values, zero interval rejection, and unknown flag rejection.
- Added integration coverage for successful `await-stats --json` output when a note exists.
- Added integration coverage for default `HEAD` behavior and human-readable stats output.
- Added integration coverage for `--commit <rev>` against an earlier commit.
- Added integration coverage for timeout behavior when the authorship note is absent.
- Added integration coverage for `--quiet` suppressing timeout output.
- Added integration coverage for unresolved commit revs returning exit code `2`.
- Added integration coverage for corrupt authorship notes returning exit code `3`.
- Updated the integration harness so `await-stats` performs the same daemon sync used by other note-reading git-ai commands.

---

<a href="https://app.devin.ai/review/git-ai-project/git-ai/pull/1401" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/src/commands/await_stats.rs
+++ b/src/commands/await_stats.rs
@@ -1,0 +1,369 @@
+use crate::authorship::authorship_log_serialization::AuthorshipLog;
+use crate::authorship::ignore::effective_ignore_patterns;
+use crate::authorship::stats::{stats_for_commit_stats, write_stats_to_terminal};
+use crate::error::GitAiError;
+use crate::git::find_repository;
+use crate::git::notes_api::read_note;
+use crate::git::repository::Repository;
+use std::thread;
+use std::time::{Duration, Instant};
+
+const DEFAULT_TIMEOUT_MS: u64 = 5_000;
+const DEFAULT_INTERVAL_MS: u64 = 100;
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+struct AwaitStatsOptions {
+    // Polling options
+    timeout_ms: u64,
+    interval_ms: u64,
+
+    // Output format options
+    json: bool,
+
+    // Commit selection
+    commit: String,
+
+    // Suppress expected timeout noise for callers that poll externally
+    quiet: bool,
+}
+
+impl Default for AwaitStatsOptions {
+    // Build the default await-stats configuration used when no flags are provided.
+    fn default() -> Self {
+        Self {
+            timeout_ms: DEFAULT_TIMEOUT_MS,
+            interval_ms: DEFAULT_INTERVAL_MS,
+            json: false,
+            commit: "HEAD".to_string(),
+            quiet: false,
+        }
+    }
+}
+
+#[derive(Debug)]
+pub enum AwaitStatsError {
+    // CLI argument errors
+    BadArgs(String),
+    // Revision did not resolve to a commit
+    BadCommit(String),
+    // Authorship note was not created before the timeout elapsed
+    Timeout { commit_sha: String, timeout_ms: u64 },
+    // Note exists, but cannot be parsed or converted to stats
+    CorruptNote(String),
+    // Lower-level git-ai error
+    GitAi(GitAiError),
+}
+
+impl AwaitStatsError {
+    // Map await-stats failures to the command's public process exit codes.
+    fn exit_code(&self) -> i32 {
+        match self {
+            Self::Timeout { .. } => 1,
+            Self::BadArgs(_) | Self::BadCommit(_) => 2,
+            Self::CorruptNote(_) => 3,
+            Self::GitAi(_) => 1,
+        }
+    }
+}
+
+impl From<GitAiError> for AwaitStatsError {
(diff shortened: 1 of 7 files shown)


Ameb8:
## PR Updated 

I added the automatic `await-stats` behavior to the managed git wrapper's successful `git commit` path. After the wrapper proxies a commit to real Git and the commit succeeds, it now waits briefly for the async authorship note, validates that note, and renders the same stats output through shared `await-stats` code instead of keeping separate polling/rendering logic in the wrapper.

The main refactor is in `src/commands/await_stats.rs`: the CLI entrypoint still behaves the same, but the note resolution, note polling, note validation, and stats rendering pieces are now reusable helpers that accept an existing `Repository`. `src/commands/git_handlers.rs` then calls those helpers from `maybe_show_async_post_commit_stats(...)` while preserving the wrapper-specific automatic-output rules: no output for dry-run, quiet, porcelain, no-status, config-quiet, or non-interactive commits, and no expensive stats computation for merge or large commits.

I wired this through the wrapper path rather than a literal `.git/hooks/post-commit` hook because the managed wrapper is the current post-commit-equivalent integration point in this codebase, and the old core hook symlink path appears to be sunset. The wrapper also already has the parsed git args, terminal context, and repo handle needed to make the automatic behavior conservative. A real post-commit hook should still be able to invoke `git ai await-stats` manually, but this change does not install or modify hook files.

src/commands/await_stats.rs
```diff
@@ -0,0 +1,409 @@
...
+    let raw_note = wait_for_authorship_note(
+        repo,
+        &commit_sha,
+        Duration::from_millis(options.timeout_ms),
+        Duration::from_millis(options.interval_ms),
+    )?
+    .ok_or_else(|| AwaitStatsError::Timeout {
+        commit_sha: commit_sha.clone(),
+        timeout_ms: options.timeout_ms,
+    })?;
```
Siddhant-K-code, line 138: This polling loop does not work for the primary literal `post-commit` hook use case from #1377. In daemon mode the authorship note is written only after the parent `git commit` root trace reaches `exit`/`atexit`, gets normalized into a `CommitCreated` event, and the daemon side-effect calls `post_commit_with_final_state`. 

A `.git/hooks/post-commit` hook runs before that parent `git commit` exits, so invoking `git ai await-stats` here blocks the commit process while waiting for a note that the daemon cannot produce until this process returns. This should either trigger/sync the daemon work explicitly from the hook context, or the supported hook design needs to avoid waiting from inside the root commit process.

tests/integration/await_stats.rs
```diff
@@ -0,0 +1,234 @@
...
+            "10",
+            "--json",
+        ])
+        .expect("await-stats should succeed");
+    let stats: CommitStats = serde_json::from_str(&extract_json_object(&raw)).unwrap();
+
+    assert_eq!(stats.ai_additions, 2);
+    assert_eq!(stats.ai_accepted, 2);
+    assert_eq!(stats.git_diff_added_lines, 2);
+}
```
Siddhant-K-code, line 72: These success tests run `await-stats` after `stage_all_and_commit` has returned, which exercises the easy path where the daemon has already had a chance to observe the commit completion and write the note. Please add coverage for the intended hook path: install a real `post-commit` hook that invokes `git ai await-stats --timeout ...` during `git commit`, then assert it prints stats instead of timing out. 

That test should fail with the current implementation because the hook waits before the root commit trace is finalized.

Siddhant-K-code (commented):
@Ameb8, would love to know your thougths on these comments ^ - can you address it?

PR #1474, opened 2026-05-31
fix(cursor): handle large file payloads with unescaped quotes

Forwardxiang:
When Cursor sends large payloads containing file content with unescaped quotes, standard JSON parsing fails. The metadata fields we need are properly escaped, so we can extract them using regex instead of relying on full JSON parsing.

---

<a href="https://app.devin.ai/review/git-ai-project/git-ai/pull/1474" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


svarlamov:
Need a Cursor upstream fix for this

Issue #1079, opened 2026-04-14
[Bug]: authorship is sometimes lost through rebases


JeanFred:
### What broke?

Since recently, authorship is sometimes lost through rebases (although I did confirm in the past it worked).

I’m on a branch with commits with git-ai information, as evidenced by git-ai log or git-ai show xyz[1]. 

After rebasing on top of main branch, all authorship information is lost (as evidenced by git-ai log).[2]

I first noticed it with interactive rebases using `[git-interactive-rebase-tool](https://github.com/MitMaro/git-interactive-rebase-tool)`, but I also tried with a more vanilla sequence editor, and also tried non-interactive rebase (ie `git rebase origin/master`) − authorship is lost.

This does not always happen though, I don’t understand why. But I can definitely reproduce on the repo I’m working on at the moment.

[1] For example
```
Notes (ai):
    ---
    {
      "schema_version": "authorship/3.0.0",
      "git_ai_version": "1.2.8",
```
[2] 
```
No authorship data found for this revision
```

### Steps to reproduce

See above

### Expected vs actual behavior

Expected: git-ai information is kept post-rebases
Actual: git-ai information is lost post-rebases

### Diagnostics (`git-ai debug`)

```shell
git-ai debug report
Generated (UTC): 2026-04-14T10:50:35.280446608+00:00

== Versions ==
Git AI version: 1.2.8
Git AI binary: /home/jeanfred/.git-ai/bin/git-ai
Git binary path: /usr/bin/git
Git version: git version 2.53.0

== Platform ==
OS family: unix
OS: linux
Arch: x86_64
Kernel: Linux 6.8.0-106-generic x86_64
Hostname: redacted
Shell: /bin/bash
Current dir: /home/jeanfred/redacted

== Hardware ==
CPU: 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz
Physical cores: <unavailable>
Logical cores: 8
Memory: 15.34 GB (16476557312 bytes)

== Repository ==
In repository: true
Workdir: /home/jeanfred/redacted
Git dir: /home/jeanfred/redacted.git
Git common dir: /home/jeanfred/redacted/.git
Branch: feature/redacted
HEAD: redacted
Remotes:
  origin = git@github.com:redacted.git

== Git Config ==
Command: /usr/bin/git config --list --show-origin --show-scope
<snip>

== Git AI Config ==
  config_file_path: /home/jeanfred/.git-ai/config.json
[...]

Issue #11, opened 2025-07-25
Can all functions be encapsulated into the extensions plug-in without installing git-ai.exe separately?


helong2048:
Can all functions be encapsulated into the extensions plug-in without installing git-ai.exe separately?

acunniffe:
Possibly - I have not spent a lot of time with git extensions. I know they exist / work re: `git lfs` but can't find good docs for how to build them. Have any resources you recommend?

svarlamov:
Added documentation on [wrapper](https://usegitai.com/docs/enterprise-deployment-wrapper) vs [hooks](https://usegitai.com/docs/enterprise-deployment-hooks) deployment options for `git-ai`

Issue #10, opened 2025-07-20
windows env , all files in the .git\refs\ai directory cannot be successfully created.


helong2048:
windows env , all files in the .git\refs\ai directory cannot be successfully created.

PR #1542, opened 2026-06-11
Rewrite rewrite-ops attribution: object-based core, trace2 ingestion, fuzzer + fix gt-create flake

svarlamov:
## Summary

Ground-up rewrite of how git-ai tracks and migrates authorship attribution across history-rewriting git operations (rebase, squash, amend, cherry-pick, `update-ref`/`commit-tree` restacks, reset, stash, revert), plus the daemon trace2 ingestion that feeds it and an attribution fuzzer that proves it. The legacy machinery is **deleted, not deprecated**.

This adopts the object-based core from the earlier `feat/attr-fuzzer-v2` effort and **fixes the load-dependent attribution-loss flake that branch never beat** (its nine "trigger ubuntu reliability run" commits are the fossil record of fighting it without a root-cause fix).

## What changed

**Old way removed** (`src/`): `rebase_authorship.rs`, `rewrite_log.rs` (+`.git/ai/rewrite_log`), `squash_authorship.rs`, `commands/hooks/{rebase,stash,push}_hooks.rs`, `diff_tree_to_tree.rs`. The git proxy is now a thin passthrough; all attribution side effects run in the shared daemon, driven by trace2.

**New core**: `src/authorship/rewrite.rs` (`RewriteEvent` + `handle_rewrite_event`), per-op modules `rewrite_{reset,stash,revert,cherry_pick}.rs`, `hunk_shift.rs`, `conflict_resolution.rs`. Note migration is batched (one `diff-tree --stdin`, one `range-diff`). Lines outside diff hunks keep attribution and shift; lines inside changed hunks are dropped (new content needs checkpoint evidence).

**Daemon ingestion**: `src/daemon/ref_cursor.rs` establishes exact command ownership via a per-family reflog cursor; commands without a cursor or immutable argv OIDs fail closed rather than guess.

**Specs** (`docs/`): rewrite-ops, daemon-trace2-ingestion, attribution-fuzzer — rewritten from first principles. `AGENTS.md` updated.

## The fix that made it stable

The family actor keeps one `RefCursor` per repo family. The authoritative *in-order* cursor marks where the next unconsumed reflog entry begins. The daemon-ingress reflog offset is captured **asynchronously** and can race ahead of (or behind) git's own reflog append.

Previously the ingress offset was allowed to **advance the cursor**. Under load, `gt create` (`…switch → commit → switch-back`) would capture the offset *past the commit's own reflog entry*, so enrichment found no HEAD transition → `OpaqueCommand` → the commit silently lost its AI attribution (status still "ok"). This was the ~30%-under-load graphite/`rebase_patch_stack` flake.

Fix: the ingress offset is now a **soft selection hint, never a cursor advance**. Entry selection prefers the first match at/after the hint (so an untraced duplicate-message commit is still skipped — the cold-start case) and falls back to the cursor's earliest match when the hint raced past the command's own entry (the graphite case). `CursorEntry`/`ReflogRecord` gain a `start_offset` so selection can compare against the hint. Two scenarios that are byte-identical but need opposite outcomes are now both correct, with dedicated `ref_cursor` unit tests.

This is a pure correctness fix — **no daemon syncs were added**. Production read commands (`blame`/`show`/`status`) never sync; the only `SyncFamily` callers are the test harness, immediately before assertions.

## Fuzzer

`tests/integration/fuzzer/` now asserts **all three attribution classes** (`Ai`/`KnownHuman`/`Untracked`) via `blame --show-prompt` (previously only AI vs non-AI, which couldn't catch known-human↔untracked degradation). Added soft-reset-recommit and stash-roundtrip operations; `fuzz_random` prints its seed for reproducibility.

## Verification

- `task test`: **all suites green** (3066 integration tests, 0 failed); `task lint` and `cargo fmt --check` clean.
- Graphite suite: **30/30** stress iterations green (was ~30% flaky).
- Marathon fuzzer: **0/3** fail (450+ pathological ops under the 3-class assertion).
- **10× parallel `task test` overload** at 4–5× CPU saturation (load avg ~140–165): **zero attribution flakes**. The only failures were acceptable resource-starvation timeouts (daemon-sync windows, `git worktree add` materialization) — none computed a wrong author.

## CI

No retry harness (`scripts/ci-test-with-retry.{sh,ps1}` were already removed); CI runs `task test` once, fail-fast, so any remaining flake surfaces immediately. (The one `nick-fields/retry` left wraps live-LLM-API calls in the nightly agent workflow — external-network retry, unrelated to the test suite.)

---

<a href="https://app.devin.ai/review/git-ai-project/git-ai/pull/1542" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


svarlamov:
@devin-ai-integration please review this PR.

(It was briefly opened as a draft before being marked ready, so the automatic review may have been skipped.)

src/authorship/rewrite.rs
```diff
@@ -0,0 +1,1345 @@
...
+    }
+
+    mappings
+}
+
+fn find_next_sha(s: &str) -> Option<(String, &str)> {
+    let bytes = s.as_bytes();
+    let mut i = 0;
+    while i + 40 <= bytes.len() {
+        let candidate = &s[i..i + 40];
```
svarlamov: **Bug (code-review):** UTF-8 char-boundary panic: find_next_sha slices `&s[i..i+40]` byte-by-byte over `git range-diff -s` output, which includes commit subjects; a multibyte char panics the rewrite, and the foreground `git-ai ci` path is not catch_unwind-guarded.

**Trigger:** Any rebase/squash/cherry-pick whose range-diff output line carries a commit subject with a non-ASCII char (accent, CJK, emoji): once `i` walks into the subject, `&s[i..i+40]` panics 'byte index N is not a char boundary'. In `git-ai ci` (ci_context.rs:run, not catch_unwind-wrapped) this crashes the CI run and produces no authorship rewrite.

src/authorship/rewrite_reset.rs
```diff
@@ -0,0 +1,233 @@
...
+        // for overlapping line ranges.
+        let existing = file_attributions.entry(fa.file_path.clone()).or_default();
+        for new_attr in shifted {
+            // Remove any existing attributions that are fully covered by this new one
+            existing.retain(|old| {
+                !(old.start_line >= new_attr.start_line && old.end_line <= new_attr.end_line)
+            });
+            // For partial overlaps, trim existing attributions
+            let mut trimmed: Vec<LineAttribution> = Vec::new();
+            existing.retain(|old| {
```
svarlamov, line 194: **Bug (code-review):** extract_attributions_from_log_shifted drops the TAIL of an existing attribution that fully encloses a new one: the head-overlap branch returns false before the tail-overlap branch can preserve [new.end+1, old.end].

**Trigger:** Backward soft/mixed reset undoing commits where a later commit's range (e.g. lines 4-6, author B) lands strictly inside an earlier commit's range (lines 1-10, author A) for the same file: the merge keeps [1-3]A and [4-6]B but silently drops [7-10]A. Reconstructed INITIAL loses attribution for lines 7-10; the next commit mis-attributes them.

src/git/refs.rs
```diff
@@ -136,7 +136,16 @@ fn batch_read_blob_contents(
 
     let stdin_data = blob_oids.join("\n") + "\n";
     let output = exec_git_stdin(&args, stdin_data.as_bytes())?;
-    parse_cat_file_batch_output_with_oids(&output.stdout)
+    let results = parse_cat_file_batch_output_with_oids(&output.stdout)?;
+    for oid in blob_oids {
+        if !results.contains_key(oid) {
+            return Err(GitAiError::Generic(format!(
+                "missing git blob object referenced by authorship note: {}",
```
svarlamov, line 143: **Bug (code-review):** Batched note-blob read hard-errors on a single missing blob, where origin/main degraded that one commit to NoLog; one dangling note-blob OID now fails an entire range/batch.

**Trigger:** A commit's authorship note references a blob absent from the object DB (partial/blobless/promisor clone, interrupted GC, corruption). `git-ai show`/log/blame over a range (get_commits_with_notes_from_list) and cherry-pick/rebase/squash/revert note-shifts (read_notes_batch callers) now return Err for the WHOLE batch instead of skipping the one bad commit, losing all attribution in that operation.

src/daemon.rs
```diff
@@ -7247,78 +4375,283 @@ impl ActorDaemonCoordinator {
...
+                            }
+                            _ => {}
+                        }
+                    }
+                    crate::daemon::domain::SemanticEvent::CommitCreated { base, new_head } => {
+                        let mut handled_as_squash_merge = false;
+                        if !new_head.is_empty()
+                            && cmd.primary_command.as_deref() == Some("commit")
+                            && let Some(pending) =
+                                self.take_pending_squash_merge_for_worktree(worktree.as_ref())?
```
svarlamov, line 4565: **Bug (code-review):** Pending `merge --squash` state is never cleared on abort/reset; a later ordinary commit whose base equals the stored `onto` is misinterpreted as the squash commit.

**Trigger:** `git merge --squash feature` (sets pending {source_head, onto}) → `git merge --abort` / `git reset --hard` while staying on onto without committing → pending survives (no clear fn; gc_stale keeps non-empty entries forever) → a later unrelated `git commit` with base==onto fires handle_rewrite_event(SquashMerge), reconstructing feature's authorship onto a commit that never included feature's changes.

src/authorship/rewrite_stash.rs
```diff
@@ -0,0 +1,595 @@
...
+        let _ = fs::remove_dir_all(&worklog_dir);
+    }
+    Ok(())
+}
+
+fn save_stash_attributions(
+    repo: &Repository,
+    stash_sha: &str,
+    head_sha: &str,
+    _pathspecs: &[String],
```
svarlamov: **Bug (code-review):** save_stash_attributions ignores its `_pathspecs` arg and copies the entire working-log dir, while clean only removes matched files; and path_matches_any dropped the old trailing-`*` glob branch.

**Trigger:** `git stash push -- src/a.txt` while AI edits exist in both a.txt and b.txt: the saved stash worklog includes b.txt; on pop, b.txt's attribution is restored and re-applied even though b.txt was never stashed (double-attribution). Separately, `git stash push -- 'src/foo*'` no longer matches src/foobar.txt in clean_working_log_for_stash, leaving stale attribution in the live working log.

src/authorship/rewrite.rs
```diff
@@ -0,0 +1,1345 @@
...
+        }
+    }
+
+    DiffTreeResult {
+        hunks_by_file,
+        renames,
+    }
+}
+
+fn extract_b_path(diff_header: &str) -> Option<String> {
```
svarlamov, line 966: **Bug (code-review):** extract_b_path uses rfind(" b/") on diff-tree -p headers and the diff-tree invocation does not force a/ b/ prefixes (uses General profile, not PatchParse), so user git config diff.noprefix/mnemonicPrefix leaks in and breaks path extraction.

**Trigger:** User has diff.noprefix=true or diff.mnemonicPrefix=true: diff-tree emits headers like `diff --git file file` (or i/ w/ prefixes), rfind(" b/") returns None, current_file stays None, no hunks recorded → attestation line numbers left unshifted → wrong attribution after squash/cherry-pick/reset/revert (the rewrite/shift path, unlike the normal-commit diff which forces a/ b/ via PatchParse).

src/authorship/conflict_resolution.rs
```diff
@@ -0,0 +1,269 @@
...
+
+    recovered_log
+        .attestations
+        .retain(|attestation| !attestation.entries.is_empty());
+    retain_referenced_metadata(&mut recovered_log);
+    merge_file_attestations(target, &recovered_log);
+    merge_authorship_metadata(target, &recovered_log);
+}
+
+pub fn merge_conflict_resolution_authorship(
```
svarlamov, line 146: **Bug (code-review):** merge_conflict_resolution_authorship dropped the old #1079 original-note remap fallback: AI attestation for conflict-resolved lines whose content CHANGED (and weren't re-checkpointed) is silently lost; only Equal-line provenance survives.

**Trigger:** rebase/cherry-pick --continue resolving a conflict in an AI-tracked file to content differing from the original commit, with no AI checkpoint during resolution: recover_exact_source_lines_from_mapping finds no Equal-line mapping for the changed region and the parent working log has no AI entry, so the rebased commit loses the original AI attestation entirely (origin/main remapped the original note as a best-effort provenance fallback).

src/git/repo_storage.rs
```diff
@@ -186,39 +178,85 @@ impl RepoStorage {
...
+            new_dir.to_path_buf(),
+            new_sha,
+            self.repo_workdir.clone(),
+            canonical,
+            None,
+        );
+
+        let mut merged_initial = old_log.read_initial_attributions();
+        let new_initial = new_log.read_initial_attributions();
+        merged_initial.files.extend(new_initial.files);
```
svarlamov: **Bug (code-review):** merge_working_log_dirs merges INITIAL maps via HashMap::extend, which overwrites old-base entries with new-base entries for shared paths — contradicting the doc's 'preserve old-base entries first' precedence.

**Trigger:** rename_working_log(old,new) when the destination already has a working log and both bases touched the same file: `merged_initial.files.extend(new_initial.files)` (and prompts/file_blobs/humans/sessions) replaces the old base's carried-over uncommitted AI attribution for that path with the destination's, losing it. Only the checkpoints Vec preserves old-first order.

src/authorship/rewrite.rs
```diff
@@ -0,0 +1,1345 @@
...
+        let candidate = &s[i..i + 40];
+        if is_hex_sha(candidate) {
+            return Some((candidate.to_string(), &s[i + 40..]));
+        }
+        i += 1;
+    }
+    None
+}
+
+fn is_hex_sha(s: &str) -> bool {
```
svarlamov: **Bug (code-review):** is_hex_sha/find_next_sha (40 chars) and is_tree_pair_separator (81 bytes) hard-code SHA-1 lengths, but is_valid_git_oid accepts 64-char SHA-256; range-diff/diff-tree parsing silently produces zero mappings/separators in SHA-256 repos.

**Trigger:** In a `--object-format=sha256` repo, range-diff emits 64-char OIDs so parse_range_diff_output finds none → rebase/squash note-shift is a silent no-op (attribution not carried across rewrites); parse_batched_diff_tree_output recognizes no 81-byte separators → all pairs get empty hunks → unshifted attestations. No error raised.

src/daemon/reducer.rs
```diff
@@ -58,100 +91,149 @@ fn apply_ref_changes(state: &mut FamilyState, cmd: &NormalizedCommand) {
...
-    else {
-        return;
+    let key = canonicalize_path(worktree);
+    let previous = state.worktrees.get(&key);
+    let head_change = cmd
+        .ref_changes
+        .iter()
+        .rfind(|change| change.reference == "HEAD");
+
+    let (head, branch, detached) = if let Some(head_change) = head_change {
```
svarlamov, line 109: **Bug (code-review):** apply_worktree_state misclassifies `git checkout/switch <existing-branch>` as detached (branch=None): a plain branch checkout moves HEAD but updates no refs/heads ref, so unique_branch_for_head_change finds no match and the argv-branch fallback arm is unreachable.

**Trigger:** `git checkout main` (existing branch at a different commit) records WorktreeState{branch:None, detached:true}. A later command needing branch-anchored ref disambiguation calls current_worktree_branch_ref → None, bypassing precise branch matching for permissive fallbacks that can mis-select or fail to pair the right branch ref when several branches share a commit/timestamp.

src/daemon/ref_cursor.rs
```diff
@@ -0,0 +1,4675 @@
...
+            || !cmd.cherry_pick_source_oids.is_empty()
+            || applied_count > 0
+        {
+            self.pending_cherry_pick_source_oids.clear();
+        }
+
+        Ok(())
+    }
+
+    fn enrich_revert(
```
svarlamov, line 412: **Bug (code-review):** enrich_revert lacks the pending-source-OID carryover that enrich_cherry_pick maintains, so on `git revert --continue` the originally-reverted commit OID is lost and attribution falls back to first-parent (only correct for `revert HEAD`).

**Trigger:** `git revert <older-non-HEAD-commit>` → conflict → resolve → `git revert --continue`: revert_source_oids is forced empty (no pending_revert_source_oids state), so apply_revert_complete_rewrite passes reverted_commit=None and handle_revert_commit falls back to first_parent_sha(parent) — wrong source base for a non-HEAD revert → mis-attributes the re-introduced lines.

src/daemon/ref_cursor.rs
```diff
@@ -0,0 +1,4675 @@
...
+        let entries = read_reflog_entries(key, &path, "HEAD", start)?;
+
+        Ok(entries.into_iter().find(|entry| {
+            !self.entry_consumed(entry)
+                && pull_reflog_action_is(&entry.message, action, "start")
+                && expected.matches_span_boundary(entry)
+        }))
+    }
+
+    fn find_head_span_start_entry(
```
svarlamov, line 1139: **Bug (code-review):** The contiguous-span finders (find_head_span_start_entry / find_rebase_start_entry / find_pull_start_entry) use plain `.find()` and ignore the ingress-offset hint that find_entry_in_log applies, so a duplicate-message span-start preceding the command's own entry can be selected.

**Trigger:** A rebase/pull/cherry-pick where an untraced HEAD entry shares the span-start message and still matches a current-state OID, sitting unconsumed after the cursor floor: the span finder returns it instead of the command's real start, shifting the wrong source commit's attribution. Heavily constrained by old_oid/message checks, so conditional.

src/authorship/rewrite.rs
```diff
@@ -0,0 +1,1345 @@
...
+    s.len() == 40 && s.bytes().all(|b| b.is_ascii_hexdigit())
+}
+
+fn derive_merge_commit_mappings(
```
svarlamov, line 675: **Bug (code-review):** derive_merge_commit_mappings greedily binds each old merge to the first new merge with corresponding parents and breaks, with no global uniqueness/scoring, so sibling merges with identical parent structure can be mispaired.

**Trigger:** Two old merges with identical parent-correspondence (e.g. both merge the same base/branches) and two matching new merges: the earlier old merge claims the first new merge, forcing the other into the leftover — swapping their attribution when the globally-correct assignment was the reverse. Requires duplicate parent structure; uncommon.

src/git/repo_storage.rs
```diff
@@ -558,6 +596,32 @@ impl PersistedWorkingLog {
...
+    pub fn observed_file_snapshot(&self) -> Result<HashMap<String, String>, GitAiError> {
+        let initial = self.read_initial_attributions();
+        let mut snapshot = HashMap::new();
+
+        for file_path in initial.files.keys() {
+            let content = self
+                .stored_initial_file_content_from(&initial, file_path)
+                .ok_or_else(|| {
+                    GitAiError::Generic(format!(
+                        "INITIAL missing persisted file snapshot for {}",
```
svarlamov, line 623: **Bug (code-review):** stored_initial_file_content_from now hard-errors on legacy blob-less INITIAL data; observed_file_snapshot propagates it on the main commit path, so a stale pre-blob INITIAL surviving an upgrade fails the next commit.

**Trigger:** User upgrades git-ai while a stale INITIAL working log (pre-2026-03-20, no file_blobs) exists for the current base commit. Next `git commit` -> post_commit_from_working_log -> observed_file_snapshot() -> Err here, failing note generation. origin/main had a graceful read_current_file_content fallback that was removed. (Also see stored_initial_file_content_from ~L730.)

src/authorship/virtual_attribution.rs
```diff
@@ -1734,12 +2137,20 @@ impl VirtualAttributions {
                         // Find nearest attributed neighbor after this line
                         let next = line_to_author.iter().find(|(l, _)| *l > line);
 
-                        // Fill only if both neighbors exist and are the same AI author
+                        // Fill if both neighbors exist and are the same AI author
```
svarlamov, line 2421: **Bug (code-review):** Attestation entries are emitted by iterating a std HashMap (committed_lines_map, just below at ~L2166 and ~L2503) whose per-process-randomized order is never sorted across author_ids, so byte-identical commits produce non-deterministic note entry ordering.

**Trigger:** A file attributed to >=2 author_ids: FileAttestation entries are written in HashMap iteration order (only ranges within an entry are sorted), so the same commit yields different note bytes run-to-run -> breaks note-content determinism, byte-stable comparisons, and insta snapshots. Collect+sort author_ids before emitting.

svarlamov:
@devin-ai-integration please review this PR. It now includes 8 follow-up commits addressing the code-review findings (#1+#10 byte-safe SHA scan + SHA-256, #2 reset enclosing-range tail, #5 stash pathspec scoping + glob, #9 working-log merge precedence, #11 deterministic attestation order), plus docs for the intentional/false-positive items (#6, #7) and deferred-limitation comments (#4, #12-#15).

PR #1588, opened 2026-06-20
Add async rewrite commit metrics

svarlamov:
## Summary
- add a new rewrite_committed metrics event with operation kind and original commit SHAs
- collect post-rewrite commit metric targets for rebase, squash, cherry-pick, amend, revert, reset/update-ref rewrite paths
- emit rewrite commit metrics asynchronously behind rewrite_metrics_events (debug true, release false)

## Validation
- task fmt
- task lint
- task build
- task test

---

<a href="https://app.devin.ai/review/git-ai-project/git-ai/pull/1588" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/src/authorship/rewrite_revert.rs
+++ b/src/authorship/rewrite_revert.rs
@@ -35,6 +35,7 @@ pub(crate) fn handle_revert_commits_with_metrics(
     if specs.is_empty() {
         return Ok(Vec::new());
     }
+    let collect_metrics = crate::authorship::rewrite::rewrite_metrics_enabled();
 
     // Resolve every spec's parent_sha (only those missing need a lookup) and the
     // source_base_sha (= first parent of the reverted commit, or of the parent
@@ -215,12 +216,19 @@ pub(crate) fn handle_revert_commits_with_metrics(
         let Ok(note_str) = log.serialize_to_string() else {
             continue;
         };
-        writes.push((r.revert_commit.clone(), note_str));
-        metric_commits.push(RewriteMetricCommit::new(
-            r.revert_commit.clone(),
-            r.original_shas.clone(),
-            RewriteMetricOperation::Revert,
-        ));
+        writes.push((r.revert_commit.clone(), note_str.clone()));
+        if collect_metrics {
+            metric_commits.push(
+                RewriteMetricCommit::new(
+                    r.revert_commit.clone(),
+                    r.original_shas.clone(),
+                    RewriteMetricOperation::Revert,
+                )
+                .with_parent_sha(r.parent_sha.clone())
+                .with_authorship_note(note_str)
+                .with_parent_diff(diff_results[added_idx[i]].clone()),
+            );
+        }
     }
 
     if !writes.is_empty() {
--- a/src/authorship/rewrite_revert.rs
+++ b/src/authorship/rewrite_revert.rs
@@ -6,6 +6,7 @@ use crate::authorship::authorship_log_serialization::{
 };
 use crate::authorship::hunk_shift::apply_hunk_shifts_to_file_attestation;
 use crate::authorship::rewrite::compute_diff_trees_batch;
+use crate::authorship::rewrite::{RewriteMetricCommit, RewriteMetricOperation};
 use crate::error::GitAiError;
 use crate::git::notes_api;
 use crate::git::repository::{Repository, exec_git, exec_git_stdin};
@@ -23,9 +24,12 @@ pub struct RevertSpec {
 /// commits were reverted: one batched first-parent rev-parse, one batched note
 /// read, one batched diff-tree (covering every commit's two diff pairs), and one
 /// batched note write. All per-commit work below is pure in-memory.
-pub fn handle_revert_commits(repo: &Repository, specs: &[RevertSpec]) -> Result<(), GitAiError> {
+pub(crate) fn handle_revert_commits(
+    repo: &Repository,
+    specs: &[RevertSpec],
+) -> Result<Vec<RewriteMetricCommit>, GitAiError> {
     if specs.is_empty() {
-        return Ok(());
+        return Ok(Vec::new());
     }
 
     // Resolve every spec's parent_sha (only those missing need a lookup) and the
@@ -36,6 +40,7 @@ pub fn handle_revert_commits(repo: &Repository, specs: &[RevertSpec]) -> Result<
         revert_commit: String,
         parent_sha: String,
         source_base_sha: String,
+        original_shas: Vec<String>,
     }
 
     // Collect the revspecs we need resolved, deduplicated.
@@ -84,6 +89,7 @@ pub fn handle_revert_commits(repo: &Repository, specs: &[RevertSpec]) -> Result<
                     revert_commit: spec.revert_commit.clone(),
                     parent_sha,
                     source_base_sha,
+                    original_shas: vec![rc.to_string()],
                 });
             }
             _ => {
@@ -93,6 +99,7 @@ pub fn handle_revert_commits(repo: &Repository, specs: &[RevertSpec]) -> Result<
                     parent_sha,
                     // Placeholder; filled from the second batch below.
                     source_base_sha: String::new(),
+                    original_shas: Vec::new(),
                 });
             }
         }
@@ -105,13 +112,16 @@ pub fn handle_revert_commits(repo: &Repository, specs: &[RevertSpec]) -> Result<
                     .get(&format!("{}^1", r.parent_sha))
                     .cloned()
                     .unwrap_or_default();
+                if r.original_shas.is_empty() && !r.source_base_sha.is_empty() {
+                    r.original_shas.push(r.source_base_sha.clone());
+                }
             }
         }
     }
 
     resolved.retain(|r| !r.source_base_sha.is_empty());
     if resolved.is_empty() {
-        return Ok(());
+        return Ok(Vec::new());
     }
 
     // Batch-read all source notes in one call.
@@ -148,6 +158,7 @@ pub fn handle_revert_commits(repo: &Repository, specs: &[RevertSpec]) -> Result<
 
     // Per-commit reconstruction is now pure in-memory.
     let mut writes: Vec<(String, String)> = Vec::new();
+    let mut metric_commits: Vec<RewriteMetricCommit> = Vec::new();
     for (i, r) in resolved.iter().enumerate() {
         let Some(shift) = shift_idx[i] else {
             continue;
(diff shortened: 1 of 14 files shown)


src/authorship/rewrite_revert.rs
```diff
@@ -105,13 +112,16 @@ pub fn handle_revert_commits(repo: &Repository, specs: &[RevertSpec]) -> Result<
                     .get(&format!("{}^1", r.parent_sha))
                     .cloned()
                     .unwrap_or_default();
+                if r.original_shas.is_empty() && !r.source_base_sha.is_empty() {
+                    r.original_shas.push(r.source_base_sha.clone());
+                }
```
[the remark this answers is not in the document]
svarlamov: Fixed in e1aaa7a0e. The legacy fallback now records the reverted commit (`parent_sha`) in `original_shas`; `source_base_sha` stays only as the note source/base. I added unit coverage for the fallback helper as well.

src/metrics/events.rs
```diff
@@ -305,6 +305,284 @@ impl EventValues for CommittedValues {
...
+    pub const AI_ADDITIONS: usize = 5;
+    pub const AI_ACCEPTED: usize = 6;
+    // Positions 7-10 intentionally unused.
+    pub const COMMIT_SUBJECT: usize = 11;
+    pub const COMMIT_BODY: usize = 12;
+    pub const AUTHORSHIP_NOTE: usize = 13;
+    pub const HUNKS: usize = 14;
+    pub const OPERATION_KIND: usize = 15;
+    pub const ORIGINAL_COMMIT_SHAS: usize = 16;
+}
```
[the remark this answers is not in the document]
svarlamov: Addressed in e1aaa7a0e by clarifying the schema comments. The gap is intentional layout alignment with the committed-event ingestion shape, and position 10 is specifically omitted because rewrite metrics do not have `first_checkpoint_ts`.

src/daemon/rewrite_metrics.rs
```diff
@@ -0,0 +1,255 @@
...
+    let human_author = repo.effective_author_identity().formatted_or_unknown();
+    let attrs = commit_metric_attrs(
+        repo,
+        &metric_commit.new_sha,
+        &base_commit_attr,
+        &human_author,
+    );
+
+    Ok(Some(MetricEvent::from_values(values, attrs.to_sparse())))
+}
```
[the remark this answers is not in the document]
svarlamov: Acknowledged. This cost is deliberate for the first version: the work is feature-gated, release-default false, scheduled only after the rewrite attribution side effects complete, and isolated to one fire-and-forget background worker so it does not block the daemon's rewrite flow. The same-shape event requires post-rewrite diff/hunk/stat calculation against the new commit data. I kept emission sequential inside the worker to avoid fanout; if runtime pressure shows up once this is enabled, batching/throttling the git subprocess calls is the next lever.

Siddhant-K-code (commented):
Resolved the merge conflict in `src/daemon.rs` (both the amend and reset paths) by layering the new rewrite-metrics emission on top of the latest `main` refactor — kept `main`'s recovery-timestamp amend flow and match-guard reset structure, and wired in the metrics spawn.

Verified locally:
- `task build`, `task lint`, `task fmt` all clean
- Full rewrite-related test suites pass (rebase / cherry-pick / amend / reset + new unit tests)
- End-to-end: a real divergent rebase migrates AI attribution to the new commit SHA correctly, and the daemon processes the rewrite with no errors. The metrics path no-ops gracefully when no authorship note is present.

A few non-blocking observations inline.

src/daemon/rewrite_metrics.rs
```diff
@@ -0,0 +1,305 @@
...
+        None => return Ok(None),
+    };
+    let authorship_log = match AuthorshipLog::deserialize_from_string(&raw_note) {
+        Ok(log) => log,
+        Err(_) => return Ok(None),
+    };
+
+    let commit = repo.find_commit(metric_commit.new_sha.clone())?;
+    let parent_count = commit.parent_count()?;
+    if parent_count > 1 {
```
Siddhant-K-code: Merge commits (`parent_count > 1`) are skipped here, so squash-merge results that land as a merge commit will not emit a rewrite metric. If that is intentional (squash merges normally produce a single-parent commit) it is fine — worth a one-line comment so the next reader does not assume it is a gap.

src/authorship/rewrite_revert.rs
```diff
@@ -105,13 +116,18 @@ pub fn handle_revert_commits(repo: &Repository, specs: &[RevertSpec]) -> Result<
                     .get(&format!("{}^1", r.parent_sha))
                     .cloned()
                     .unwrap_or_default();
+                if r.original_shas.is_empty()
+                    && let Some(original_sha) = legacy_revert_metric_original_sha(&r.parent_sha)
+                {
+                    r.original_shas.push(original_sha);
```
Siddhant-K-code, line 123: On the legacy revert path (no explicit `reverted_commit`), `original_shas` is set to `parent_sha` — the revert commit's parent, not the commit that was actually reverted. For metrics whose semantics are "the original commit SHAs," this is an approximation. Probably acceptable since the true reverted SHA is unknown here, but a brief comment noting it is a best-effort fallback would prevent confusion when this field is consumed downstream.

Siddhant-K-code:
### End-to-end execution results (dev backend)

Built this branch, installed it, and ran real rewrite flows (rebase, cherry-pick, amend, revert) against the dev backend with AI-attributed commits. Tested both `git_notes` and `http` notes backends.

**Client side works correctly.** With a real (non-mock) tool attribution, the `rewrite_committed` (event_id 7) metric is emitted with the expected shape on rebase, cherry-pick, and amend. Example payload for a rebase:

```
op_kind        = "rebase"
original_shas  = ["<pre-rewrite SHA>"]
tool_model_pairs = ["all", "codex::gpt-5"]
ai_additions   = [1, 1]
ai_accepted    = [1, 1]
human_additions = 0
commit_subject = "feat: ai line"
authorship_note + hunks = populated
```

Attribution also migrates correctly across the rewrite (AI line stays attributed to the new commit SHA).

**Blocking concern — backend does not accept event_id 7 yet.** Every emitted event 7 was rejected by the dev backend on upload:

```
metrics upload batch result ... delivered_events=0 errored_events=1
errors=[MetricsUploadError { error: "Invalid event_id: 7" }]
```

So with the flag on, this produces a steady stream of upload failures and undelivered metrics. Please confirm the ingestion side supports event_id 7 (and the position schema, incl. the intentionally-omitted position 10) **before** enabling `rewrite_metrics_events` in any environment. Since the release default is `false`, debug/dev builds are the only ones affected today — but the dev backend rejecting them is exactly the case to resolve first.

**Two behavior notes (non-blocking):**
- **Revert** emits no event: reverting an AI commit produces a deletion-only commit with no AI additions, so `metric_tool_model_breakdown` yields nothing and the metric is skipped. Reasonable, but worth confirming this is intended (revert is listed as a covered op in the PR description).
- **mock_ai is filtered**: any commit whose AI attribution is solely `mock_ai::*` is intentionally dropped by `metric_tool_model_breakdown`. Worth a note for anyone testing locally with the mock checkpoints — they will see zero events until a real tool/model is involved.

Notes-backend observation: on the `http` backend, the dev `/worker/notes/` GET returned 401 (while `/worker/metrics/upload` POST returned 200 with the same key), so the synchronous note read in the metric build returned nothing and events were skipped. Switching to the `git_notes` backend (local refs) made the note read succeed and events build. Flagging in case the notes endpoint auth on dev needs a look — this also affects note-dependent flows beyond metrics.

src/authorship/rewrite_revert.rs
Siddhant-K-code: Correction to my own comment above: on a closer read this is actually fine. For a linear revert `A→B→C` where C reverts B, the revert commit C's first parent (`parent_sha`) **is** B — i.e. the reverted commit — so `original_shas = [parent_sha]` records the correct SHA. `source_base_sha` (the grandparent A) is only used as the base for reading the authorship note, not for the metric.

This matches Devin's re-review which marked the same concern as resolved. No change needed here — withdrawing the suggestion.

src/daemon/rewrite_metrics.rs
svarlamov: Squashes are a rewrite event and we should not skip them. It should produce a squash event with the original commit shas->new commit sha. That said, haven't reviewed closely enough to know if a squash merge should be treated any differently. What's your take?

Siddhant-K-code: I think, Squash merges are actually already covered and not skipped here. The `parent_count > 1` guard only excludes true multi-parent merge commits. `git merge --squash` produces a single-parent commit (the merge relationship isn't recorded), so it passes this check.

Squash merges flow through the dedicated `SquashMerge` event path in `daemon.rs` (`handle_rewrite_event_with_metrics` -> `handle_squash_merge`), which emits a `squash_merge` event mapping the original source commit SHAs to the new squash commit SHA, which is the shape you described.

So this `parent_count > 1` branch only fires for regular/octopus merge commits, where there's no single parent to diff against (the stat/hunk pipeline diffs against `parent[0]`) and per-commit attribution wouldn't be meaningful. My original comment conflated "squash merge" with "merge commit". They're distinct, and the squash case is handled. The one-liner I'd suggest is just clarifying that: `// Multi-parent merge commits have no single base to diff against; squash merges are single-parent and handled via the SquashMerge path.`

svarlamov:
### Rewrite metrics local verification

Ran this branch through the requested local verification using the installed dev build.

Setup:
- Ran `task dev` successfully; installed `git-ai 1.6.5 (debug)` and confirmed `git` routes through `~/.git-ai/bin/git`.
- Effective config during correctness run: `rewrite_metrics_events=true`, `notes_backend.kind=git_notes`.
- Validation source was only the local SQLite DB at `~/.git-ai/internal/metrics-db`; server delivery/upload behavior was ignored.

Temp repo correctness run (`rewrite-metrics-manual-1782752141-4135098`):
- Created real `codex::gpt-5` checkpoints, not `mock_ai`.
- Exercised rebase, cherry-pick, squash merge, amend, update-ref, reset/non-fast-forward, and revert.
- SQLite contained 8 `rewrite_committed` rows: 2 rebase rows plus 1 each for `cherry_pick`, `squash_merge`, `amend`, `update_ref`, `non_fast_forward`, and `revert`.
- Verified each row's `commit_sha` and `original_commit_shas` against the recorded old/new SHAs. All mappings matched.
- Verified each row had `tool_model_pairs=["all","codex::gpt-5"]`, populated migrated authorship note, note `base_commit_sha` equal to the new commit, expected AI addition stats, and null subject/body/hunks as currently emitted by the rewrite event path.

React performance run (`react-perf-1782752318-4140224`):
- Cloned `facebook/react` with 7,243 checked-out files.
- Built two equivalent 120-commit AI-attributed stacks and rebased each onto a one-commit advanced base.
- With `rewrite_metrics_events=false`: rebase took 1623 ms and produced 0 rewrite metric rows.
- With `rewrite_metrics_events=true`: rebase took 1666 ms and produced 120 rewrite metric rows.
- Delta: +43 ms / +2.65% for the rebase command.
- The 120 enabled rows appeared in SQLite 635 ms after the rebase returned.
- Verified all 120 React rebase rows map each old stack commit to the corresponding new rebased commit, with no missing rows or bad mappings.

I restored the local config after the performance comparison; effective debug config is back to `rewrite_metrics_events=true` via defaults.

PR #1606, opened 2026-06-23
Add enterprise configuration client support

svarlamov:
## Summary
- add enterprise configuration metadata, validation, fetch response parsing, and API-key scoped cache storage
- apply enterprise overrides separately from user config for compatible policy fields
- bootstrap enterprise config during install-hooks and daemon startup, then refresh in a daemon worker
- fail enterprise API-key installs when the first config fetch cannot complete without a valid cache

## Tests
- task fmt
- task lint
- cargo test --test config_fresh_test
- cargo test commands::install_hooks::tests --lib
- cargo test commands::config::tests --lib

---

<a href="https://app.devin.ai/review/git-ai-project/git-ai/pull/1606" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -323,6 +323,9 @@ fi
 
 echo "Setting up IDE/agent hooks..."
 if ! ${INSTALL_DIR}/git-ai install-hooks; then
+    if [ -n "${API_KEY:-}" ]; then
+        error "Failed to set up enterprise git-ai configuration. Please retry the installer."
+    fi
     warn "Warning: Failed to set up IDE/agent hooks. Please try running 'git-ai install-hooks' manually."
 else
     success "Successfully set up IDE/agent hooks"
--- a/install.ps1
+++ b/install.ps1
@@ -503,8 +503,18 @@ if ($env:INSTALL_NONCE -and $env:API_BASE) {
 Write-Host 'Setting up IDE/agent hooks...'
 try {
     & $finalExe install-hooks | Out-Host
-    Write-Success 'Successfully set up IDE/agent hooks'
+    if ($LASTEXITCODE -ne 0) {
+        if ($env:API_KEY) {
+            Write-ErrorAndExit 'Failed to set up enterprise git-ai configuration. Please retry the installer.'
+        }
+        Write-Warning "Warning: Failed to set up IDE/agent hooks. Please try running 'git-ai install-hooks' manually."
+    } else {
+        Write-Success 'Successfully set up IDE/agent hooks'
+    }
 } catch {
+    if ($env:API_KEY) {
+        Write-ErrorAndExit 'Failed to set up enterprise git-ai configuration. Please retry the installer.'
+    }
     Write-Warning "Warning: Failed to set up IDE/agent hooks. Please try running 'git-ai install-hooks' manually."
 }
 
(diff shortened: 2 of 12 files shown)


Siddhant-K-code (dismissed):
Manual E2E tested the client side against the monorepo enterprise config endpoint.

What I did:
- Saved enterprise config through the web UI.
- Verified the worker endpoint returned the saved config.
- Ran a fresh `git-ai install-hooks` with an org API key against the local web endpoint.
- Checked the generated enterprise cache file.
- Checked effective `git-ai config` values using the same API key.

Expected / observed:
- `install-hooks` fetched and cached `config.enterprise.json`.
- The cache contained the server-provided org config.
- `git-ai config` applied the enterprise overrides correctly:
  - `allow_repositories: ["*github.com/acme/e2e-*"]`
  - `update_channel: "enterprise-next"`
  - `transcript_streaming_lookback_days: 4`

So the full admin-save -> worker endpoint -> client fetch/cache -> effective config path is working end to end.

Siddhant-K-code (approved):
Re-reviewed after rebasing onto current `main`.

Manual E2E coverage from the paired monorepo flow still applies:
- web UI saved enterprise config successfully
- worker endpoint returned the saved config
- fresh `git-ai install-hooks` fetched and cached `config.enterprise.json`
- effective `git-ai config` applied the enterprise overrides from the same org API key

Also verified locally after the rebase:
- `task fmt`
- `cargo check --bin git-ai`

Looks good from the client side; CI is still running on the refreshed head.

Siddhant-K-code:
Screenshot evidence from the localhost E2E run used for the approval review.

This screenshot captures the worker endpoint response and the fresh `git-ai` client effective config after `install-hooks` fetched `config.enterprise.json` from localhost.

Worker endpoint and git-ai client evidence

install.sh
```diff
@@ -323,6 +323,9 @@ fi
 
 echo "Setting up IDE/agent hooks..."
 if ! ${INSTALL_DIR}/git-ai install-hooks; then
+    if [ -n "${API_KEY:-}" ]; then
```
svarlamov, line 326: is this going to log for every user who uses the installer? a bit confused as OSS users will not have API Keys so this is not an error in and of itself

svarlamov: lmk what I'm missing

svarlamov (commented):
Few questions and also (unfortunately) needs to get updated against most recent updates on main ty!

install.sh
Siddhant-K-code: I checked this path. The fatal installer branch is not hit for normal OSS installs:

- `API_KEY` here is the installer-provided enterprise env var, not `GIT_AI_API_KEY` or a missing/default config value.
- `git-ai install-hooks` calls enterprise bootstrap, but when no configured API key exists it returns the `NoApiKey` outcome and exits successfully.
- If hook setup ever exits nonzero without `API_KEY`, the Unix/PowerShell installers still keep the existing best-effort warning path.

@svarlamov - I'd suggest to add a code comment for that

```diff
+    # API_KEY is only supplied by the enterprise installer flow. OSS installs
+    # omit it, so hook setup failures remain a best-effort warning there.
     if [ -n "${API_KEY:-}" ]; then
```

PR #1635, opened 2026-06-24
Document Windows development setup checks

Siddhant-K-code:
## Summary

- Add Windows-specific contributor setup notes for native PowerShell/Windows Terminal vs WSL.
- Clarify that the Task CLI is required for setup commands.
- Clarify that `task dev` installs the branch build, updates the shim, runs install hooks, and restarts the daemon, while `cargo build` is only a compile check.
- Add a focused Windows daemon startup verification flow based on the lessons from #1587.

## Verification

- `git diff --check`
- Attempted `task fmt`, but `task` is not installed in this container.
- Attempted direct fallback `cargo fmt`, but `cargo` is not installed in this container, so `task lint` / clippy could not be run here.


svarlamov (dismissed):
Thank you for this! All nits, but this is docs so I'll be pedantic:

1. Can we just require users to have the task cli (it's easy to get on windows)
2. Can we remove the stuff about missing powershell? I think all developers would realistically need to have powershell

Siddhant-K-code:
@svarlamov - Ack!

Will update it today. Movin it to draft for now!

Siddhant-K-code:
@svarlamov updated per your review:

- Task CLI is now required for setup commands; removed the direct `scripts/dev.ps1` fallback.
- Removed the PowerShell-blocked environment guidance.
- Kept the normal Windows `task dev` setup and daemon verification flow.

The branch is restacked on current `main` as a single docs commit with no co-author trailer.

PR #1644, opened 2026-06-24
Rank bash attribution recovery candidates

Siddhant-K-code:
## Superseded

Closed in favor of #1645.

#1645 covers the same session/workdir/time ranking goal and additionally adds the bash-call interval guard plus integration coverage for manual edits before/after unrelated bash calls.

The installed-build scenario validation from this PR remains useful context, but the implementation should converge on #1645 rather than keeping two overlapping ranking PRs open.

### DIFF
--- a/docs/bash-attribution-scenarios.md
+++ b/docs/bash-attribution-scenarios.md
@@ -1,0 +1,77 @@
+# Bash Attribution Recovery Scenarios
+
+## Manual Installed-Build Matrix
+
+| Scenario | Expected result |
+| --- | --- |
+| Bash hook from worktree A writes a new file in sibling worktree B, then commit in B | New file lines are attributed to the agent session |
+| Same cross-worktree bash write, then a later human-created file is committed in B | Bash-created file is attributed to the agent; later human file remains human/untracked |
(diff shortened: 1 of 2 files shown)


Siddhant-K-code:
Addressed in `20aaf78d0`: `candidate_workdir_contains_path` now uses `canonicalize_path_or_parent` for both the candidate workdir and target path, so nonexistent target files still compare against the canonicalized parent directory. I also updated the parent-workdir ranking test to create the target parent directory, which covers the macOS `/var` -> `/private/var` style mismatch this comment calls out.

Verified locally with:
- `task lint`
- `task test TEST_FILTER=bash_candidate_ranking NO_CAPTURE=true`
- `task test TEST_FILTER=bash_attribution NO_CAPTURE=true`
- installed-build ranking scenarios after `task dev`

Siddhant-K-code:
Closing this in favor of #1645.

#1645 covers the same session/workdir/time ranking goal, and it adds the stricter bash-call interval guard plus integration coverage for manual edits before/after unrelated bash calls. That makes it the better candidate to carry forward for this fix. The installed-build scenario results from this PR are still useful validation context, but the implementation should converge on #1645 rather than keeping two overlapping ranking PRs open.

PR #1645, opened 2026-06-24
Rank bash attribution recovery candidates

Siddhant-K-code:
## Summary

Follow-up to #1625.

This PR makes bash mtime attribution recovery more precise when multiple bash sessions are plausible candidates for the same committed file.

Ranking now applies these rules before falling back to raw timestamp distance:

1. Prefer a bash session whose generated GitAI session ID already appears in the commit authorship log.
2. If no candidate session already appears in the commit, prefer a candidate whose recorded bash workdir is a parent of the repo workdir being recovered.
3. If neither signal applies, pick the closest candidate by file mtime/ctime distance within the recovery window.

It also requires the file timestamp to fall inside the matched bash call interval. For unclosed calls, the interval is bounded to `start_time_ns + BASH_RECOVERY_WINDOW_NS`. This protects manual edits immediately before or after an unrelated bash call from being recovered as AI just because they landed inside the old symmetric +/-3 second query window.

## Expected behavior

- If the commit already contains AI-attributed lines from one session, recovered bash holes prefer that same session over a closer unrelated session.
- If no existing commit session matches, a candidate whose bash workdir is an ancestor of the target repo workdir wins over an unrelated closer candidate.
- If there is no session or workdir signal, recovery keeps the existing closest-by-time fallback.
- Manual edits before or after an unrelated no-op bash call stay unattributed/human.
- PR #1625's main recovery cases still pass: post-snapshot failure, cross-repo/workdir bash recovery, dirty-before-bash recovery, shifted dirty-line recovery, and edge extension.

## Implementation notes

- Adds a ranked `BashCandidateSelection` in `src/authorship/attribution_recovery.rs`.
- Adds `selection_tier` and `candidate_count` to the recovery metadata JSON for observability.
- Adds unit tests for each ranking tier and the time-range guard.
- Adds integration guardrails for manual edits before/after unrelated bash calls.

## Verification

Ran:

```bash
cargo test attribution_recovery --lib -- --nocapture
cargo test --test integration bash_attribution -- --nocapture
cargo fmt --check
git diff --check
```

Results:

- `cargo test attribution_recovery --lib -- --nocapture`: 8 passed.
- `cargo test --test integration bash_attribution -- --nocapture`: 12 passed.
- `cargo fmt --check`: passed.
- `git diff --check`: passed.

## How to test manually

1. Create an AI-authored line in a commit from session A, then create an unattributed bash-recoverable line while session B has a closer timestamp. Expected: recovery chooses session A.
2. Run a bash checkpoint from a parent directory of the target repo and a closer unrelated checkpoint elsewhere. Expected: recovery chooses the parent-workdir checkpoint.
3. Run a no-op bash checkpoint, then manually create a file after the bash call and commit it. Expected: the manual line remains unattributed/human.
4. Manually create a file before a no-op bash checkpoint and commit it. Expected: the manual line remains unattributed/human.

### DIFF
--- a/src/authorship/attribution_recovery.rs
+++ b/src/authorship/attribution_recovery.rs
@@ -16,6 +16,7 @@ use std::path::Path;
 use std::time::{SystemTime, UNIX_EPOCH};
 
 const BASH_RECOVERY_WINDOW_NS: u128 = 3_000_000_000;
+const BASH_RECOVERY_COARSE_TIMESTAMP_NS: u128 = 1_000_000_000;
 const EDGE_EXTENSION_MAX_LINES: usize = 3;
 
 pub(crate) fn recover_attribution(
@@ -360,8 +361,19 @@ fn recovery_distance_to_call_window(timestamp_ns: u128, call: &BashCheckpointCal
         .end_time_ns
         .unwrap_or_else(|| start.saturating_add(BASH_RECOVERY_WINDOW_NS));
 
-    if timestamp_ns < start || timestamp_ns > end {
+    if timestamp_ns > end {
         return None;
+    }
+
+    if timestamp_ns < start {
+        let start_skew_ns = start.saturating_sub(timestamp_ns);
+        // Low-resolution filesystems can truncate mtimes to whole seconds.
+        // Only grant start-side grace to timestamps that look truncated.
+        if start_skew_ns >= BASH_RECOVERY_COARSE_TIMESTAMP_NS
+            || timestamp_ns % BASH_RECOVERY_COARSE_TIMESTAMP_NS != 0
+        {
+            return None;
+        }
     }
 
     Some(distance_to_call_window(timestamp_ns, call))
@@ -802,6 +814,29 @@ mod tests {
             selection.is_none(),
             "nearby candidates outside their matching bash time range must not recover attribution"
         );
+    }
+
+    #[test]
+    fn bash_candidate_ranking_allows_coarse_timestamp_rounded_before_start() {
+        let candidates = vec![bash_call(
+            1,
+            "coarse-session",
+            "tool-coarse",
+            "/repo",
+            2_500_000_000,
+            Some(3_000_000_000),
+        )];
+
+        let selection = select_best_bash_candidate(
+            &candidates,
+            &[2_000_000_000],
+            &HashSet::new(),
+            "/repo",
+        )
+        .expect("expected coarse timestamp to match");
+
+        assert_eq!(selection.candidate.tool_use_id, "tool-coarse");
+        assert_eq!(selection.distance_ns, 500_000_000);
     }
 
     #[test]
--- a/src/authorship/attribution_recovery.rs
+++ b/src/authorship/attribution_recovery.rs
@@ -12,6 +12,7 @@ use crate::metrics::{CheckpointValues, EventAttributes, MetricEvent, PosEncoded}
 use serde_json::json;
 use std::collections::{BTreeMap, HashMap, HashSet};
 use std::fs;
+use std::path::Path;
 use std::time::{SystemTime, UNIX_EPOCH};
 
 const BASH_RECOVERY_WINDOW_NS: u128 = 3_000_000_000;
@@ -92,14 +93,21 @@ fn recover_bash_mtime(
         return Ok(());
     }
 
+    let existing_commit_sessions = existing_commit_session_ids(authorship_log);
     for (file_path, unknown_lines) in unknown_by_file {
         let Some(timestamps) = timestamps_by_file.get(&file_path) else {
             continue;
         };
-        let Some((candidate, distance_ns)) = select_best_bash_candidate(&candidates, timestamps)
-        else {
+        let Some(selection) = select_best_bash_candidate(
+            &candidates,
+            timestamps,
+            &existing_commit_sessions,
+            &repo_work_dir,
+        ) else {
             continue;
         };
+        let candidate = selection.candidate;
+        let distance_ns = selection.distance_ns;
         if distance_ns > BASH_RECOVERY_WINDOW_NS {
             continue;
         }
@@ -126,6 +134,8 @@ fn recover_bash_mtime(
             "end_time_ns": candidate.end_time_ns,
             "start_trace_id": candidate.start_trace_id,
             "end_trace_id": candidate.end_trace_id,
+            "selection_tier": selection.tier.as_str(),
+            "candidate_count": candidates.len(),
         });
         record_recovery_metric(RecoveryMetricInput {
             repo,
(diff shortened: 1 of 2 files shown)


src/authorship/attribution_recovery.rs
```diff
@@ -249,26 +259,137 @@ fn system_time_to_ns(time: SystemTime) -> u128 {
...
+    let end = call
+        .end_time_ns
+        .unwrap_or_else(|| start.saturating_add(BASH_RECOVERY_WINDOW_NS));
+
+    if timestamp_ns < start || timestamp_ns > end {
+        return None;
+    }
+
+    Some(distance_to_call_window(timestamp_ns, call))
+}
```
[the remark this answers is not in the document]
Siddhant-K-code: Addressed in 86e6a72ec022b3c40560f846bf60d6ce6cf34148. `recovery_distance_to_call_window` now keeps the strict end-boundary check, but allows a narrow start-side grace only for timestamps that look like whole-second filesystem truncation and are less than one second before the bash start. That should cover the low-resolution mtime false-negative without reopening the old symmetric before/after recovery window. I also added `bash_candidate_ranking_allows_coarse_timestamp_rounded_before_start` to pin this case down.

src/authorship/attribution_recovery.rs
```diff
@@ -249,26 +260,148 @@ fn system_time_to_ns(time: SystemTime) -> u128 {
...
+    }
+}
+
+fn bash_candidate_tier(
+    candidate: &BashCheckpointCall,
+    existing_commit_sessions: &HashSet<String>,
+    target_repo_work_dir: &str,
+) -> BashCandidateTier {
+    let session_id = bash_candidate_session_id(candidate);
+    if existing_commit_sessions.contains(&session_id) {
```
svarlamov, line 343: **Over-attribution snowball (commit-wide, path-blind tier).** `existing_commit_session_ids` collects every AI session attested *anywhere* in the commit, and `bash_candidate_tier` promotes any candidate whose session is in that set to tier-0 — with no requirement that the bash call touched this file/path (a `BashCheckpointCall` records no edited-file list, and `candidates_near_timestamps` filters on time only). Since tier dominates `distance_ns` in `min_by`, a genuinely-human untracked hole in `b.rs` that merely falls within the 3s window of an unrelated no-op bash call (`ls`, `git status`) from session S — which authored some *other* file `a.rs` in the same commit — gets attributed as AI to S, even beating a closer-in-time correct candidate. One legitimate session becomes a magnet that absorbs nearby untracked holes across the whole commit.

svarlamov (commented):
Automated review of the bash-attribution-recovery ranking. Findings inline, most severe first. The strongest is the commit-wide, path-blind ExistingCommitSession tier (over-attribution snowball).

src/authorship/attribution_recovery.rs
```diff
@@ -249,26 +260,148 @@ fn system_time_to_ns(time: SystemTime) -> u128 {
...
+        return BashCandidateTier::ExistingCommitSession;
+    }
+
+    if path_is_equal_or_child(target_repo_work_dir, &candidate.repo_work_dir) {
```
svarlamov, line 347: **WorkdirAncestor tier is defeated by asymmetric path normalization.** The target is canonicalized — `repo_worktree_key` calls `.canonicalize()` — but `candidate.repo_work_dir` is stored raw (`worktree_root_for_path` walks parents without canonicalizing; the daemon stores `repo_root.to_string_lossy()`). `path_is_equal_or_child` does a plain lexical `starts_with`. On macOS (`/var` → `/private/var` symlink) or any repo under a symlinked path (including the test framework's temp dirs), target `/private/var/.../repo` vs candidate `/var/.../repo` fails the comparison, so a same-repo bash call is silently demoted from WorkdirAncestor to TimeOnly and can lose to a closer unrelated-repo candidate. The PR's tests use abstract `/repo` paths that never exercise canonicalization.

src/authorship/attribution_recovery.rs
```diff
@@ -249,26 +260,148 @@ fn system_time_to_ns(time: SystemTime) -> u128 {
...
+
+    if timestamp_ns > end {
+        return None;
+    }
+
+    if timestamp_ns < start {
+        let start_skew_ns = start.saturating_sub(timestamp_ns);
+        // Low-resolution filesystems can truncate mtimes to whole seconds.
+        // Only grant start-side grace to timestamps that look truncated.
+        if start_skew_ns >= BASH_RECOVERY_COARSE_TIMESTAMP_NS
```
svarlamov, line 372: **Coarse-timestamp start grace is miscalibrated to 1s vs the codebase's 2s.** `bash_tool.rs` uses `MTIME_GRACE_WINDOW_SECS = 2` for low-resolution filesystems (HFS+/FAT truncate mtime to 2s). Here the start-side skew bound is hardcoded to `BASH_RECOVERY_COARSE_TIMESTAMP_NS = 1s`. A truncated mtime on a 2s-resolution FS that lands between 1s and 2s before the recorded `start_time_ns` has `start_skew_ns >= 1s`, so this returns `None` and drops a legitimate recovery candidate — exactly the 2s-resolution case the existing 2s grace was sized for. Consider reusing the established constant.

src/authorship/attribution_recovery.rs
```diff
@@ -249,26 +260,148 @@ fn system_time_to_ns(time: SystemTime) -> u128 {
...
+    generate_session_id(&candidate.agent_id.id, &candidate.agent_id.tool)
+}
+
+fn recovery_distance_to_call_window(timestamp_ns: u128, call: &BashCheckpointCall) -> Option<u128> {
+    let start = call.start_time_ns;
+    let end = call
+        .end_time_ns
+        .unwrap_or_else(|| start.saturating_add(BASH_RECOVERY_WINDOW_NS));
+
+    if timestamp_ns > end {
```
svarlamov, line 364: **Asymmetric window: `ts > end` is a hard reject with no end-side grace.** `candidates_near_timestamps` still admits DB rows up to +window after `end` (via `distance_to_call_window`), but this function rejects any timestamp past `end` outright, while granting coarse grace on the start side. A file whose mtime is flushed/updated just after a bash process exits (or truncated up past `end` on a coarse FS) is now silently dropped — a recovery that worked under #1625. The start side gets 1s grace; the end side gets zero. Worth making the two sides symmetric (or documenting why the end side is intentionally strict).

src/authorship/attribution_recovery.rs
```diff
@@ -249,26 +260,148 @@ fn system_time_to_ns(time: SystemTime) -> u128 {
...
+        // Low-resolution filesystems can truncate mtimes to whole seconds.
+        // Only grant start-side grace to timestamps that look truncated.
+        if start_skew_ns >= BASH_RECOVERY_COARSE_TIMESTAMP_NS
+            || !timestamp_ns.is_multiple_of(BASH_RECOVERY_COARSE_TIMESTAMP_NS)
+        {
+            return None;
+        }
+    }
+
+    Some(distance_to_call_window(timestamp_ns, call))
```
svarlamov, line 379: **Open-call admission and distance use different `end` definitions.** For `end_time_ns == None`, this function admits when `ts <= start + 3s` (synthetic end), but the returned distance comes from `distance_to_call_window`, which uses `end = start` for open calls — so a ts 2.9s after start is admitted yet scored with distance ≈2.9s. A still-running session that legitimately authored a file 1–3s after its start is then out-scored on distance by any completed call whose real window brackets the timestamp (distance 0), biasing attribution away from in-flight sessions toward the wrong completed call. Consider scoring open calls against the same synthetic window used for admission.

src/authorship/attribution_recovery.rs
```diff
@@ -249,26 +260,148 @@ fn system_time_to_ns(time: SystemTime) -> u128 {
...
+            let session = ai_session_key(&entry.hash);
+            if session.starts_with("s_") {
+                sessions.insert(session.to_string());
+            }
+        }
+    }
+    sessions
+}
+
+fn path_is_equal_or_child(child: &str, parent: &str) -> bool {
```
svarlamov, line 395: **Reuse:** `path_is_equal_or_child` is a weaker re-implementation of existing path-ancestor logic (`Repository::path_is_in_workdir` in `src/git/repository.rs`, which handles canonicalization/symlinks). Two divergent definitions of "path under path" will drift; reusing the canonicalizing helper would also fix the WorkdirAncestor normalization mismatch noted above.

src/authorship/attribution_recovery.rs
```diff
@@ -249,26 +260,148 @@ fn system_time_to_ns(time: SystemTime) -> u128 {
...
+
+    if path_is_equal_or_child(target_repo_work_dir, &candidate.repo_work_dir) {
+        return BashCandidateTier::WorkdirAncestor;
+    }
+
+    BashCandidateTier::TimeOnly
+}
+
+fn bash_candidate_session_id(candidate: &BashCheckpointCall) -> String {
+    generate_session_id(&candidate.agent_id.id, &candidate.agent_id.tool)
```
svarlamov, line 355: **Efficiency:** `bash_candidate_session_id` runs `generate_session_id` (SHA-256 + allocations) for every candidate, and `select_best_bash_candidate` is called once per unknown file inside `recover_bash_mtime`'s loop. Since each candidate's session id (and indeed its full tier — `existing_commit_sessions` and `target_repo_work_dir` are file-invariant) doesn't change across files, this is O(files × candidates) redundant hashing. Precompute the per-candidate tier once before the file loop.

PR #1652, opened 2026-06-24
Speed up `git-ai stats` range mode by batching authorship-note reads

jwiegley:
## Summary

Makes `git-ai stats <a>..<b>` (range mode) faster — **~1.7× on a 100-commit range up to ~2.5× on a 1000-commit range** — by removing redundant git-subprocess fan-out in the blame path. `range_stats` output is byte-identical.

Profiling (release build, real repo) showed the cost is **git subprocesses, not SQLite** — the default `GitNotes` backend touches **0 ms of SQLite** on this path, so a #1630-style index would not help `git-ai stats`. The range path blames every changed file **three times** (`diff_ai_accepted_stats` + start/end `VirtualAttributions`), and each blame re-read every blamed commit's authorship note with a separate `git notes --ref=ai show` — thousands of per-commit note spawns, uncached across files and passes.

## Changes (both output-preserving)

1. **Batch the note reads.** New `notes_api::read_authorship_v3_batch` resolves many commits' notes with one `git ls-tree` + `cat-file --batch` (via `refs::notes_for_commits`) instead of one `git notes show` per commit. It returns exactly what `read_authorship_v3(repo, sha).ok()` yields per commit — the v3 parse is factored into `refs::parse_reference_as_authorship_log_v3` and shared by both paths, and the `Http` backend delegates per commit to preserve its cache-hit semantics. Both blame overlays (`overlay_ai_authorship`, `populate_ai_human_authors`) pre-seed their existing per-commit cache from one batched read.
2. **Skip dead work on the stats path.** `blame()` with `no_output` discards the blame hunks, so the `populate_ai_human_authors` pass that annotates them with `ai_human_author` (a second per-commit note read) is wasted there. Adds `GitAiBlameOptions::skip_human_author_population` (default `false`, debug-asserted to only be set on `no_output` paths) and sets it on the two stats blame call sites. It is output-invariant because `overlay_ai_authorship` derives `line_authors` per line independently of hunk grouping.

## Performance (release build, real repo, same-run A/B, `range_stats` byte-identical)

The speedup grows with range size because it eliminates per-commit note fan-out, which scales with the number of blamed commits:

| command | baseline (same run) | optimized | speedup |
|---|---|---|---|
| `git-ai stats HEAD~100..HEAD` | 69.0 s (noisy: 69–93 s across runs) | 40.0 s | ~1.7× |
| `git-ai stats HEAD~1000..HEAD` | 272.3 s | 110.2 s | ~2.5× |

Under a subprocess-counting harness on `HEAD~100..HEAD`, per-commit `git notes show` collapses from ~7,200 to ~50 (the remainder shifts into batched `cat-file`/`ls-tree`); total git spawns ~14.5k → ~9.3k same-harness. (Exact spawn totals are method-sensitive; the `notes show` collapse and the wall-time deltas are the robust figures.)

## Validation

- `task fmt`, `task lint` (clippy `-D warnings`), full `task test` — green.
- Same-run A/B byte-identical check of `range_stats` on real data across 100- and 1000-commit ranges (range_1000 = 120,390 AI lines across 32 tool/model combos — identical before/after).
- Regression tests: batched read ≡ per-commit read (commits with notes, without notes, non-existent); empty-input fast path; and skip-dead-pass gate output-invariance (blame line authors identical with the pass on vs off).
- Two independent adversarial reviews (the batching, and the skip-dead-pass gate) — no blockers. Documented parity boundary: on GitNotes the batched read decodes note blobs with `from_utf8_lossy` (matching the existing `CommitAuthorship` path) vs the per-commit strict decode — unreachable for real notes (always valid UTF-8 JSON). The `Http` and batch-error fallback arms delegate per commit to the existing reader (equivalent by construction; not exercised by tests).

## Note

The #1630 index pattern still applies — just to a different command, `git-ai usage` (`MetricsDatabase::get_metric_history` is a full-table scan that deserializes every row). Not touched here.

### DIFF
--- a/src/authorship/diff_ai_accepted.rs
+++ b/src/authorship/diff_ai_accepted.rs
@@ -49,6 +49,9 @@ pub fn diff_ai_accepted_stats(
             options.line_ranges = line_ranges;
             options.no_output = true;
             options.use_prompt_hashes_as_names = true;
+            // Only the per-line authors are used below; the blame hunks are discarded,
+            // so skip the ai_human_author population pass (a redundant per-commit note read).
+            options.skip_human_author_population = true;
         }
 
         let blame_result = repo.blame(&file_path, &options);
--- a/src/authorship/virtual_attribution.rs
+++ b/src/authorship/virtual_attribution.rs
@@ -3412,6 +3412,9 @@ fn compute_attributions_for_file(
         ai_blame_opts.newest_commit = Some(base_commit.to_string());
         ai_blame_opts.oldest_commit = blame_start_commit;
         ai_blame_opts.oldest_date = Some(*OLDEST_AI_BLAME_DATE);
+        // Only the per-line blame authors are consumed below; the hunks are discarded,
+        // so skip the ai_human_author population pass (a redundant per-commit note read).
+        ai_blame_opts.skip_human_author_population = true;
     }
 
     // Run blame at the base commit
--- a/src/commands/blame.rs
+++ b/src/commands/blame.rs
@@ -170,6 +170,15 @@ pub struct GitAiBlameOptions {
     // When true, a single git blame hunk may be split into multiple hunks
     // if different lines were authored by different humans working with AI
     pub split_hunks_by_ai_author: bool,
+
+    // Skip populating `ai_human_author` on blame hunks (the `populate_ai_human_authors`
+    // pass). That pass reads each blamed commit's authorship note and runs per-line
+    // attribution purely to annotate hunks. Callers that consume only `line_authors` /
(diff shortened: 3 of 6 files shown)


jwiegley:
* **#1652** <a href="https://app.graphite.com/github/pr/git-ai-project/git-ai/1652?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/git-ai-project/git-ai/1652?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1666, opened 2026-06-26
feat(config): CLI coverage for allow_superuser, transcript lookback, custom_attributes

svarlamov:
## Summary

Three persisted `FileConfig` fields had **no** `git-ai config` CLI handling — they could only be configured by hand-editing `~/.git-ai/config.json` or via env vars:

- `allow_superuser` (bool) — allow running git-ai as root/superuser
- `transcript_streaming_lookback_days` (`u32`, `0` = unlimited) — how far back transcript sweeps look
- `custom_attributes` (`string → string` object) — custom telemetry attributes, plus nested `custom_attributes.<key>`

This PR wires all three through the CLI's `show-all`, `get`, `set`, and `unset` paths (plus the nested get/set/unset path for `custom_attributes`), matching the conventions of the existing fields.

### Notable details

- **Restart hint**: `git-ai config set transcript_streaming_lookback_days <n>` now prints the same `Run \`git-ai bg restart\`` hint as the transcript feature flags, because the daemon reads this value once at worker startup (`Config::get()`).
- **`0 = unlimited`**: the runtime normalizes `0` → `None`; show-all/get surface unlimited as `0` so it round-trips through `set`.
- **`custom_attributes` value coercion**: object values accept string/number/bool (coerced to strings), mirroring how the `GIT_AI_CUSTOM_ATTRIBUTES` env override is parsed at config load.

## Regression guard (cannot silently regress)

`tests/integration/config_cli_coverage.rs` adds a guard that **derives the field list from a fully-populated `FileConfig` via serde** (not a hardcoded list) and asserts every field is reachable through the *real* CLI:

- `test_every_file_config_field_has_cli_get_coverage` — `get_config_value` has an arm for every field + an "Unknown config key" catch-all, making it the canonical completeness check.
- `test_every_file_config_field_has_cli_unset_coverage` — exercises the write-handler catch-all (`notes_backend` is documented as nested-only via an explicit allowlist).

Adding a new persisted `FileConfig` field without CLI wiring now fails the suite. Verified by temporarily deleting the `allow_superuser` arm — the guard fails with a precise pointer to the missing handler.

Also covers set/get/unset roundtrips for the three new keys (incl. `--add` upsert and invalid-input rejection).

## Testing

- `task fmt` ✅
- `task lint` ✅ (clippy `-D warnings`)
- `task test TEST_FILTER=config` ✅ (all config unit + integration tests pass)
- New unit tests for `parse_custom_attributes_object` + 7 integration tests in `config_cli_coverage`

### DIFF
--- a/src/commands/config.rs
+++ b/src/commands/config.rs
@@ -115,6 +115,11 @@ fn print_config_help() {
     println!("  include_prompts_in_repositories  Repos to include for prompt storage (array)");
     println!("  default_prompt_storage       Fallback storage mode for non-included repos");
     println!("  quiet                        Suppress chart output after commits (bool)");
+    println!("  allow_superuser              Allow running git-ai as root/superuser (bool)");
+    println!(
+        "  transcript_streaming_lookback_days  Days to look back when sweeping transcripts (0 = unlimited)"
+    );
+    println!("  custom_attributes            Custom telemetry attributes, string->string (object)");
     println!("  git_ai_hooks                 Hook name -> shell commands map (object)");
     println!("  codex_hooks_format           Codex hook install format (config_toml/hooks_json)");
     println!("  notes_backend.kind           Notes backend kind (git_notes/http)");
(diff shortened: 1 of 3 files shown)


svarlamov:
@devin-ai-integration please review this PR.

svarlamov:
Ran a thorough self-review of the diff (8 finder angles + verification) since automated review hadn't posted. Addressed three real findings in 0be8a3f:

1. **Empty-object normalization** — `config set custom_attributes '{}'` now stores `None` (omits the key from `config.json`) instead of persisting a redundant `{}`, matching the existing `author`/`git_ai_hooks` convention.
2. **Trim asymmetry** — nested `unset custom_attributes.<key>` now trims the attribute name like the nested `set` path, so a name set with surrounding whitespace is removable by the same dotted key.
3. **Doc clarity** — `parse_custom_attributes_object` now documents that it shares the `GIT_AI_CUSTOM_ATTRIBUTES` env override's scalar coercion but, unlike the env path (which silently drops non-scalar values), the CLI rejects them so a malformed `set` fails loudly.

Added regression tests for #1 and #2. Other surfaced candidates (shared-helper extraction across `git_ai_hooks`/`custom_attributes`) were intentionally left out of scope to avoid refactoring already-working unrelated code.

PR #1719, opened 2026-07-02
[codex] Fix stash checkpoint storage growth

svarlamov:
## Summary
- store stash attribution in compact versioned `.git/ai/stashes_v2/<stash>/` working-log snapshots instead of copying raw checkpoint history
- remove stashed paths from live `INITIAL` and `checkpoints.jsonl`, then restore by replacing those paths instead of appending checkpoint records
- delete legacy `.git/ai/stashes` on stash operations so old oversized stash data cannot be reintroduced

## Root Cause
Stash create copied the current working log into a per-stash directory, including `checkpoints.jsonl`. Stash apply/pop then appended that checkpoint file back into the live working log. Repeating stash/pop duplicated the same checkpoint history over and over, so later attribution processing had to load a rapidly growing `checkpoints.jsonl` file.

## Verification
- `task test TEST_FILTER=test_repeated_stash_pop_does_not_duplicate_checkpoints`
- `task test TEST_FILTER=test_stash_operation_deletes_legacy_stashes_dir`
- `task test TEST_FILTER=test_stash_push_pathspec_excludes_unstashed_file_from_stash_log`
- `task test TEST_FILTER=stash_attribution`
- `task fmt`
- `task build`
- `task lint`
- `git diff --check`

### DIFF
--- a/src/authorship/rewrite_stash.rs
+++ b/src/authorship/rewrite_stash.rs
@@ -1,11 +1,13 @@
 use std::collections::{BTreeMap, HashMap, HashSet};
 use std::fs;
+use std::io::{BufRead, BufReader, BufWriter, Write};
 use std::path::PathBuf;
 use std::time::{SystemTime, UNIX_EPOCH};
 
 use crate::authorship::attribution_tracker::LineAttribution;
 use crate::authorship::authorship_log::{HumanRecord, PromptRecord, SessionRecord};
 use crate::authorship::imara_diff_utils::{DiffOp, capture_diff_slices};
+use crate::authorship::working_log::{Checkpoint, CheckpointKind};
 use crate::error::GitAiError;
 use crate::git::repo_storage::{InitialAttributions, PersistedWorkingLog};
 use crate::git::repository::{
@@ -42,6 +44,10 @@ fn stash_entry_dir(repo: &Repository, stash_sha: &str) -> PathBuf {
 
 fn stash_metadata_path(repo: &Repository, stash_sha: &str) -> PathBuf {
     stash_entry_dir(repo, stash_sha).join("metadata.json")
+}
+
+fn filtered_stash_working_log_base(stash_sha: &str) -> String {
+    format!("_stash_filter_{}", stash_sha)
 }
 
 fn working_log_for_dir(repo: &Repository, dir: PathBuf, base_commit: &str) -> PersistedWorkingLog {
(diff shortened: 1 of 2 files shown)


Siddhant-K-code:
Reviewing this with the compact-stash approach in mind: I think the direction is good, but I would not merge this as-is.

1. Blocking: `stash pop -q` / `stash apply -q` can lose attribution.

This PR clears the live checkpoint log during stash push and depends on pop/apply restore to put attribution back. But `RefCursor::enrich_stash` still uses `stash_args.get(1)` for pop/apply target resolution, so `git stash pop -q` / `git stash apply -q` treats `-q` as the stash target. That can make stash SHA resolution fail and skip `handle_stash_pop_or_apply_with_head`, so the compact stash data is never restored. Agents commonly use quiet stash commands, so this needs a positional non-flag stash target parser plus regression tests for `pop -q` and `apply -q`.

2. Blocking for the reported incident: first stash after upgrade can still OOM on an already-bloated live log.

`save_stash_attributions` compacts by calling `VirtualAttributions::from_persisted_working_log` before `clean_working_log_for_stash` truncates the live `checkpoints.jsonl`. If a user already has a multi-GB checkpoint file from repeated stash cycles, this path still reads/parses that file and can crash before it self-heals. The compact format is good, but the migration path needs a bounded/streaming fallback or an emergency truncation/recovery path for pathological logs.

3. High: partial-stash filtering is not fully bounded either.

`write_path_filtered_checkpoints` streams into a filtered temp log, but the subsequent compaction step reads that filtered working log through `VirtualAttributions::from_persisted_working_log`. If the selected pathspec carries most of the bloated data, memory use is still unbounded.

Overall: I like the compact stash storage model here more than continuing to move raw checkpoint history around, but I think this PR needs the quiet-flag fix and an already-bloated-log recovery test/mitigation before merge.

Siddhant-K-code:
Follow-up with concrete fixes I think would make this mergeable.

### Recommended path

I like the compact stash format in this PR. To make it production-safe, I would keep this approach but add two missing pieces from the narrower fix in #1720.

### 1. Fix stash target parsing for quiet commands

`git stash pop -q` and `git stash apply -q` currently treat `-q` as the stash target because `RefCursor::enrich_stash` uses positional indexing (`stash_args.get(1)`). Once this PR clears live checkpoints on stash push, missing the restore means attribution is lost.

Action:

```rust
fn stash_positional_arg(stash_args: &[String], position: usize) -> Option<&String> {
    stash_args
        .iter()
        .skip(1)
        .filter(|arg| !arg.starts_with('-'))
        .nth(position)
}
```

Use that in `enrich_stash`:

- `pop` / `apply` / `drop`: `stash_positional_arg(stash_args, 0)`
- `branch`: `stash_positional_arg(stash_args, 1)`

Add tests:

- `test_stash_pop_quiet_flag_restores_attribution`
- `test_stash_apply_quiet_flag_restores_attribution`

### 2. Make compaction safe for already-bloated logs

Right now `save_stash_attributions` calls `VirtualAttributions::from_persisted_working_log` before `clean_working_log_for_stash` truncates the live checkpoint log. If the user already has a multi-GB `checkpoints.jsonl`, this can still load the huge file and crash before the fix self-heals.

Action:

Before calling `VirtualAttributions`, create a temporary compact/deduped working log by streaming the source log:

1. Copy/filter `INITIAL`.
2. Stream `checkpoints.jsonl` line-by-line.
3. Keep only relevant pathspec entries.
4. Deduplicate checkpoint records while streaming.
5. Copy only referenced blobs.
6. Run `VirtualAttributions::from_persisted_working_log` on the temp log.
7. Delete the temp log.

This keeps the compact stash design, but avoids feeding pathological raw checkpoint history into `VirtualAttributions`.

### 3. Add an already-bloated-log regression test

Test shape:

1. Create a repo and an AI checkpoint.
2. Manually duplicate the current working log’s `checkpoints.jsonl` lines many times.
3. Run `git stash push` and `git stash pop`.
4. Assert the command completes, final checkpoint count is bounded, and committed attribution is preserved.

This directly covers the upgrade/recovery case from the reported incident.

### Why this keeps the best part of this PR

#1720 is safer today because it fixes `-q` parsing and truncates full-stash live checkpoints without reading them first. But this PR has the better long-term storage model: stash attribution should be compact path attribution, not raw checkpoint history. With the two fixes above, this approach should be both cleaner and safe for the incident case.

PR #1720, opened 2026-07-02
fix(stash): stop exponential checkpoints.jsonl growth on stash push/pop cycles

svarlamov:
## Problem

Since v1.5.9 (the rewrite-ops merge that introduced content-based stash attribution), repeated `git stash` / `git stash pop` cycles cause **exponential growth** of `.git/ai/working_logs/<head>/checkpoints.jsonl` and daemon RSS explosions (reported 18GB file / 20GB+ RSS after ~20 stash/unstash cycles by a coding agent).

## Root cause

1. **Push copies but never clears.** `save_stash_attributions` copies the whole working log (checkpoints.jsonl + blobs) into `.git/ai/stashes/<sha>_worklog/`, but `clean_working_log_for_stash` only cleared INITIAL — the live `checkpoints.jsonl` kept all N lines.
2. **Pop raw-appends.** `restore_stash_attributions` appended the stash copy onto the never-cleared live file with no dedup: N → 2N → 4N → ... 20 cycles ≈ 10^6×.
3. **RAM amplifier.** `append_checkpoint` reads and rewrites the entire file on every checkpoint, so once the file is GBs every checkpoint costs multi-GB allocations — the observed RSS explosion.

## Fix

All in the async side-effect layer — nothing added to the ingestion path, no new git spawns:

- **`clean_working_log_for_stash`** now truncates the live `checkpoints.jsonl` on a full stash (via `write_all_checkpoints(&[])` *without reading it first*, so a push on an already-bloated log is an O(1) self-heal), and complement-filters it on pathspec'd stash — the exact mirror of `filter_stash_checkpoints_to_pathspecs` on the stash copy.
- **`restore_stash_attributions`** dedups byte-identical lines on append, making repeated `stash apply` and conflict-tolerant re-restores idempotent.
- **Versioned stash dir (`stashes` → `stashes_v2`) + legacy dir deletion** on any stash op. Pre-fix users may carry multi-GB `_worklog` dirs; restoring one would slurp it into RAM. We deliberately sacrifice pre-fix stash attribution data to guarantee those can never be re-appended or loaded again.
- **`merge_working_log_dirs`** dedups checkpoints by `(kind, diff, timestamp, author)` so HEAD-move merges can't duplicate either.
- **Bonus bug found during end-to-end verification:** `git stash pop -q` / `apply -q` never restored attribution at all — `enrich_stash` treated the first arg after the subcommand as the stash target, so boolean flags broke stash-sha resolution and the restore silently no-op'd. Agents typically pass `-q`, which is why the doubling (rather than missing restores) was the visible symptom. Fixed with a flag-skipping positional lookup.

## Testing (TDD)

Every fix landed behind a test that was watched fail first:

- `test_stash_pop_cycles_do_not_grow_checkpoints` — the repro: 5 push/pop cycles, line count must stay constant (was 32× before the fix)
- `test_stash_push_clears_live_checkpoints_and_pop_restores`
- `test_stash_apply_repeated_is_idempotent`
- `test_stash_push_pathspec_keeps_unstashed_checkpoints_in_live_log`
- `test_stash_pop_after_head_move_stays_bounded` (shifted-pop path)
- `test_legacy_stashes_dir_is_removed`
- `test_stash_pop_quiet_flag_restores_attribution` / `test_stash_apply_quiet_flag_restores_attribution`
- `test_merge_working_log_dirs_dedups_identical_checkpoints` (unit)

End-to-end verification with a real installed daemon: 10 files × 300 lines with AI checkpoints, 20 manual `stash push -q` / `pop -q` cycles → `checkpoints.jsonl` byte-identical to baseline (7,931 bytes), daemon RSS flat (3.5MB), zero leftover stash dirs, legacy `stashes/` dir removed, and `git-ai blame` shows full AI attribution after commit.

Full test suite, `task lint`, and `task fmt` are green.

## Deferred (follow-up PRs)

- `append_checkpoint`'s O(n²) read-all/write-all (latency-sensitive checkpoint path; needs a `prune_old_char_attributions` redesign — with the log now staying small it's cheap again)
- `--keep-index` attribution retention (pre-existing INITIAL semantics, unchanged)

---

<a href="https://app.devin.ai/review/git-ai-project/git-ai/pull/1720" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/src/authorship/rewrite_stash.rs
+++ b/src/authorship/rewrite_stash.rs
@@ -1,4 +1,4 @@
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::fs;
 use std::path::PathBuf;
 use std::time::{SystemTime, UNIX_EPOCH};
@@ -19,8 +19,22 @@ pub struct StashMetadata {
     pub pathspecs: Vec<String>,
 }
 
+// Versioned so upgrades can't resurrect pre-fix stash data: v1 ("stashes")
+// could balloon to GBs (push copied the working log without clearing it and
+// pop re-appended it, doubling checkpoints.jsonl per cycle). Restoring such a
+// worklog would slurp it all into memory, so we abandon the old dir entirely.
 fn stashes_dir(repo: &Repository) -> PathBuf {
-    repo.storage.ai_dir.join("stashes")
+    repo.storage.ai_dir.join("stashes_v2")
+}
+
+/// Delete the legacy (pre-versioning) stash dir if present, sacrificing any
+/// attribution stashed by pre-fix versions in exchange for guaranteeing its
+/// potentially multi-GB worklogs are never restored or loaded again.
+fn cleanup_legacy_stashes_dir(repo: &Repository) {
+    let legacy_dir = repo.storage.ai_dir.join("stashes");
+    if legacy_dir.exists() {
+        let _ = fs::remove_dir_all(&legacy_dir);
+    }
 }
 
 fn path_matches_any(path: &str, pathspecs: &[String]) -> bool {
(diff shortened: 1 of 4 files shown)


Siddhant-K-code:
It is superseed by #1719

PR #1722, opened 2026-07-02
fix(stash): quiet-flag stash target resolution + working-log merge dedup + stash cycle regression tests

svarlamov:
## Summary

Follow-up to #1719 (which fixed the stash checkpoint storage growth). While independently root-causing the same 18GB `checkpoints.jsonl` / 20GB+ daemon RSS explosion, end-to-end verification with a real daemon surfaced two additional bugs that #1719 did not cover, plus a broader regression-test suite for the stash/pop explosion.

### 1. `git stash pop -q` / `apply -q` never restored attribution (src/daemon/ref_cursor.rs)

`enrich_stash` treated the first argv token after the stash subcommand as the stash target: `stash_args.get(1)`. For `git stash pop -q` that's `-q`, so stash-sha resolution failed and the attribution restore silently no-op'd — the stash entry was consumed by git but its attribution was dropped, and on pop the `stashes_v2` entry was leaked (never cleaned up). Coding agents typically pass `-q`, so this is exactly the flow from the original bug report.

Fixed with `stash_positional_arg`, which skips flags when locating the positional target (and the branch-name/target positions for `stash branch`). Repro before fix: `git stash push -q && git stash pop -q` → attribution gone; without `-q` → restored.

### 2. `merge_working_log_dirs` duplicated identical checkpoints (src/git/repo_storage.rs)

When a working log merges into an existing one on HEAD moves (checkout/switch), checkpoints were concatenated with plain `Vec::extend` — a log merged forward and back accumulated duplicates, the same compounding shape as the stash bug at lower frequency. Now dedups by `(kind, diff, timestamp, author)` (`diff` is a content hash, so distinct checkpoints never collide).

### 3. Regression tests for the stash explosion (tests/integration/stash_attribution.rs)

TDD-written tests (each watched fail on the pre-#1719 code) covering the full bug surface, adapted to the compact `stashes_v2` snapshot design:

- `test_stash_pop_cycles_do_not_grow_checkpoints` — 5 push/pop cycles, live line count must not grow (was 32× pre-fix), no leaked stash entries
- `test_stash_push_clears_live_checkpoints_and_pop_restores`
- `test_stash_apply_repeated_is_idempotent` — incl. the conflict-tolerant re-restore after a failed second apply
- `test_stash_push_pathspec_keeps_unstashed_checkpoints_in_live_log` — live-log complement of the stash-side pathspec filter
- `test_stash_pop_after_head_move_stays_bounded` — shifted-pop path
- `test_legacy_stashes_dir_is_removed`
- `test_stash_pop_quiet_flag_restores_attribution` / `test_stash_apply_quiet_flag_restores_attribution` — the ref_cursor fix; fail without it
- `test_merge_working_log_dirs_dedups_identical_checkpoints` (unit) — the repo_storage fix; fails without it

## Verification

End-to-end with a real installed daemon (`task dev`): 10 files × 300 lines with AI checkpoints, then the reported repro — 20 manual `stash push -q` / `pop -q` cycles. `checkpoints.jsonl` byte-identical to baseline (7,931 bytes), daemon RSS flat (~3.5MB), zero leaked stash entries, legacy `stashes/` dir removed, and `git-ai blame` shows full AI attribution after commit (broken before the ref_cursor fix).

Full stash suite (114 tests), ref_cursor/checkout/switch suites, `task lint`, `task fmt` all green locally.

---

<a href="https://app.devin.ai/review/git-ai-project/git-ai/pull/1722" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/src/daemon/ref_cursor.rs
+++ b/src/daemon/ref_cursor.rs
@@ -822,9 +822,9 @@ impl RefCursor {
 
         if matches!(kind, "apply" | "pop" | "drop" | "branch") {
             let target = if kind == "branch" {
-                stash_args.get(2)
+                stash_positional_arg(stash_args, 1)
             } else {
-                stash_args.get(1)
+                stash_positional_arg(stash_args, 0)
             };
             cmd.stash_target_oid = self.resolve_stash_target_at_cursor(target)?;
         }
@@ -836,7 +836,7 @@ impl RefCursor {
                 cmd.ref_changes.push(entry_to_ref_change(&entry));
             }
         } else if matches!(kind, "pop" | "drop") {
-            self.consume_destructive_stash_operation(stash_args.get(1), cmd)?;
+            self.consume_destructive_stash_operation(stash_positional_arg(stash_args, 0), cmd)?;
         }
 
         if matches!(kind, "apply" | "pop" | "branch")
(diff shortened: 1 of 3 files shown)


Siddhant-K-code:
See: https://github.com/git-ai-project/git-ai/pull/1721

PR #1731, opened 2026-07-03
Don't upload metrics when `telemetry_oss` flag is turned off even if `api_key` is set

a-churchill:
Caveat: this only happens if `api_key` is set.

In that case, it's still unexpected that any telemetry data is being sent to an external server when `telemetry_oss` is turned off. In particular it directly violates this: https://github.com/git-ai-project/git-ai/blob/da0aaddd25ae3bb5f5be1e3513bce584b611ea98/data-privacy.md#L7

This won't affect most people since `api_key` is not set by default.

---

<a href="https://app.devin.ai/review/git-ai-project/git-ai/pull/1731" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/src/api/metrics.rs
+++ b/src/api/metrics.rs
@@ -16,11 +16,21 @@ static LAST_METRICS_UPLOAD_STARTED_AT: OnceLock<Mutex<Option<Instant>>> = OnceLo
 
 /// Returns whether metrics are allowed to upload for the current API context.
 ///
-/// The server always requires authentication (API key or OAuth login).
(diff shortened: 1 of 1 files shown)


svarlamov:
`telemetry_oss` _does_ block telemetry. This conflates two unrelated concepts -- when the `api_key` is set, data is sent to the git ai cloud based on the user's explicit configuration. In that case, `telemetry_oss` is still active and does not report usage/error telemetry

PR #1821, opened 2026-07-11
fix(runtime): bound daemon worker arenas

svarlamov:
## Bug
The daemon used Tokio's CPU-count worker default (32 here), and every checkpoint created another CPU-sized helper runtime on a scoped OS thread. Glibc retained a 64 MiB arena per worker: the old installed daemon grew virtual memory by 2.27 GiB over 100 commits and started with 40 threads.

Constrained CI also exposed a pre-existing trace2 routing race. A parent-repository commit can discover an embedded repository and emit a second `def_repo` on the root SID, followed by child-SID frames for that nested repository. Both ingress metadata and `TraceNormalizer` previously let those later frames overwrite the command's parent worktree. Under scheduler pressure, the parent commit could therefore be analyzed as a subrepo command and finish without its authorship note.

## Fix
Use a fixed four-worker/16-blocker daemon runtime and one eagerly initialized shared helper runtime with two workers/four blockers. Nested synchronous bridges reuse that helper instead of constructing runtimes, eliminating repeated allocator-arena growth.

Treat the first root-SID `def_repo` as authoritative in both trace ingress and normalization. Later root-SID discoveries and child-SID `def_repo` frames cannot retarget the command. The ingress guard is a bounded per-root hash-set entry that is removed at terminal cleanup; it adds no Git calls, repository reads, or non-constant-time work to the critical path.

This root-cause fix is intentionally the base of the memory-hardening stack, so every later TestRepo daemon and CI layer runs on bounded pools and correct repository routing. The base also gives bounded-runtime TestRepo work the existing cross-platform 120-second total/45-second idle sync budget and seeds the oversized-stash fixture below a realistic 64 KiB test limit, preventing the harness from racing its own 64-byte truncation threshold.

## TDD coverage
Runtime unit tests assert worker/blocker ceilings and shared helper reuse. The TestRepo plateau regression performs 12 fully attributed warmup commits, records a 20-commit bounded-initialization window, then requires a second 20-commit window to grow `VmSize` by less than 256 MiB. Three simultaneous stress runs recorded first-window growth of 67,668 KiB, 0 KiB, and 67,668 KiB and late-window growth of 0 KiB in all three. Measuring the late window distinguishes sustained per-checkpoint growth from bounded lazy pool initialization.

For routing, deterministic unit tests reproduce both the second root-SID `def_repo` and child-SID nested-repository frames at ingress and normalization. They failed before the change and now require the terminal command to retain the parent worktree and family. The family assertion canonicalizes the expected `.git` path so macOS's `/var` to `/private/var` alias cannot create a false failure while the parent-vs-nested check remains exact. The existing nested-subrepo TestRepo regression now asserts exact human and AI line attribution after every commit. The unfixed root failed the real test in repeated four-process/four-core pressure rounds by rounds 3 and 5; the fix completed 48/48 concurrent reproductions.

## Verification
After the final restack and review follow-ups, `task fmt`, `task lint`, and `task build` pass. The complete `task test` matrix passes: 2,213 library tests, 58 daemon tests, 3,282 integration tests (85 expected ignores), 20 notes-sync tests, and four doctests (two expected ignores). A separate four-core/four-thread integration run passed all 3,282 executed tests in 477.22 seconds, including the nested-subrepo regression, saturated IPC backpressure, contended telemetry delivery, missing working-log blob recovery, Graphite, rebase, subdirectory, overload, memory, and exact-attribution workflows.

With exact final tip `3a467a77b8` installed through `task dev`, a fresh 300-commit checkpoint workload warmed from 1,115,328 KiB to 1,384,980 KiB `VmSize` by commit 50, then changed by 0 KiB through commit 300. RSS/HWM ended at 45,316 KiB with 17 threads. Blame reported 301 lines with exactly 300 `mock_ai` owners and one initial Codex line; all 301 commits had authorship notes.

On the same detached installed process, a 64-file fanout repository exercised the warm bounded pools. The three-pass workload issued 192 checkpoints while holding the same daemon PID; `VmSize` stayed within a 40 KiB band and ended at 1,384,940 KiB. RSS/HWM ended at 47,668 KiB with 15 threads. Exact blame across all files reported 192 `mock_ai` lines, 64 initial Codex lines, four authorship notes, and no other owners.

### DIFF
--- a/src/commands/daemon.rs
+++ b/src/commands/daemon.rs
@@ -16,6 +16,12 @@ use std::time::{Duration, Instant};
 #[cfg(windows)]
 use std::{ffi::OsStr, path::Path};
 
+// CPU-count defaults and the 512-thread blocking default reserve large per-thread
+// allocator arenas in this long-lived process. Daemon work already has narrower
+// queues and semaphores, so fixed pools preserve concurrency without arena growth.
+const DAEMON_RUNTIME_WORKER_THREADS: usize = 4;
+const DAEMON_RUNTIME_MAX_BLOCKING_THREADS: usize = 16;
+
 pub fn handle_daemon(args: &[String]) {
     if args.is_empty() || is_help(args[0].as_str()) {
         print_help();
@@ -186,10 +192,8 @@ fn handle_run(args: &[String]) -> Result<(), String> {
             e
         )
     })?;
-    let runtime = tokio::runtime::Builder::new_multi_thread()
-        .enable_all()
-        .build()
-        .map_err(|e| e.to_string())?;
+    let runtime = build_daemon_runtime()?;
+    crate::tokio_runtime::initialize();
     let exit_action = runtime
         .block_on(async move { crate::daemon::run_daemon(config).await })
         .map_err(|e| e.to_string())?;
@@ -219,6 +223,15 @@ fn handle_run(args: &[String]) -> Result<(), String> {
     }
 
     Ok(())
+}
+
+fn build_daemon_runtime() -> Result<tokio::runtime::Runtime, String> {
+    tokio::runtime::Builder::new_multi_thread()
+        .worker_threads(DAEMON_RUNTIME_WORKER_THREADS)
+        .max_blocking_threads(DAEMON_RUNTIME_MAX_BLOCKING_THREADS)
+        .enable_all()
+        .build()
+        .map_err(|e| e.to_string())
 }
 
 pub(crate) fn ensure_daemon_running(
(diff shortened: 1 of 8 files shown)


svarlamov:
* **#1821** <a href="https://app.graphite.com/github/pr/git-ai-project/git-ai/1821?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#1780](https://github.com/git-ai-project/git-ai/pull/1780) <a href="https://app.graphite.com/github/pr/git-ai-project/git-ai/1780?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#1822](https://github.com/git-ai-project/git-ai/pull/1822) <a href="https://app.graphite.com/github/pr/git-ai-project/git-ai/1822?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/git-ai-project/git-ai/1821?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1904, opened 2026-07-18
fix: exclude out-of-hook human edits from new-file AI line count

### LINKED ISSUE #1627: [Bug]:  New file  The AI line count is incorrect
### What broke?

Step 1: AI creates a new file without committing. git ai status reports the correct line count..
Step 2: Human edits some lines in the middle of the file without committing. git ai status still reports correctly..
Step 3: Commit the code, then run git ai stats --json. The AI line count is incorrect - it equals: AI new-file additions + human edit lines.. Running on Windows with OpenCode. Human edits file with vscode (Version:
1.85.1 without gi-ai vscode hook).

Version: 1.85.1 (user setup).
Commit: 0ee08df0cf4527e40edc9aa28f4b5bd38bbff2b2. Date:2023-12-13T09:49:37.021Z Electron: 25.9.7.
ElectronBuildId:25551756 Chromium:114.0.5735.289. Node.js: 18.15.0
V8:11.4.183.29-electron.0 OS: Windows_NT x64 10.0.19044
(Carl)

### Steps to reproduce

up

### Expected vs actual behavior

Humans edit a new file without Git AI hook tools, yet the operation can still be counted correctly.

### Diagnostics (`git-ai debug`)

```shell
no errors
```

### Extra context (optional)

```shell

```

mvanhorn:
## Summary

`git ai stats --json` over-counted AI lines for a brand-new file when a human edited some of its lines outside the git-ai hook (a plain editor with no extension). After commit, those human edits were absorbed into the AI total, because new-file line attribution trusted the AI attestation's line ranges by line number and never checked that the committed content still matched what the AI actually wrote.

## Why this matters

Two users reported the same class of bug on #1627: FisherChen saw the AI count equal "AI new-file additions + human edit lines", and gokulsignals independently reported broken commit-time attribution for AI-generated code. For a tool whose whole purpose is attributing authorship between human and AI, silently crediting human edits to the AI undermines the core statistic. The pre-commit `git ai status` path already reported the split correctly; only the post-commit reconciliation was wrong, so this leaves that path untouched.

## Testing

The fix makes new-file attribution content-aware: at reconciliation time, committed lines whose content diverges from the AI-attested content are demoted to human, so out-of-hook edits are excluded from the AI count. The same demotion is applied on the normal commit path, the `git commit --amend` path, and the gap-filling path, so a human insertion that happens to duplicate an adjacent AI line is not double-counted as both AI and human.

- `cargo build` passes and `cargo fmt --check` is clean
- The attribution test suite passes (159 tests), with new regression cases for the out-of-hook human edit, the amend variant, and the duplicate-line insertion

Closes #1627

AI was used for assistance.

---

<a href="https://app.devin.ai/review/git-ai-project/git-ai/pull/1904" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/src/authorship/post_commit.rs
+++ b/src/authorship/post_commit.rs
@@ -754,14 +754,44 @@ pub(crate) fn post_commit_amend_with_recovery_timestamps_detailed(
         "initial".to_string()
     };
 
-    let (mut authorship_log, initial_attributions, initial_file_contents) = working_va
-        .to_authorship_log_and_initial_working_log(
+    let (mut authorship_log, initial_attributions, initial_file_contents, out_of_hook_human_lines) =
+        working_va.to_authorship_log_and_initial_working_log_with_precomputed_diff(
             repo,
             &parent_sha,
             amended_commit,
             Some(&pathspecs),
             Some(&final_state_snapshot),
+            AuthorshipLogDiffContext::default(),
         )?;
+
+    // Reconciliation can prove that a line in a newly-created file diverged
+    // from the AI checkpoint before amend. Record those lines as human so
+    // later attribution recovery cannot infer AI ownership from neighboring
+    // AI lines.
+    if !out_of_hook_human_lines.is_empty() {
+        let human_hash = crate::authorship::authorship_log_serialization::generate_human_short_hash(
+            &human_author,
+        );
+        authorship_log
+            .metadata
+            .humans
+            .entry(human_hash.clone())
+            .or_insert_with(|| crate::authorship::authorship_log::HumanRecord {
+                author: human_author.clone(),
+            });
+        for (file_path, lines) in out_of_hook_human_lines {
+            if lines.is_empty() {
+                continue;
+            }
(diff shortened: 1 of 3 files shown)


svarlamov:
Hey @mvanhorn! Appreciate the PR, but unfortunately I'll have to close. The behavior is intentional, to cover the cases where AI changes at the margins don't match up because of the algorithmic difference between iterative diffs/attribution that accumulates edit-by-edit vs git's final diff at commit time.

The root cause fix (which I would be happy to accept PRs for) would be adding `known_human` hook support for the IDE(s)/editor(s) in question. Thanks and happy to chat more on Discord!

mvanhorn:
Totally fair - thanks for explaining the root cause (iterative edit-by-edit attribution accumulating vs git's final commit-time diff). The known_human hook support for the editor(s) in question sounds like the right fix; I may take a crack at that and will follow up on Discord if I do. Appreciate the thoughtful close.

PR #1920, opened 2026-07-21
fix: preserve attribution across autocrlf commits

svarlamov:
## Summary

- treat CRLF and LF as equivalent throughout post-commit virtual attribution and carryover comparisons
- reuse one shared, allocation-free content equality helper while normalizing whole snapshots only when line-based diffing requires it
- add a `TestRepo` end-to-end regression covering `core.autocrlf=true`, committed blame, and `git-ai stats`

## Root cause

The immutable commit snapshot/carryover path introduced in v1.5.9 compared raw snapshot bytes. With `core.autocrlf=true`, the working snapshot contains CRLF while the committed blob contains LF, so identical lines were interpreted as unstaged or conflicting carryover and their AI attribution was lost. The pre-commit status path already handled EOL normalization, which explains the status/stats mismatch.

This fix changes only in-memory comparisons and adds no Git calls, object lookups, file reads, or ingestion-path work. Equality checks use constant auxiliary memory; line-based diff paths allocate at most one normalized string per CRLF-bearing snapshot and otherwise borrow the original content.

## Validation

- `task fmt`
- `task lint`
- `task build`
- `task test`
- focused CRLF, mixed-EOL carryover, and allocation fast-path tests

The exact Ubuntu reproduction fails before this change and reports all three additions as AI-authored after it.

### DIFF
--- a/src/authorship/imara_diff_utils.rs
+++ b/src/authorship/imara_diff_utils.rs
@@ -258,10 +258,41 @@ pub fn compute_line_changes<'a>(old: &'a str, new: &'a str) -> Vec<LineChange<'a
 /// index alignment between normalized diff hunks and original line arrays in
 /// `compute_line_changes`.
 pub(crate) fn normalize_line_endings(s: &str) -> std::borrow::Cow<'_, str> {
-    if !s.contains('\r') {
+    if !s.contains("\r\n") {
         return std::borrow::Cow::Borrowed(s);
     }
     std::borrow::Cow::Owned(s.replace("\r\n", "\n"))
+}
+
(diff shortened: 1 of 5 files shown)


svarlamov:
* **#1920** <a href="https://app.graphite.com/github/pr/git-ai-project/git-ai/1920?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/git-ai-project/git-ai/1920?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1887, opened 2026-07-16
Use `notes_api` instead of `refs` when handling amends

a-churchill:
Problem: path for updating notes on commit amend was not using the `notes_api` to access commit authorship data, so when a custom HTTP notes backend was enabled, the notes on amend would be incorrect. 

<details>

<summary>Here's an (anonymized) example from real-life usage:</summary>

**Valid note (pre-amend, never pushed to GitHub)**

Commit: `7c1f20476f91a172d2bccdbbb7fd46fb33d54681`
Note written: 2026-07-14 03:31:30

```
path/to/file.kt
  s_46dcc697a1f9ae::t_e01adbd3e1e2bb 18,48-49
---
{
  "schema_version": "authorship/3.0.0",
  "git_ai_version": "1.6.10",
  "base_commit_sha": "7c1f20476f91a172d2bccdbbb7fd46fb33d54681",
  "prompts": {},
  "sessions": {
    "s_46dcc697a1f9ae": {
      "agent_id": {
        "tool": "claude",
        "id": "fa83d539-eadf-46a7-bc66-4f22e5d8a6aa",
        "model": "unknown"
      },
      "human_author": "author <author@company.com>"
    }
  }
}
```

**Invalid / malformed note (post-amend, the actual PR commit)**

Commit: `87f74996540ad5e6fe6d65f96d3d6542924c51a3`
Author date: 2026-07-14 03:31:26 · Commit date: 2026-07-14 03:32:10 · Note written: 2026-07-14 03:32:18

```
path/to/file.kt
  s_46dcc697a1f9ae::t_e01adbd3e1e2bb 18,48-49
---
{
  "schema_version": "authorship/3.0.0",
  "git_ai_version": "1.6.10",
  "base_commit_sha": "87f74996540ad5e6fe6d65f96d3d6542924c51a3",
  "prompts": {}
}
```

Note: same file, same attestation hash, same line ranges (18, 48-49) as the valid note — but the
`sessions` key is gone entirely.
</details>

The regression test in this PR reproduces this case exactly.

Solution: use `notes_api::read_authorship_v3` instead of `refs::get_reference_as_authorship_log_v3` in the amend path.

---

<a href="https://app.devin.ai/review/git-ai-project/git-ai/pull/1887" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open in Devin Review
  </picture>
</a>

PR #1975, opened 2026-07-24
Fetch missing source notes from the HTTP backend before the git refs fetch

scouredimage:
## Problem

When a rewrite (rebase/restack/cherry-pick) needs source notes that are not in the local notes-db cache, `fetch_missing_notes_for_commits` has exactly one recovery path: a git refs fetch of `refs/notes/ai` from each remote. On the **HTTP notes backend** that is the wrong tool twice over:

1. the server — not `refs/notes/ai` — is the source of truth, so the refs fetch often has nothing to offer even when the backend holds every note;
2. the fetch needs working SSH auth in the daemon's environment, which is not a given (the daemon's child `git fetch` may not see the user's SSH agent).

Observed at scale in a large deployment: dozens of restacked PRs permanently lost AI attribution because translation aborted with

```
failed to fetch authorship notes for source commits [...]: Git CLI (... fetch ... +refs/notes/ai:refs/notes/ai-remote/origin) failed with exit code 128: git@github.com: Permission denied (publickey).
```

…while the HTTP backend held every single source note. #1887/#1888 made the *presence check* backend-aware (cached sources skip the network), but sources missing from the local cache still funnel into the SSH-only fetch.

## Fix

- New `notes_api::warm_cache_for_commits(shas)`: batch-fetches specific SHAs from the HTTP backend into notes-db as already-synced rows. Shares the fetch/cache core (`fetch_and_cache_notes`) and the cached-SHA filter with `warm_cache_for_remote`, whose behavior and logging are unchanged.
- `fetch_missing_notes_for_commits`: when the HTTP backend is enabled, pull the missing sources from it first; if that covers everything, done — no refs fetch, no SSH. Otherwise fall through to the existing refs fetch and the existing still-missing error contract (unchanged for the git-notes backend).

Constant git cost: zero new git spawns — the backend fetch is one HTTP call per 100 SHAs and a notes-db write.

## Tests

- `fetch_missing_notes_pulls_sources_from_http_backend_when_refs_fetch_fails` — repo with a remote whose fetch always fails (stand-in for broken SSH auth), note available only on a mock backend server: previously returned the stranding error, now succeeds and caches the note as synced. **Red without the fix (verified by disabling the new path), green with it.**
- `fetch_missing_notes_still_errors_when_backend_lacks_note` — when neither the backend nor any remote has the note, the loud failure with the missing SHAs is preserved.
- Existing `warm_cache_for_remote_*` and notes_api cache tests pass unchanged.

Related: #1961 replays rebases that completed while no daemon was observing; its replay funnels through this same fetch, so the two fixes compose — #1961 finds the missed rewrite, this PR lets its note translation succeed without SSH.

### DIFF
--- a/src/git/notes_api.rs
+++ b/src/git/notes_api.rs
@@ -381,7 +381,6 @@ pub fn materialize_notes_for_display(repo: &Repository, limit: usize) -> Result<
 /// This function is a best-effort operation: errors are logged but not propagated
 /// (callers should treat failure as a cache miss, not a hard error).
 pub fn warm_cache_for_remote(repo: &Repository, remote: &str) -> Result<(), GitAiError> {
-    use crate::api::client::{ApiClient, ApiContext};
     use crate::git::repository::exec_git;
 
     // 1. Walk recent history. Prefer the remote's default branch; fall back to HEAD.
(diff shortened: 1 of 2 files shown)


scouredimage:
One clarification that strengthens the motivation here: on an HTTP-backend deployment the git refs fetch in `fetch_missing_notes_for_commits` can never actually find notes — `push_authorship_notes` early-returns for the HTTP backend, so `refs/notes/ai` on the remotes is empty by construction. The current behavior is therefore strictly lose-lose: with working SSH auth the fetch "succeeds", finds nothing, and the shift silently proceeds without the missing sources; with broken SSH auth it errors and aborts the translation outright. (The `backend=git_notes` field in `fetch_authorship_notes`'s log line is hardcoded, which makes this path easy to misread as backend-dispatched — it runs for HTTP-backend deployments too.) This PR makes the rewrite path consult the store that actually holds the notes.

PR #1902, opened 2026-07-18
Fix high memory usage on large rebases

a-churchill:
When the daemon classifies a completed git rebase as a rewrite, it migrates authorship notes from the old commits to the rebased ones. When there's a big diff, it has to be read into memory and processed all at once. That caused unbounded memory usage and blocked all Git operations in degenerate cases (common in a large monorepo).

This change streams the diff instead of 

<details>
<summary>Here's a repro for the initial problem that's fixed on this branch (not sure the best way to include it in CI):</summary>

```sh
#!/usr/bin/env bash
#
# Reproduces the git-ai daemon memory blowup during rewrite (rebase) attribution.
#
# Root cause (see README.md in this directory):
#   shift_authorship_notes_with_existing_mode (src/authorship/rewrite.rs) queues one
#   (old_commit, new_commit) full-root-tree diff pair per rebased commit that has an
#   authorship note, then compute_diff_tree_stdin runs ONE
#   `git diff-tree --stdin -p -U0 -M -r` over all pairs and buffers the ENTIRE output
#   in memory (exec_git_stdin -> wait_with_output, src/git/repository.rs).
#   After a rebase across a large trunk delta, EVERY pair's tree diff contains the
#   whole trunk delta, so the buffered output is (trunk delta bytes) x (stack size).
#
# This script builds a fully synthetic repo, drives a real `git rebase` through an
# ISOLATED per-run daemon (same wiring as the integration-test harness: trace2 socket
# + isolated HOME; nothing is installed system-wide), samples daemon RSS while the
# daemon processes the rewrite, and prints the peak alongside the raw size of the
# equivalent batched diff-tree output.
#
# Usage:
#   ./repro.sh                 # default scale (~0.5 GB daemon RSS peak, a few minutes)
#   SCALE=small ./repro.sh     # quick sanity run (~75 MB peak, ~2 min)
#   SCALE=large ./repro.sh     # ~2.9 GB daemon RSS peak
#   STACK_COMMITS=24 TRUNK_FILES=200 LINES_PER_FILE=2500 ./repro.sh   # custom
#
# Knobs (env vars):
#   STACK_COMMITS    number of feature-branch commits carrying authorship notes
#   TRUNK_FILES      number of NEW files added on trunk after the branch forked
#   LINES_PER_FILE   lines per trunk file (~73 bytes/line)
#   RENAMED_FILES    pre-fork files renamed+edited on trunk (engages -M rename detection)
[...]

Issue #1677, opened 2026-06-30
macOS: git-ai can consume >20 GB memory during git pull


peckfly:
### Summary

On macOS, the `git-ai` process may consume extremely high memory while pulling the latest code.

### Observed behavior

During a code pull/update flow, Activity Monitor showed the `git-ai` process using about **21.17 GB** of memory. This appears high enough to severely impact the machine.

### Reproduction

1. Switch to a relatively old branch. It does not have to be a very old branch.
2. Run a pull/update to get the latest code.
3. Observe `git-ai` memory usage on macOS.

### Expected behavior

`git-ai` should keep memory usage bounded during `git pull` / fetch / merge operations, even when processing many changed files or trace events.

### Notes

- Platform: macOS
- Screenshot evidence: Activity Monitor showed `git-ai` at approximately **21.17 GB** memory usage.
- This may be related to pull/fetch/merge handling, trace2 ingestion, rewrite tracking, or processing a larger diff/history range, but I have not confirmed the exact cause yet.

<img width="1214" height="158" alt="Image" src="https://github.com/user-attachments/assets/fdc8a753-8a6c-4fbb-ad4e-3d5ba5f6163a" />

@svarlamov

Siddhant-K-code:
Thanks for reporting this, we're aware of the issue. We've had another user report similar memory behavior and are actively investigating. We're currently gathering concrete metrics to validate the root cause & measure the fix before shipping it.

A couple of questions to help us narrow this down:

1. How many changed files / commits were involved in the pull? (rough estimate is fine)
2. Which tool do you use Cursor, Copilot, Claude Code, or another AI coding tool alongside git-ai?
3. Was this the first git pull after installing git-ai, or does it happen on subsequent pulls too?
4. What git-ai version are you on? (`git-ai --version`)

Appreciate the report and the screenshot, very helpful! ❤️

Siddhant-K-code:
Could you run these and share the output? It'll help us pinpoint the cause:

```bash
# 1. Version
git-ai --version

# 2. Internal file permissions & sizes
ls -la ~/.git-ai/internal/

# 3. AI tool transcript sizes
for dir in ~/.cursor ~/.codex ~/.claude ~/.windsurf ~/.amp ~/.continue ~/.gemini; do
  [ -d "$dir" ] && echo "$dir: $(du -sh "$dir" 2>/dev/null | cut -f1)"
done

# 4. Metrics DB size
[...]

PR #1978, opened 2026-07-25
Bound range-diff squash fabrication in rewrite mapping derivation

scouredimage:
## Problem

A **restack undo** — `git reset --keep <pre-rebase-tip>`, exactly what Graphite runs when a restack is undone or aborted — drives the daemon's non-fast-forward rewrite handling with `old_tip = rebased tip`, `new_tip = original tip`. The merge base of those two is the branch's **fork point**, so the old side of the range-diff contains **every trunk commit landed since the branch forked** — thousands on a busy monorepo.

`parse_range_diff_output` mapped **every** unmatched `<` commit onto a neighboring kept commit (the squash heuristic, applied without any bound). The daemon literally derives **604 mappings for a 4-commit branch move**. Each fabricated mapping whose source commit carries an authorship note (on a fleet with notes on effectively every commit: all of them) becomes a full-root-tree `-U0 -M` diff pair that **reverses the entire trunk delta**, and the parsed per-pair structures (`added_lines_by_file`: one `u32` per `+` line, plus hunks) accumulate until every pair is processed:

```
daemon RSS ≈ (trunk commits since fork) × (lines modified on trunk) × ~12 B
```

This fires **despite** the streamed-diff fix (#1902 / b0efde3fb) — the raw patch text is streamed, the parsed structures are not. Production incident: **~30 GB RSS** on a large-monorepo deployment, with machine-wide memory pressure making every git command crawl. Fully synthetic, isolated repro with RSS measurements included in this PR under `scripts/repro/daemon-restack-undo-mapping-bomb/` (150 fabricated mappings at SCALE=small; 604 at default, +206 MB RSS from a 28 MB baseline on a debug build).

## Fix

Two independent bounds:

1. **`parse_range_diff_output` drops fabricated squash mappings wholesale when the dropped-commit count dwarfs the matched pairs** (limit: `max(64, 8 × matched)`). A genuine squash folds a handful of commits into an adjacent kept one; thousands of unmatched commits mean the two ranges are not versions of the same branch, and mapping them is meaningless — a restack undo is not a squash of 3,000 trunk commits into a 4-commit stack. Matched (`=`/`!`) pairs always survive, so real note migration is unaffected — and on a restack undo the original tips already hold their notes, so nothing is lost.
2. **`derive_mappings_from_range_diff` skips derivation when either range side exceeds `GIT_AI_RANGE_DIFF_COMMIT_LIMIT` (default 1000) commits** — an outer wall for the `range-diff`/`diff-tree` spawn cost itself. One `git rev-list --count` per side (constant spawns).

Deliberately **not** included (noted as follow-up): chunking `compute_diff_trees_batch` so `DiffTreeResult`s are dropped as applied. With fabrication bounded, pair counts return to real stack sizes and per-pair diffs return to commit-sized deltas, so the accumulation ceases to be the dominant factor.

## Tests

- `parse_range_diff_keeps_proportionate_squash_mappings` — genuine squash fabrication preserved (existing squash tests also still pass).
- `parse_range_diff_drops_disproportionate_squash_mappings` — 70 dropped vs 2 matched: only the matched pairs survive. **Red without the fix.**
- `derive_mappings_restack_undo_yields_only_stack_mappings` — real repo, repro-shaped at small scale: 2-commit branch, 70-commit trunk advance, rebase, then mapping derivation in the undo direction returns exactly the 2 stack mappings with zero trunk leakage. **Red without the fix.**
- `derive_mappings_skips_when_range_exceeds_commit_limit` — the outer wall, env-overridable limit. **Red without the guard.**
- All 36 `authorship::rewrite` unit tests green.

Related: #1902 fixed the raw-text buffering on this same path; #1677 reported the same user-visible symptom from an earlier variant.

### DIFF
--- a/scripts/repro/daemon-restack-undo-mapping-bomb/README.md
+++ b/scripts/repro/daemon-restack-undo-mapping-bomb/README.md
@@ -1,0 +1,167 @@
+# Repro: daemon memory blowup on restack undo (range-diff mapping bomb)
+
+Minimal, fully synthetic reproduction of a production incident where the git-ai
+daemon reached ~30 GB RSS (and made every git command crawl via machine-wide
+memory pressure) while processing the rewrite-attribution side effect of a
+**restack undo** — moving a branch from its rebased tip back to the pre-rebase
+tip, which is exactly what Graphite runs (`git reset --keep <old-tip>`) when a
+restack is undone or aborted.
+
+This fires **despite** the streaming diff-tree fix ("Fix high memory usage on
+large rebases", `b0efde3fb`, shipped in v1.6.15). That fix stopped buffering
+the raw patch text; this bug is the *parsed* structures, multiplied by an
+unbounded mapping count.
+
+Everything here is isolated: the script creates a brand-new temp repo, a fake
+`$HOME`, and a **per-run daemon** wired via trace2 sockets (the same wiring the
+integration-test harness uses). Nothing is installed system-wide and nothing
+touches your real `~/.git-ai`.
+
+## Root cause
+
+A non-hard `git reset` to a non-ancestor (the daemon's Reset handling in
+`src/daemon.rs`, non-fast-forward arm) raises
+`RewriteEvent::NonFastForward { old_tip: rebased_tip, new_tip: orig_tip }`,
+which flows into `handle_non_fast_forward_rewrite_with_operation`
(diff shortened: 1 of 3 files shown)


svarlamov:
Thanks! Closing as superseded by @a-churchill PR #1983

PR #1985, opened 2026-07-26
Preserve large squash authorship mappings

svarlamov:
## Summary

- Preserve genuine leading squash-source authorship for generic `update-ref` / non-fast-forward rewrites up to the historical 64-commit safety boundary, including edited and empty metadata-only sources.
- Preserve arbitrarily large leading squash groups only when a trace2-observed rebase provides an immutable old-range boundary.
- Discard the leading prefix for ambiguous rebases and groups larger than 64 on generic ref moves, so divergent restack-undo history never reaches note reads or diff-tree work.
- Keep the historical unbounded behavior for `<` entries after the first confirmed match; those are established between-match squash sources.
- Bound downstream memory and platform argument usage by deserializing source notes lazily, deduplicating target-note parsing, and feeding tree/merge-parent batches to one Git process over stdin.
- Update `repro_1985.sh` to use this checkout's `target/debug/git-ai` by default and assert that the divergent prefix is discarded.

## Correctness model

A rebase may preserve every leading source only when its old range is immutable and scoped to the operation. The daemon derives that boundary without reading mutable repository state at processing time:

- full upstream OIDs are already immutable;
- `HEAD` / `@` parent-ancestry expressions are anchored to the trace2/reflog-captured original tip;
- standard named rebases use the exact cursor-bounded checkout OID when there is no explicit `--onto`;
- explicit `--fork-point`, `--root`, mutable old boundaries used with `--onto`, and other ambiguous forms fail closed to leading-prefix discard.

Generic ref moves do not have equivalent provenance. They retain the old <=64 behavior for real Graphite/update-ref squashes, while the 65th leading unmatched source discards the whole group. Once range-diff establishes a real pair, later unmatched sources keep mapping to the preceding destination as before.

All boundary capture is pure trace2/reflog-derived parsing. This adds no Git spawn, object lookup, ref lookup, or file read to the critical ingestion path. Rewrite processing remains constant-spawn; the large batches changed here use stdin instead of argv.

## Regression coverage

- Graphite-style `commit-tree` + `update-ref` leading squash
- edited leading squash source
- metadata-only empty leading source
- exact 64/65 generic boundary
- 65-source interactive rebase using an immutable `HEAD~N` boundary
- divergent named `rebase --onto` that must not copy old-upstream authorship
- 65-commit restack undo
- arbitrarily large post-match squash group
- 2,048-item tree and merge-parent stdin batches
- streaming merge across the chunk boundary

## Validation

- `task build`
- `task fmt`
- `task lint`
- `task test` (2,120 unit tests; 3,250 integration tests passed, 85 intentionally ignored; doctests passed)
- full `commit_tree_update_ref` test binary
- focused rebase/restack regression matrix after restacking onto current `main`
- `bash -n repro_1985.sh`
- `SCALE=small ./repro_1985.sh` (150 divergent commits -> 4 real mappings; flat daemon RSS)

### DIFF
--- a/repro_1985.sh
+++ b/repro_1985.sh
@@ -1,0 +1,508 @@
+#!/usr/bin/env bash
+#
+# Regression-checks the git-ai daemon behavior on a "restack undo" — moving a
+# branch ref from a rebased tip back to its pre-rebase tip via
+# `git reset --keep <old-tip>` (exactly what Graphite runs when undoing /
+# aborting a restack). NOTE: `reset --hard` does NOT reproduce this — the
+# daemon's ResetKind::Hard arm only deletes the working log; the non-fast-
+# forward rewrite path fires for the non-hard reset kinds (daemon.rs Reset
+# handling).
+#
+# Older builds created one mapping for every unmatched trunk commit. Current
+# builds must bound and discard that divergent group before note reads or
+# diff-tree work:
+#
+#   For a restack undo, the old range (merge-base..rebased-tip) contains every
+#   trunk commit since the branch forked — thousands on a busy monorepo. If
+#   those leading `<` entries are mapped, each noted trunk commit becomes a
+#   (trunk_commit, stack_commit) full-root-tree diff pair. Each pair reverses
+#   the whole trunk delta:
+#
+#     daemon RSS  ~=  (trunk commits since fork) x (lines modified on trunk) x ~12B
+#
+# This script builds a fully synthetic repo, drives a real rebase + reset
+# through an ISOLATED per-run daemon (same wiring as the integration-test
+# harness; nothing installed system-wide), samples daemon RSS during rewrite
+# processing, and verifies the mapping count stays close to STACK_COMMITS
+# rather than TRUNK_COMMITS.
+#
+# Usage:
+#   ./repro_1985.sh                 # default regression scale (a few minutes)
+#   SCALE=small ./repro_1985.sh     # quick sanity run
+#   SCALE=large ./repro_1985.sh     # high-load regression run
+#   TRUNK_COMMITS=600 BASE_FILES=80 LINES_PER_FILE=1000 ./repro_1985.sh   # custom
+#
+# Knobs (env vars):
+#   STACK_COMMITS    feature-branch commits with parseable AI authorship notes
+#   TRUNK_COMMITS    commits on main after the fork; EVERY one gets an
+#                    authorship note (models an org-wide notes fleet) and each
+#                    would become one bogus full-root-tree mapping without the
+#                    leading-prefix guard
+#   BASE_FILES       pre-fork files that the first trunk commit rewrites
+#   LINES_PER_FILE   lines per pre-fork file; every line is modified on trunk,
+#                    so per-pair '+' lines = BASE_FILES * LINES_PER_FILE
+#   PROCESS_TIMEOUT_SECS  how long to wait for the daemon to finish (default 1800)
+#   KEEP=1           keep the temp workdir + daemon logs
+#   GIT_AI_BIN       path to a git-ai binary (default: <repo>/target/debug/git-ai)
+#
+# Expected mappings            ~= STACK_COMMITS (trunk prefix discarded)
+# Expected per-pair '+' lines  ~= BASE_FILES * LINES_PER_FILE
+# Expected daemon RSS delta    ~= mappings * plus_lines * ~12B
+set -euo pipefail
+
+# ---------------------------------------------------------------------------
+# Knobs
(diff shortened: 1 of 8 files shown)


svarlamov:
* **#1985** <a href="https://app.graphite.com/github/pr/git-ai-project/git-ai/1985?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/git-ai-project/git-ai/1985?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

svarlamov:
Addressed all review findings through current head `01721075f`.

The final correctness policy distinguishes the ambiguous rewrite shapes instead of treating them uniformly:

- Rebases with an immutable old-range boundary preserve every leading squash source.
- Implicit/explicit fork-point rebases do not trust the observed checkout as an old boundary, but preserve a legitimate leading squash group only while it stays within the historical 64-commit bound.
- Explicit `rebase --onto` remains `DiscardLeading`, including short prefixes, because its old upstream may be unrelated divergent history.
- **Pull rebases never treat the observed checkout as a trusted old boundary.** A pull fetches first, so the checkout observed at `rebase_start` can be the newly fetched upstream while Git selects replay commits from the pre-fetch remote-tracking reflog. Without an immutable pre-fetch fork point, the range remains untrusted.
- If a supposedly trusted boundary fails to resolve, range-diff falls back to merge-base and downgrades to `DiscardLeading`; it does not abort note migration or restore unsafe prefix mappings.
- Generic non-rebase ref moves, including `update-ref` squashes, retain bounded leading-squash preservation.
- Unparseable source notes are rejected before any tree resolution or `diff-tree` work is queued.
- `repro_1985.sh` constructs scalable commits and notes with two bounded `fast-import` processes; no loop spawns Git per commit, note, file, object, or ref.

The pull fix is a constant-time check of the already-normalized command kind during deferred rewrite handling. It adds no trace2 ingestion-path I/O, Git spawn, object lookup, or ref check.

TDD coverage includes a real conflicted `git pull --rebase origin <branch>` after a force-push. The test creates 65 abandoned upstream commits with one batched `fast-import` invocation, proves Git replays only the local feature while range-diff exposes more than 64 leading `<` entries, and verifies abandoned upstream authorship is not copied into the rebased feature note. Before the fix it failed with a leaked `shared.txt` attestation; after the fix it passes. The existing successful, conflict, abort, alias, config, autostash, skip, worktree, and merge pull-rebase cases remain green.

Local verification on `01721075f`:

- `task fmt`, `task lint`, and `git diff --check`
- Complete `task test` green: 2,128 library tests, 3,342 integration tests, 22 notes-sync tests, and all runnable doc tests
- The workstation's real `~/.local/bin/droid` was excluded from `PATH` because one integration test intentionally requires no discoverable system `droid`

Fresh performance matrix:

| Scale | Divergent `<` sources | Retained mappings | Daemon RSS baseline → peak | Peak child Git RSS | Async rewrite time after reset |
|---|---:|---:|---:|---:|---:|
| small | 150 | 4 | 25 → 26 MB | 0 MB | 0s |
| default | 600 | 4 | 27 → 27 MB | 17 MB | 0s |
| large | 1,200 | 4 | 28 → 29 MB | 24 MB | 0s |

At large scale the guard discarded all 1,200 bogus leading sources, retained exactly four real mappings, and avoided an estimated ~31 GB of bogus mapped-diff structures while daemon RSS grew by only the 1 MB sampling granularity.

Fresh CI and Devin review are now running for `01721075f`; this comment will be updated with the final matrix and review-thread state.

svarlamov:
Addressed the `trusted_old_range_base` validation finding, with the final implementation in `fae2083b9`.

A rejected trusted boundary no longer aborts rewrite processing. If `range-diff` fails, the handler resolves/peels the candidate to distinguish an invalid boundary from an unrelated Git error. An invalid candidate falls back to the old/new merge-base and downgrades to `DiscardLeading`; valid matched mappings survive without letting the fallback restore the divergent-prefix mapping bomb. Valid rewrites take the original path with **zero additional Git spawns**; the extra resolution/retry work exists only on the invalid-boundary failure path. There is no trace2 ingestion-path work and no per-commit/object spawning.

Added a real-repository TDD regression that first reproduced the fatal bad-revision error, then verifies both fallback mapping and leading-source discard. `task fmt`, `task lint`, all 47 rewrite unit tests, the 65-source rebase test, divergent-`--onto` test, and 65-commit restack-undo test are green locally. The full local suite is also green (2,123 unit, 3,252 integration, notes-sync, and doc tests; expected ignores only).

PR #2065, opened 2026-08-01
Count KnownHuman lines in range stats

### LINKED ISSUE #1875: Bug: Range query stats inverts human_additions and unknown_additions values
# Bug: Range query stats inverts human_additions and unknown_additions values

## Description
When using `git ai stats <range>` to get authorship statistics for a commit range, the returned values for `human_additions` and `unknown_additions` are **inverted** compared to the sum of per-commit statistics.

### Test Case
Create two commits with distinct authorship patterns:
1. **Commit A** - AI-generated code (91 AI lines, 0 human, 0 unknown)
2. **Commit B** - Human-written documentation (6 human lines, 0 AI, 0 unknown)

### Commands and Results

**Per-commit queries:**
```bash
git ai stats <commit-A-sha> --json
```
Output:
```json
{
  "human_additions": 0,
  "unknown_additions": 0,
  "ai_additions": 91
}
```

```bash
git ai stats <commit-B-sha> --json
```
Output:
```json
{
  "human_additions": 6,
  "unknown_additions": 0,
  "ai_additions": 0
}
```

**Expected sum from per-commits:**
```
human_additions: 6
unknown_additions: 0
ai_additions: 91
```

**Range query:**
```bash
git ai stats <commit-A-sha>..<commit-B-sha> --json
```
Output:
```json
{
  "range_stats": {
    "human_additions": 0,
    "unknown_additions": 6,
    "ai_additions": 91
  }
}
```

## Expected Behavior
Range query should report aggregated stats that match the sum of per-commit queries:
```json
{
  "human_additions": 6,
  "unknown_additions": 0,
  "ai_additions": 91
}
```

## Actual Behavior
Range query reports inverted values for human vs unknown attribution:
```json
{
  "human_additions": 0,
  "unknown_additions": 6,
  "ai_additions": 91
}
```

## Versions Affected
- v1.5.1 ✅ (exhibits issue)
- v1.6.11 ✅ (exhibits issue)

Both versions show the same discrepancy, indicating this is a regression or longstanding bug in range aggregation logic.

## Impact
- Range-based authorship metrics are unreliable
- Dashboard queries using range stats may report incorrect human vs unknown line counts
- Per-commit queries work correctly; only range aggregation is affected
- Affects any downstream systems relying on `git ai stats <range>` for PR/branch measurements

## Analysis
The `ai_additions` and `ai_accepted` fields aggregate correctly across the range, suggesting the bug is specific to the `human_additions` / `unknown_additions` handling in the range aggregation logic.

## Workaround
Use per-commit evaluation instead of range queries:
- Walk commits in the range individually
- Run `git ai stats <commit-sha>` for each commit
- Sum the individual results
- Filter out zero-diff commits (e.g., merge commits)

This approach works correctly and is what the commit-worker already does.

## Suggested Fix
Ensure range query aggregation uses consistent logic with per-commit aggregation, or at minimum ensure that summing per-commit results matches the range query output.

nosey-dewdrop:
Fixes #1875

## The report

`git ai stats <A>..<B> --json` disagrees with the sum of the per-commit queries: a commit whose 6 added lines are KnownHuman reports `human_additions: 6` on its own, but inside a range those same 6 lines come back as `unknown_additions: 6, human_additions: 0`. The issue describes it as an inversion; it is narrower than that.

## Root cause

`calculate_range_stats_direct` (`src/authorship/range_authorship.rs`) passes a hard-coded `0` as `known_human_accepted`:

```rust
let stats = stats_from_authorship_log(
    Some(&authorship_log),
    git_diff_added_lines,
    git_diff_deleted_lines,
    diff_ai_stats.total_ai_accepted,
    0,                                   // <-- known_human_accepted
    &diff_ai_stats.per_tool_model,
);
```

`stats_from_authorship_log` then computes `unknown = added - ai_accepted - known_human_accepted`, so with `known_human_accepted = 0` every `h_`-attested line in the range lands in `unknown_additions` and `human_additions` can never be anything but zero. The per-commit path (`stats_for_commit_stats_from_hunks`) passes the real count from `accepted_lines_from_attestations`, which is why the two paths disagree.

## The fix

`diff_ai_accepted_stats` already blames every added line in the range with `use_prompt_hashes_as_names`, and blame reports KnownHuman lines under their `h_` hash (`src/commands/blame.rs`, the `hash.starts_with("h_")` branch). Those lines are simply never counted, because the existing loop only counts hashes present in `prompt_records`.

So the count is taken from the same blame pass, in the same loop, and surfaced as `DiffAiAcceptedStats::total_known_human_accepted`. `calculate_range_stats_direct` passes it through instead of the literal `0`.

No extra git spawns, no extra object lookups, no per-item work — the data was already in hand.

## Tests

TDD, red first. New regression test in `tests/integration/range_authorship_unit.rs` asserts the invariant the issue is really about: **range stats must equal the sum of the per-commit stats.** It builds a base commit, an AI commit (`mock_ai`) and a KnownHuman commit (`mock_known_human`), asserts line-level attribution after every commit, then compares the range against the per-commit ground truth.

Before the fix:

```
test range_authorship_unit::test_range_stats_match_sum_of_per_commit_stats_for_known_human_lines ... FAILED
assertion `left == right` failed: range dropped the KnownHuman line that the per-commit query reports
  left: 0
 right: 1
```

Three existing assertions had pinned the old behaviour as expected, each carrying a comment that named the hard-coded zero as the reason:

- `range_authorship_unit.rs` — `test_range_authorship_mixed_commits`
- `range_authorship_unit.rs` — `test_range_authorship_mixed_lockfile_and_source`
- `stats.rs` — `test_stats_cli_empty_tree_range` (CLI level, `git ai stats <range> --json`)

Each now asserts the corrected numbers, and the comments are updated to say what the behaviour is rather than why it was wrong.

Local results (macOS):

```
cargo test range_authorship   → 29 passed; 0 failed
cargo test stats::            → 25 passed; 0 failed
cargo fmt -- --check          → clean
cargo clippy --all-targets    → no new findings in the touched files
```

The `*_in_worktree` variants do not run on this machine — its git is 2.39.5, and `git worktree add --orphan` needs git >= 2.42. They fail identically on unmodified `main` here, so they are left to CI.

---

<a href="https://app.devin.ai/review/git-ai-project/git-ai/pull/2065" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/src/authorship/diff_ai_accepted.rs
+++ b/src/authorship/diff_ai_accepted.rs
@@ -8,6 +8,9 @@ use crate::git::repository::Repository;
 #[derive(Debug, Default)]
 pub struct DiffAiAcceptedStats {
     pub total_ai_accepted: u32,
+    /// Lines carrying an `h_`-prefixed KnownHuman attestation. Counted from the
+    /// same blame pass as the AI lines, so this costs no extra git work.
+    pub total_known_human_accepted: u32,
     pub per_tool_model: BTreeMap<String, u32>,
     pub per_prompt: BTreeMap<String, u32>,
 }
@@ -64,14 +67,20 @@ pub fn diff_ai_accepted_stats(
         }
 
         for line in &lines {
-            if let Some(author_hash) = line_authors.get(line)
-                && prompt_records.contains_key(author_hash)
-            {
+            let Some(author_hash) = line_authors.get(line) else {
+                continue;
+            };
+            if prompt_records.contains_key(author_hash) {
                 stats.total_ai_accepted += 1;
                 *stats.per_prompt.entry(author_hash.clone()).or_insert(0) += 1;
                 if let Some(tool_model) = author_tool_map.get(author_hash) {
                     *stats.per_tool_model.entry(tool_model.clone()).or_insert(0) += 1;
                 }
+            } else if author_hash.starts_with("h_") {
+                // KnownHuman attestation: blame reports the `h_` hash as the author
+                // when `use_prompt_hashes_as_names` is set, and it is never a prompt
+                // record. Without this branch these lines fall through to "unknown".
+                stats.total_known_human_accepted += 1;
             }
         }
     }
@@ -173,6 +182,7 @@ mod tests {
     fn test_diff_ai_accepted_stats_default() {
         let stats = DiffAiAcceptedStats::default();
         assert_eq!(stats.total_ai_accepted, 0);
+        assert_eq!(stats.total_known_human_accepted, 0);
         assert_eq!(stats.per_tool_model.len(), 0);
         assert_eq!(stats.per_prompt.len(), 0);
     }
@@ -181,8 +191,7 @@ mod tests {
     fn test_diff_ai_accepted_stats_debug() {
         let stats = DiffAiAcceptedStats {
             total_ai_accepted: 10,
-            per_tool_model: BTreeMap::new(),
-            per_prompt: BTreeMap::new(),
+            ..Default::default()
         };
         let debug_str = format!("{:?}", stats);
         assert!(debug_str.contains("DiffAiAcceptedStats"));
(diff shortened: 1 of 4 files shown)


svarlamov:
Thank you for the contribution, but I have to close this as a dupe of #1880

Issue #2093, opened 2026-08-04
Multi-GB memory on gt move / restack-undo persists on v1.6.20: unbounded post-match mapping fabrication + range-diff child memory


scouredimage:
## Summary

Restack/move-heavy Graphite users still hit multi-GB memory blowups on **v1.6.20**. The daemon-heap bomb fixed by #1983 is confirmed fixed (our repro of that path now shows **+0 MB** daemon RSS delta on 1.6.20) — but a sibling path survives: for "branch moved off commits that sit **on top of** matched ones" rewrites (`gt move`, restack/absorb undo, trunk overwrite), rewrite processing drives a **daemon-spawned `git` child (range-diff/diff-tree path) to unbounded memory**, scaling with dropped-commit-count × diff-size.

Two production reports this week, both on 1.6.20, both heavy `gt` users: one hit ~15 GB during a `gt sync` benchmark, another "most of my memory". Both previously suffered the #1983 bomb, so from the user's perspective the leak "is not fixed".

## Where the bound is missing

`src/authorship/rewrite.rs` (`parse_range_diff_output`), current main:

- Pre-match drops are bounded: `pending_dropped` is capped by `MAX_PENDING_DROPPED_COMMITS` and overflow discards the queue. ✅ (#1983-era hardening)
- **Post-match drops are not**: once any `=`/`!` pair sets `previous_new_sha`, every subsequent `<` line pushes a fabricated mapping with no cap (the `if let Some(new_sha) = previous_new_sha` branch, ~line 1109). N dropped commits after one match ⇒ N mappings ⇒ N downstream diff processings.
- Independently, the spawned `git range-diff` child's own memory is unbounded and dominates: it scales with (old-range length × new-range length × per-commit diff size). The daemon streams the *output*, but the child buffers pairwise cost internally.

## Repro (isolated daemon, nothing touches the real install)

Attached `repro_move.sh` (a variant of the restack-undo repro): builds

- `feature` = 4 AI-noted commits `S1..S4` + `T1..T150` big commits on top (every commit rewrites 25×400-line files; all noted),
- `feature-prime` = rewritten siblings `S1'..S4'` (same trees, new SHAs),

then runs traced `git reset --keep feature-prime` on `feature` — the `gt move`/undo shape. `git range-diff base..feature base..feature-prime` yields `S* = S*'` matches followed by 150 `<` drops **after** the match.

Measured on the v1.6.20 release binary (macOS/arm64), toy scale:

[...]

PR #2100, opened 2026-08-05
Skip the refs/notes/ai fetch for missing source notes on the http backend

### LINKED ISSUE #2099: Push-time authorship-notes fetch blocks git push for minutes (killing the daemon makes pushes instant); pure-cost fetch on HTTP notes backend
## Summary

On v1.6.20, `git push` (and Graphite's `gt submit`/`gt push`, which wrap it) can stall for minutes while the daemon runs its **push-time authorship-notes reconciliation fetch** against the remote. Killing the daemon makes the same push complete instantly — because the traced git process stops queuing behind the daemon's synchronous child-git work.

Related but distinct from #2093: that issue is the *rewrite-time* grind (unbounded range-diff child on move/undo shapes). This one is the *push-time* notes fetch. Both share the same structural problem — the daemon blocks trace-ingestion/op-processing on long-running spawned git work, and every traced git command on the machine waits behind it — but the triggers and hot paths differ, so filing separately.

## Production evidence (enterprise deployment, HTTP notes backend, macOS/arm64, v1.6.20)

User report: "`gt submit` and `gt push` take forever. I noticed killing `git-ai` makes the process super fast."

Daemon event timeline from that user's session (timestamps UTC):

```
22:46:14  fetching authorship notes
22:47:34  git write op completed  op=reset
22:47:52  fetching authorship notes   (x2)
22:48:30  fetching authorship notes
22:49:37  git write op completed  op=push      <- slow push #1
...
23:00:44  fetching authorship notes
23:19:51  git write op completed  op=push      <- slow push #2, next op ~19min after the fetch
```

Every slow push is immediately preceded by `fetching authorship notes`. On a separate machine (same deployment) we measured these fetch windows at **1–3 minutes each** during rewrite handling — daemon log gaps of 66s/89s/109s/195s between a fetch pair and the next completed write op. On large monorepo remotes the fetch is a full SSH negotiation and can also fail outright after burning its window (`source-note fetch failed; shifting the notes that are available locally`).

Note: on this deployment the notes backend is HTTP (`notes_backend.kind = "http"`), so the git-fetch of `refs/notes/ai` can never find anything on the remote — the ref is never pushed under the HTTP backend. The fetch is pure cost on this configuration (this is the same backend-blindness previously addressed for missing source notes in #1975; the push-time path still issues the git fetch).

## Why killing the daemon "fixes" it

The daemon processes trace2 ingestion and op handling serially enough that a multi-minute child git-fetch stalls the pipeline; traced git processes block on the trace socket until it drains. Removing the daemon removes the socket, so git proceeds untraced at full speed. (Same mechanism family as the saturation behavior described in #2093.)

## Suggested fixes

1. On HTTP-backend deployments, skip the git-fetch of `refs/notes/ai` entirely at push time (it cannot succeed by construction) — query the HTTP backend instead, as #1975 did for the source-note path.
2. Time-box / background the reconciliation fetch so it never blocks op processing; a push should not wait on note fetching.
3. Generally: never let trace-socket ingestion back up behind spawned child git work (shared root cause with #2093).

Environment: git-ai 1.6.20 (release), macOS Darwin 25.5.0 arm64, git 2.55.0, enterprise config with HTTP notes backend + allow_repositories.

scouredimage:
Fixes #2099 (first suggested fix there); related to #2093 and a follow-up to #1975.

## Problem

On http-backend deployments, `fetch_missing_notes_for_commits` warms the local cache from the server (#1975) — but whenever the backend legitimately lacks a note (the common case: fresh commits that never had notes), it **falls through to the per-remote `refs/notes/ai` git fetch**. On the http backend that ref is never pushed, so the fetch cannot succeed by construction; it only costs a full negotiation per remote (measured at 1–3 minutes on large monorepos, sometimes failing outright on SSH auth) while daemon op-processing blocks behind it.

Production symptom (#2099): `git push` / `gt submit` stall for minutes; killing the daemon makes the same push instant. Daemon timelines show `fetching authorship notes` immediately preceding every slow push.

## Fix

When the http backend is enabled, its answer is final: warm the cache from the server, and if notes are still missing, report them missing (same loud error contract, SHAs named) **without** attempting the refs fetch. The `git_notes` backend path is unchanged.

## Test

`fetch_missing_notes_skips_refs_fetch_on_http_backend` (TDD — failed before the fix): the repo's origin is a real local remote that *does* carry the note in `refs/notes/ai`, while the mocked http backend lacks it. Pre-fix, the fall-through fetch created the tracking ref and returned Ok; post-fix no tracking ref is created and the error names the missing SHA. The #1975 regression test (backend has the note → cached) and the missing-everywhere error-contract test still pass.

Pre-existing on main, unrelated to this change: 5 lib test failures (`test_run_command_capture_with_timeout_reports_partial_output`, `check_for_update_available_returns_update_ready_when_cache_has_pending_update`, `test_run_logged_command_with_timeout_reports_partial_output`, `test_load_ai_touched_files_for_specific_commits`, `conflict_resolution_note_read_errors_are_not_silently_ignored`) fail identically on clean `origin/main` in this environment.

### DIFF
--- a/src/git/sync_authorship.rs
+++ b/src/git/sync_authorship.rs
@@ -156,6 +156,28 @@ pub fn fetch_missing_notes_for_commits(
                 tracing::debug!("HTTP backend fetch for missing source notes failed: {}", e);
             }
         }
+
+        // On the HTTP backend the server is the sole source of truth:
+        // refs/notes/ai is never pushed there, so the refs fetch below
+        // cannot produce these notes by construction — it only costs a
+        // full negotiation per remote (minutes on large repos) while
+        // daemon op-processing blocks behind it. A note the server does
+        // not have is missing, period.
+        let still_missing: Vec<&String> = {
+            let noted = noted_commits(repository, source_commits);
+            source_commits
+                .iter()
+                .filter(|sha| !noted.contains(sha.as_str()))
+                .collect()
+        };
+        if still_missing.is_empty() {
+            return Ok(());
+        }
+        return Err(GitAiError::Generic(format!(
+            "authorship notes for source commits {:?} are missing from the notes backend \
+             (refs fetch skipped: refs/notes/ai is not populated on the http backend)",
+            still_missing
+        )));
     }
 
     tracing::debug!(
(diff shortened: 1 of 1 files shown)


svarlamov:
@scouredimage thank you for this contribution. I'm going to close this patch, but I'm already working on a blanket fix for this class of problem.

Issue #2157, opened 2026-08-17
Control-socket failures (EINVAL on set_recv_timeout, EPIPE) stall the daemon runtime and block traced git commands


scouredimage:
## Summary

Daemon control-socket connections fail with OS-level errors — `Invalid argument (os error 22)` from `set_recv_timeout` on the accepted local socket, and `Broken pipe (os error 32)` — and the failure does not stay contained to the control plane. On affected machines the daemon's periodic work overruns badly (`telemetry flush exceeded its scheduling interval elapsed_ms=30386 interval_ms=3000`), and in the most recent production report the daemon **stopped draining trace2 input entirely, blocking every traced `git` command** until the user ran `git-ai bg shutdown --force` from another shell.

We have been tracking this class in fleet telemetry since late July and now have a user-visible hang on **v1.6.22** (current release), so filing with everything we know.

## Observed log signature (v1.6.22, macOS arm64, 2026-08-17)

```
ERROR daemon control connection failed ... error=IO error: Broken pipe (os error 32)
ERROR daemon control connection failed ... error=Generic error: failed setting daemon control receive timeout: Invalid argument (os error 22)
WARN telemetry flush exceeded its scheduling interval elapsed_ms=30386 interval_ms=3000
WARN telemetry flush exceeded its scheduling interval elapsed_ms=30068 interval_ms=3000
```

User-visible effect: ordinary git commands in a traced repo hang indefinitely. `git-ai bg shutdown --force` cleared the socket and unblocked them. The engineer's own investigation summarized it as: the control-socket handler throws an OS-level error it doesn't recover cleanly from, and the runtime it shares with the trace2-draining task stalls with it.

## Fleet telemetry timeline (enterprise deployment, ~1.2k installs, macOS)

- **Zero occurrences** of the EINVAL signature in daemon log uploads before **2026-07-30 00:00 UTC**.
- Eruption: ~282k errors across 298 installs on Jul 30, ~357k on Jul 31 (fleet then mostly v1.6.17) — ~640k in two days. The onset coincided with a backend-outage recovery period on our side (daemon upload storms), suggesting a load/backpressure trigger rather than a version regression.
- Tapered sharply on v1.6.20 (~5 errors/install/day vs ~1,150 on v1.6.17) but never reached zero.
- Still occurring on v1.6.21 and **v1.6.22** at low per-machine rates — the machine in the report above logged EINVAL + EPIPE uploads the same day as the hang.
[...]

PR #2193, opened 2026-08-21
Log periodic flush-skip reasons on state change, not every pass

svarlamov-git-ai:
## Problem

Post-merge stress testing of the #2157 stack surfaced one regression: the telemetry flush loop (3s cadence) logs its skip reasons at INFO unconditionally on every pass.

- `metrics: skipping pending upload, not authenticated` — every 3s for any not-logged-in daemon ≈ **29,000 lines/day** in `~/.git-ai/internal/daemon/logs/` (measured: 60 lines per 3-minute idle window; v1.6.23 idles at ~5 lines total).
- `notes: skipping flush, backend is not Http` — the default backend is `GitNotes`, so **every default-config user** hits this on every pass that flushes a telemetry batch, authenticated or not.
- `notes: skipping flush, notes_backend.backend_url is not configured` / `notes: skipping flush, not authenticated` — same cadence for the corresponding configs.

These lines were added for diagnosability (a note/metric that silently never uploads should be explainable from the daemon log), so deleting or blanket-demoting them would trade one problem for another.

## Fix

`TransitionLogGate`: the skip reason logs at INFO when it first applies or **changes**, logs at debug for unchanged repeats, and the gate re-arms when the gated stage proceeds — so a later regression back into a skipping state logs at INFO again. Diagnosability is preserved (the reason is always in the log once), the repetition is gone.

Verified with a release build: an unauthenticated idle daemon now writes 6 log lines per 3 minutes (was 65), matching the v1.6.23 baseline.

## Audit of other periodic log output

Dug through every daemon loop for the same fire-per-iteration pattern:

| Loop | Cadence | Verdict |
|---|---|---|
| Telemetry flush loop | 3s | **the 4 lines fixed here**; upload/batch lines only fire on actual uploads |
| Socket health check | 30s | INFO once at startup; in-loop logs are anomaly-gated (probe failure/restart), where repetition is deliberate loudness |
| Update check | 3600s | ~24 lines/day, negligible |
| Memory watchdog | poll interval | already state-transition gated (warn once on failure, info once on recovery) |
| Ingest worker / family actors / stream workers | per event | activity-proportional per-checkpoint/per-git-op lines, pre-date the #2157 stack (log volumes measured equal between v1.6.23 and main under identical load) |
| `await` progress | while awaiting | already rate-limited by `log_interval` |

## Testing (TDD)

- `daemon_flush_skip_reason_logs_once_not_per_pass` (integration): drives 3 serialized flush passes through `ControlRequest::FlushNotes` on an unauthenticated default-config daemon and asserts the skip line appears exactly once. Red on the previous code (`left: 3, right: 1`), green with the gate.
- `transition_log_gate_*` unit tests: first occurrence → INFO, repeat → debug, reason change → INFO, `clear` re-arms.
- Full suite: 5766 passed, 0 failed. `task lint` / `task fmt` clean.

### DIFF
--- a/src/daemon/telemetry_worker.rs
+++ b/src/daemon/telemetry_worker.rs
@@ -34,6 +34,53 @@ const MAX_DAEMON_LOG_BUFFER_EVENTS: usize = 5000;
 
 static METRICS_UPLOAD_AVAILABLE: AtomicBool = AtomicBool::new(false);
 static METRICS_METADATA_BACKFILL_STARTED: AtomicBool = AtomicBool::new(false);
+static METRICS_SKIP_LOG_GATE: TransitionLogGate = TransitionLogGate::new();
+static NOTES_SKIP_LOG_GATE: TransitionLogGate = TransitionLogGate::new();
+
+/// Gates a log line that a periodic loop would otherwise emit every pass.
+///
+/// The flush loop runs every 3 seconds, so an unconditional INFO skip line
+/// ("not authenticated", "backend is not Http", ...) writes tens of
+/// thousands of identical lines per day into the daemon log file. The skip
+/// reason still has to be diagnosable from the log, so `log_skip` emits INFO
+/// when the message differs from the previous pass (including the first
+/// occurrence and after `clear`), and demotes unchanged repeats to debug.
+/// Call `clear` when the gated stage proceeds, so a later regression to a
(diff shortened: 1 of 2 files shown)


svarlamov-git-ai:
Closing in favor of a simpler approach: demote the skip lines to debug.

PR #2206, opened 2026-08-24
Add token_usage parsing core ported from ccusage

svarlamov-git-ai:
## Summary

First PR of the TokenUsage stack: a new `src/token_usage/` library module that extracts per-entry token usage from Claude Code and Codex transcripts, ported from [ccusage](https://github.com/ccusage/ccusage) (MIT, attribution in the module headers) and adapted to git-ai's incremental line-oriented streaming. Pure library code — no daemon changes.

- **Claude** (`claude.rs`): usage-line substring prefilter before any JSON parsing (the ccusage perf trick), strict ccusage schema validation (null-field rejection, semver version, non-empty ids), cache-creation duration breakdown (5m/1h), advisor iterations, `costUSD` precedence, and the sidechain-aware replacement policy (applied later at the DB layer since entries stream in across runs).
- **Codex** (`codex.rs`): cumulative `total_token_usage` delta computation with persisted parser state, `last_token_usage` preference, cached<=input clamp, zero-total derivation to input+output, ccusage's field aliases + string-encoded numbers, content-derived entry keys over the event's full identity (ccusage's codex dedup key — so forked/subagent rollouts replaying parent events dedup while distinct turns never collide), fork detection via `session_meta`, and the rewritten-burst replay heuristic adapted to streaming (anchoring on every usage-carrying event, matching `detect_rewritten_burst`).
- **Cost** (`cost.rs`): per-entry micro-USD estimation via the existing models.dev pricing catalog, including ccusage's 2x input rate for 1h ephemeral cache writes; transcript `costUSD` wins when present (ccusage "auto" mode).
- 5-minute UTC bucket helper for the TokenUsage metric event introduced later in the stack.

Deviations from ccusage are documented in the module headers (headless codex format, service tiers/fast pricing, parent-prefix replay matching, line-shape strictness, timestamp leniency).

## Review notes

- The port was adversarially compared line-by-line against the ccusage Rust reference; the fixup commit addresses the divergences that review found (identity-based codex keys, zero-total derivation, burst anchoring on zero-delta repeats).
- Known v1 limitation (documented): codex fork replay always uses the rewritten-burst fallback rather than parent-prefix matching, since reading parent files is not possible in an incremental single-file pass.

## Testing

- 38 unit tests covering ccusage edge cases: streaming re-emits, sidechain replay + message-id fallback, replacement policy matrix, advisor entries, null-field rejection, codex cumulative dedup, burst scenarios (incl. zero-delta repeats), state round-trip mid-stream == single pass, aliased/string-encoded fields.


src/token_usage/claude.rs
```diff
@@ -0,0 +1,554 @@
...
+                field_end -= 1;
+            }
+        }
+        if line.get(field_end) == Some(&b'"') {
+            let mut field_start = field_end.saturating_sub(1);
+            while field_start > 0 && line[field_start] != b'"' {
+                field_start -= 1;
+            }
+            if line.get(field_start) == Some(&b'"')
+                && is_unsupported_nullable_field(&line[field_start + 1..field_end])
```
svarlamov-git-ai, line 294: **[review finding 7/15 — crash, medium] Reversed slice range panics on a line starting with `":null`, creating a permanent poison-pill retry loop.**

For a garbled/truncated line whose first bytes are `":null` (and which contains `"usage":{` so it passes `wants_line`): `null_index = 1`, `field_end = 0`, `line.get(0) == Some(&b'"')`, `field_start = 0`, and `&line[field_start + 1..field_end]` evaluates to `&line[1..0]` — an out-of-order slice range that panics in both debug and release ('slice index starts at 1 but ends at 0'). The panic happens inside the extraction pass before `commit_batch` advances the cursor, so the file re-panics on every stream notification and every 30-minute sweep forever, and all later usage in the file is never counted (a panic also bypasses `record_error`). Guard with `field_start + 1 <= field_end` (or restructure with `match_indices`/`str` ops).

src/token_usage/codex.rs
```diff
@@ -0,0 +1,761 @@
...
+    /// Run one usage-carrying event through the fork-replay filter, returning
+    /// the entries that count as the session's own usage. `event` is `None`
+    /// for events that produced no delta but still mark activity.
+    fn filter_replay(&mut self, ts_ms: i64, event: Option<PendingEvent>) -> Vec<UsageEntry> {
+        let within_burst =
+            |anchor_ts_ms: i64| (0..=REWRITTEN_BURST_PAUSE_MS).contains(&(ts_ms - anchor_ts_ms));
+        match std::mem::take(&mut self.state.replay) {
+            ReplayState::Done => event.map(|e| vec![make_entry(e)]).unwrap_or_default(),
+            ReplayState::AwaitingFirst => {
+                self.state.replay = ReplayState::AwaitingSecond {
```
svarlamov-git-ai, line 318: **[review finding 12/15 — correctness/undercount, low-medium] A forked/subagent session whose transcript ends after a single usage event never counts that event.**

`AwaitingFirst` parks the first usage event's delta in `AwaitingSecond{pending}`, and the only release path is a second usage-carrying event arriving >1s later. The pending delta is durably persisted via state_json, so a fork/subagent rollout that produces exactly one token_count event (one-turn subagent doing a single tool call is realistic — `thread_spawn` also sets AwaitingFirst) parks that usage forever; there is no EOF/finalization or staleness flush anywhere. ccusage's whole-file scan resolves the ambiguity at EOF and counts a lone leading event (a burst needs a pair), so this is a genuine divergence from the port target, not just heuristic parity. A staleness-based release on a later sweep (pending older than the burst window by wall-clock) would recover it.

src/token_usage/codex.rs
```diff
@@ -0,0 +1,761 @@
...
+            || line.contains("turn_context")
+            || line.contains("session_meta")
+    }
+
+    fn extract_line(&mut self, line: &str) -> Vec<UsageEntry> {
+        let Ok(raw) = serde_json::from_str::<RawLine>(line) else {
+            return Vec::new();
+        };
+        match raw.entry_type.as_deref() {
+            Some("session_meta") => {
```
svarlamov-git-ai, line 164: **[review finding 15/15 — attribution/reuse, low-medium] Second, divergent Codex model parser: session_meta models are never read and the model_id/modelId aliases are missing.**

This arm only checks `is_forked_session` — it never seeds `state.model` — and `model_from_parts` accepts `model`/`model_name`/`metadata.model` only. The pre-existing parser `extract_model_from_codex_jsonl_line` (src/streams/model_extraction.rs) reads `model`/`model_id`/`modelId` from BOTH session_meta and turn_context (plus a .codex/config.toml fallback), added for a documented real case (Codex CLI 0.142.4 emitting the model in session_meta / config only). For such rollouts, SessionEvent attribution resolves the true model while token usage silently prices under FALLBACK_MODEL "gpt-5" — and every future codex format change now needs fixing in two parsers. The alias set is faithful ccusage parity, but seeding `state.model` from session_meta via the shared extraction helper would be a strict superset that keeps the two in agreement.

src/token_usage/claude.rs
svarlamov-git-ai: Fixed — the byte scan now requires `field_start < field_end` before slicing, so a line starting with `":null` can no longer compute a reversed range. Regression tests cover the `":null...` prefixes and the extractor path. (The panic-loop consequence is also independently mitigated: PR 2209 adds per-file error backoff, so even a future panic in a pass would back off instead of re-running on every trigger.)

src/token_usage/codex.rs
svarlamov-git-ai: Fixed — `UsageExtractor` gained `has_pending()`/`flush(now_ms)`: at end of file the worker flushes the parked first event once the 1s burst window has passed in wall-clock time (no later event can join a burst at that point), so a single-turn forked/subagent rollout emits its usage in the same pass. If the event is still inside the window, a `pending_flush` marker persisted with the batch makes sweeps re-process the file even though its bytes haven't changed. Covered by extractor unit tests and a worker-level test.

src/token_usage/codex.rs
svarlamov-git-ai: Fixed — the extractor now reads the model from `session_meta` payloads and accepts the `model_id`/`modelId` aliases (priority: model → model_name → model_id → metadata.model), matching `streams::model_extraction` so usage prices under the same model SessionEvents resolve. Kept as a local parser rather than reusing `extract_model` because that helper does whole-file tail/head scans per call, while the extractor needs the *sticky* per-position model for multi-model sessions — but the accepted field set is now the same. Unit tests cover both aliases and session_meta.

src/token_usage/codex.rs
```diff
@@ -0,0 +1,848 @@
...
+            Some("event_msg") => self.handle_event_msg(&raw),
+            _ => Vec::new(),
+        }
+    }
+
+    fn state_json(&self) -> Option<String> {
+        serde_json::to_string(&self.state).ok()
+    }
+
+    fn restore_state(&mut self, json: &str) {
```
svarlamov-git-ai: **[Token-usage accuracy review — finding 2/15]** (stack-wide review vs upstream ccusage)

`restore_state` silently resets to defaults on any deserialization failure while the worker keeps the advanced byte cursor (token_usage_worker.rs:611-616), breaking the module's own documented invariant that parser state must stay coupled to the cursor - the next cumulative token_count event then computes its delta against `prev_totals = None`, booking the session's entire history to date as one fresh entry (new content-derived key, not deduplicable), i.e. a full double count.

**Failure scenario:** Any future rename/variant addition in `CodexState`/`ReplayState` (the internally-tagged `#[serde(tag = "kind")]` enum errors on unknown kinds) or a daemon downgrade reading newer state: a mid-file session at 500K cumulative tokens emits a bogus ~500K-token entry on its next turn. The safe recovery is to reset the cursor to 0 alongside the state (idempotent via entry-key dedup, exactly what the shrunken-file path already does). The existing test `corrupt_state_resets_to_defaults` pins the hazardous half only; no test covers restore-failure with a nonzero offset.

src/token_usage/codex.rs
```diff
@@ -0,0 +1,848 @@
...
+                    // belong to the parent's history.
+                    self.state.replay = ReplayState::SkippingBurst { last_ts_ms: ts_ms };
+                    Vec::new()
+                } else {
+                    // A real pause: the session recorded its own turns from
+                    // the start.
+                    pending.into_iter().chain(event).map(make_entry).collect()
+                }
+            }
+            ReplayState::SkippingBurst { last_ts_ms } => {
```
svarlamov-git-ai, line 339: **[Token-usage accuracy review — finding 4/15]** (stack-wide review vs upstream ccusage)

Zero-delta cumulative repeats EXTEND the rewritten-burst window in git-ai (`filter_replay(ts, None)` advances `SkippingBurst.last_ts_ms` on every usage-carrying event) but not in upstream ccusage, where `SkippingRewrittenBurst` only advances via `visit_filtered`, which all-zero deltas never reach (parser.rs:339-345 returns before `visit`) - so a chain of sub-second zero-delta heartbeats can bridge the window across the child's real first turn and permanently discard it (prev_totals has already advanced, so the tokens are never recovered).

**Failure scenario:** Fork replay: event A (delta) at t=0, zero-delta repeat at t=0.9s, child's real first turn at t=1.7s. Upstream counts the child turn (1.7s > 1s from A); git-ai skips it (1.7-0.9 = 0.8s <= 1s) - silent undercount. The existing test `zero_delta_repeats_still_anchor_the_rewritten_burst` pins only the anchor/detection case (which does match upstream); the extension divergence is untested. Either align with upstream (only non-zero-delta events extend the burst) or pin the deliberate deviation with a test.

src/token_usage/codex.rs
```diff
@@ -0,0 +1,848 @@
...
+
+impl UsageExtractor for CodexUsageExtractor {
+    fn wants_line(&self, line: &str) -> bool {
+        line.contains("token_count")
+            || line.contains("turn_context")
+            || line.contains("session_meta")
+    }
+
+    fn extract_line(&mut self, line: &str) -> Vec<UsageEntry> {
+        let Ok(raw) = serde_json::from_str::<RawLine>(line) else {
```
svarlamov-git-ai, line 160: **[Token-usage accuracy review — finding 6/15]** (stack-wide review vs upstream ccusage)

Zero hostile/malformed-input tests for the extractors, where upstream has many (jsonl.rs lenient-shape tests, loader.rs `loads_headless_usage_with_unexpected_noncritical_field_types`) precisely because Codex has changed field types before: git-ai's whole-line skip on any unexpected shape (strict `RawLine`/`RawInfo` vs upstream's `deserialize_optional_object_lossy`) means one field-type change zeroes ALL codex usage; on the claude side a third `usage.speed` variant or float-typed token counts fail the whole line's parse; and skipped token_count lines also desync `prev_totals`, folding the skipped usage into a later delta with the wrong timestamp/model or losing it outright when the next event carries last_token_usage.

**Failure scenario:** Codex ships `"metadata": "auto"` or an array-valued `last_token_usage` on token_count lines -> every line fails `RawLine` parse, all Codex token reporting silently drops to zero, no test fails before or after. Needed: fixtures with scalar/array payloads, wrong-typed noncritical fields, unknown speed values, float counts - asserting both what still counts and that `prev_totals`/replay state survive skipped lines.

src/token_usage/codex.rs
```diff
@@ -0,0 +1,848 @@
...
+            }
+        )
+    }
+
+    /// A forked session whose transcript ends while its first usage event is
+    /// still parked in `AwaitingSecond` would never release it (a single-turn
+    /// subagent rollout, for example). Once the burst window has passed in
+    /// wall-clock time, no later event can be within it, so the buffered
+    /// event is real usage and is released.
+    fn flush(&mut self, now_ms: i64) -> Vec<UsageEntry> {
```
svarlamov-git-ai, line 228: **[Token-usage accuracy review — finding 7/15]** (stack-wide review vs upstream ccusage)

The fork/replay machinery is untested across incremental pass boundaries - its normal production shape: (a) the park-in-pass-1/release-in-pass-2 chain (pending_flush persisted -> sweep re-enqueues unchanged file -> quiet-skip bypass via has_pending) is never exercised across two passes (the only test uses a past timestamp so flush fires in the same pass); (b) `ReplayState::AwaitingSecond{pending}`/`SkippingBurst` are never round-tripped through state_json (the roundtrip test only covers the non-fork path); (c) the wall-clock flush can misfire when a pass reaches EOF between burst events >1s after the first was written, releasing replayed parent history as real usage and setting Done so the rest of the burst is also counted; and (d) the entry-key doc claim that dedup catches replays is wrong for rewritten bursts (Codex rewrites replay timestamps to the fork instant, so keys never match the parent's).

**Failure scenario:** A serde change to ReplayState/PendingEvent, or dropping either pending_flush override, permanently undercounts every forked/subagent Codex session's first turn - or counts replayed parent history (overcount) whenever a notification-driven pass lands mid-burst - with all existing tests green. Needed: a two-pass test that parks in pass 1 (fresh timestamps) and asserts release by a later sweep pass; a state roundtrip test through AwaitingSecond-with-pending and SkippingBurst; and a burst-split-across-passes test.

src/token_usage/codex.rs
```diff
@@ -0,0 +1,848 @@
...
+        let last_usage = info
+            .and_then(|info| info.last_token_usage)
+            .map(CodexTotals::normalized);
+        if total_usage.is_none() && last_usage.is_none() {
+            return Vec::new();
+        }
+        let cumulative_advanced =
+            total_usage.is_none_or(|totals| self.state.prev_totals != Some(totals));
+        let delta = last_usage
+            .filter(|_| cumulative_advanced)
```
svarlamov-git-ai, line 276: **[Token-usage accuracy review — finding 8/15]** (stack-wide review vs upstream ccusage)

The load-bearing `.filter(|_| cumulative_advanced)` on last_token_usage is unpinned: no git-ai test feeds a token_count event carrying a non-zero `last_token_usage` alongside an UNCHANGED `total_token_usage` (Codex duplicates final snapshots on close/compaction), the case upstream pins twice (`skips_repeated_last_usage_when_cumulative_total_is_unchanged`, `skips_missing_parent_replay_when_duplicate_snapshot_is_suppressed`); the existing `skips_repeats_of_unchanged_cumulative_totals` uses total-only lines.

**Failure scenario:** A refactor simplifying the delta selection to `last_usage.or_else(...)` passes every existing test yet re-counts the full final-turn usage on every duplicated snapshot - systematic inflation for all Codex sessions. One extractor test with (unchanged total + repeated last) closes it.

src/token_usage/cost.rs
```diff
@@ -0,0 +1,120 @@
...
+
+/// 1-hour ephemeral cache writes are priced at 2x the input rate (ccusage
+/// `CACHE_CREATE_1H_INPUT_MULTIPLIER`); the flat `cache_write` rate covers
+/// the 5-minute TTL.
+const CACHE_WRITE_1H_INPUT_MULTIPLIER: f64 = 2.0;
+
+/// Estimated cost of one entry in micro-USD: the transcript's own `costUSD`
+/// when present, otherwise computed from the models.dev pricing catalog.
+/// `None` when the model has no known pricing.
+pub fn entry_cost_micro_usd(entry: &UsageEntry) -> Option<u64> {
```
svarlamov-git-ai, line 34: **[Token-usage accuracy review — finding 11/15]** (stack-wide review vs upstream ccusage)

Cost-estimation divergences from upstream are untested and partly undocumented: (a) cache_read is priced at $0 for any model whose catalog entry omits a cache-read rate (ModelPricing defaults 0.0) where upstream falls back to the input rate (pinned by `charges_cached_input_at_input_rate_when_codex_pricing_omits_cache_read_rate`) - cache reads are the dominant Codex token class; (b) the fast-speed multiplier is dropped (upstream: 6x for claude-opus-4-6/4-7) with no test pinning any cost for a has_speed entry; (c) long-context tiered pricing dropped (documented in claude.rs but not codex.rs, ~2x undercount on the most expensive turns); (d) codex-auto-review resolves to no pricing at all -> $0 and wrong model attribution; (e) `micro_usd` has no sanity bound: a corrupt finite costUSD like 1e15 saturates to u64::MAX then clamps to the TOKEN count ceiling = a $1,000,000 bucket cost faithfully emitted, and negative costUSD becomes Some(0), suppressing catalog pricing entirely; (f) the result also disagrees with the existing `estimate_cost` in src/metrics/local_stats.rs:582 (which lacks the 1h-cache multiplier), so `git-ai usage` and uploaded TokenUsage report different dollars for the same tokens.

**Failure scenario:** A fast-mode opus session with no transcript costUSD reports 1/6 of the ccusage cost; one malformed line inflates a bucket by $1M; any cache-heavy model without a catalog cache_read rate reports near-zero cost. Each behavior needs a pinning test (or a documented deviation note), a costUSD sanity clamp distinct from the token ceiling, and one shared per-token-bundle cost helper so the two estimators cannot drift.

src/token_usage/claude.rs
```diff
@@ -0,0 +1,590 @@
...
+    speed: Option<Speed>,
+    #[serde(default)]
+    cache_creation: Option<RawCacheCreation>,
+}
+
+/// ccusage `Speed`: unknown values fail the line's parse, matching ccusage's
+/// strict schema.
+#[derive(Deserialize, Clone, Copy)]
+#[serde(rename_all = "lowercase")]
+enum Speed {
```
svarlamov-git-ai, line 94: **[Token-usage accuracy review — finding 15/15]** (stack-wide review vs upstream ccusage)

Bundle of upstream pinning tests not carried over plus small confirmed gaps, each individually able to let an accuracy regression ship: the :null rejection list is pinned only for requestId (upstream also pins speed/model/sessionId - dropping one entry silently changes counts); the empty-string validity clauses for sessionId/requestId/model are untested (an empty requestId would collide dedup key `m1|` with the absent-request-id key); millisecond precision of burst detection is unpinned (second-truncation would misclassify real turns straddling a second boundary, upstream pins 00.986->01.009); saturating total derivation has no near-u64::MAX test (upstream pins it; reachable via string-encoded values); self-parent forks (forked_from_id == own id) are unhandled/untested (upstream guards and pins); advisor cost composition is never asserted through the DB (upstream pins parent costUSD 1.23 + advisor 26.0); emitted_seq is never asserted on real e2e events despite being the server's ordering key; no insta snapshot pins the raw sparse wire encoding of event id 9; and v1->v2 migration corrections emit with an empty-string external_session_id (the '' backfill default is exactly what flagged v1 rows have, and sessions whose transcripts are gone are never repaired).

**Failure scenario:** Each is a one-line regression with green CI: e.g. removing b"model" from is_unsupported_nullable_field starts counting lines ccusage rejects; routing codex timestamps through a seconds-precision parser silently mis-skips fork first turns; a token_usage_pos renumbering ships undetected and corrupts server-side field decoding. Port the enumerated upstream tests and add the seq/wire/e2e assertions - all are cheap unit tests.

src/token_usage/codex.rs
svarlamov-git-ai: Fixed as suggested — `restore_state` now returns success, and the worker resets the read cursor to 0 alongside a failed restore (the shrunken-file path's recovery), so a cross-version or corrupt state can never book the session's cumulative history as one fresh delta; the full re-read dedups by entry identity. Test: `unreadable_state_resets_the_cursor_and_recounts_idempotently` (corrupt state + nonzero offset + a new turn: only the new turn emits, per-bucket counts unchanged).

src/token_usage/codex.rs
svarlamov-git-ai: Fixed — zero-delta events now neither extend nor resolve an active `SkippingBurst` (upstream's skip machine never sees them), and when a zero-delta second event confirms a burst the skip anchor stays at the first event's timestamp. Your exact scenario (delta t=0, zero repeat t=0.9s, child turn t=1.7s) is pinned in `zero_delta_repeats_do_not_extend_an_active_burst` — the child turn now counts.

src/token_usage/codex.rs
svarlamov-git-ai: Addressed with a hostile-input battery in codex.rs: per-field garbage now degrades per field via `lossy_u64` (strings/negatives/floats/objects → 0 for that field only), pinned in `hostile_token_counts_degrade_per_field`; `hostile_shapes_never_panic_and_never_corrupt_state` runs deep-nested/wrong-typed/huge-number lines through a full extract→state→restore cycle and asserts subsequent well-formed turns still count correctly. The one deliberate deviation from upstream (a line whose overall shape fails serde is skipped whole rather than field-salvaged) is documented in the module header.

src/token_usage/codex.rs
svarlamov-git-ai: Addressed — the fork/burst state machine now roundtrips through every state: `state_roundtrip_through_awaiting_second_matches_single_pass` and `state_roundtrip_through_skipping_burst_matches_single_pass` serialize/restore mid-`AwaitingSecond` (with a parked pending turn) and mid-`SkippingBurst` and assert identical output to a single pass; `flush_releases_parked_turn_after_burst_window` covers park→persist→restore→flush. The doc comment you flagged was wrong and is fixed (entry keys are content-derived; the burst heuristic is what suppresses fork replays). On the flush-timing concern: `flush` only releases a parked turn once `now_ms` is past the 1s window relative to the parked event's own timestamp, and a genuine burst second event is by definition written within that window, so an EOF flush between the two events can only happen after the window has already expired — at which point releasing is the correct (non-burst) interpretation.

src/token_usage/codex.rs
svarlamov-git-ai: Fixed — `total_token_usage` advancing while `last_token_usage` repeats the previous turn's numbers no longer books the stale last: preference for `last_token_usage` is gated on it being consistent with the cumulative delta, falling back to the per-field cumulative subtraction otherwise. Pinned in `stale_last_usage_with_advancing_total_uses_the_delta` (and the inverse gate in `last_usage_preferred_when_consistent`).

src/token_usage/cost.rs
svarlamov-git-ai: Addressed — (a) cache-read pricing missing from the catalog now falls back to input×0.1 (ccusage's default; cache reads dominate Codex traffic, so $0 badly undercounted) with a unit test on synthetic pricing; (b) transcript `costUSD` is sanitized: negative/non-finite values are dropped to None, and every entry cost is clamped to a $10k sanity ceiling (`MAX_ENTRY_COST_MICRO_USD`) so one corrupt line can't inflate a bucket by millions; (c) fast-speed entries are pinned at base rate in `fast_speed_costs_base_rate_pinned` and the remaining deviations from ccusage (no long-context tiers, no speed multipliers, no release-date model mapping) are documented in the cost.rs module header alongside a note that TokenUsage — not `git-ai usage` — is the authoritative figure.

src/token_usage/claude.rs
svarlamov-git-ai: Ported/addressed item by item: self-parent fork guard (payload.id == forked_from_id ignored, tested); ms-precision burst window (timestamps kept in millis end-to-end, `burst_window_is_millisecond_precise` pins 999ms vs 1001ms); saturating arithmetic throughout delta/aggregate math with the 1e12 `TOKEN_VALUE_CEILING` clamp at the DB boundary; `emitted_seq` now asserted >0 in the happy-path e2e; the sparse wire encoding is pinned with an insta snapshot (`test_token_usage_wire_encoding_is_pinned`) so a position renumbering can't slip through; and v1-migrated reconcile corrections no longer emit an empty-string `external_session_id` — the attribute is omitted when no id was recorded. Advisor-cost composition is covered at the extractor level (parent costUSD honored, advisor iterations priced from catalog with `:advisor:{i}` key salts) — the DB layer is model-agnostic on cost, so a separate through-DB test would re-test the same seam.

src/token_usage/codex.rs
svarlamov-git-ai: **Re-review at c2977628a: PARTIAL (extractor level fixed, worker level open).** `fork_states_roundtrip_through_persisted_state` covers AwaitingSecond-with-pending, SkippingBurst persistence, and flush-in-a-later-pass — good. Two residuals:
1. The worker-side park/release chain is still never exercised across two passes: `pending_flush` forcing a sweep re-enqueue is untested (`sweep_enqueues_only_changed_supported_transcripts` has no pending_flush case), and the quiet-skip bypass via `has_pending()` is only hit in the same-pass flush test. Dropping either override silently strands parked turns forever with green tests.
2. The wall-clock flush misfire is neither prevented nor listed under the spec's new "Accepted limitations": a pass that reaches EOF between burst events >1s after the first was written releases replayed parent history as real usage and sets `Done`, so the rest of the burst also counts. Rare (requires catching the file mid-burst-write), but it's an overcount path — worth a line in the accepted-limitations section if it's accepted.

svarlamov-git-ai: Closed both halves — `parked_fork_stays_pending_and_a_later_sweep_pass_releases_it` now drives the full worker chain across two real passes through `process_task_blocking`: pass 1 parks the turn and persists `pending_flush`, the byte-unchanged file is asserted to remain a sweep candidate, and pass 2 releases and emits after the burst window, clearing the flag and the candidacy. The wall-clock flush misfire is now listed in the spec's accepted limitations (a fork burst split across passes by >1s of transcript write lag can over-count that one pair — a race upstream's post-hoc whole-file reads can't observe).

tests/fixtures/codex-session-token-count.jsonl
```diff
@@ -0,0 +1,10 @@
+{"timestamp":"2026-08-18T20:34:01.217Z","ordinal":0,"type":"session_meta","payload":{"session_id":"01a00000-0000-7000-8000-000000000001","id":"01a00000-0000-7000-8000-000000000001","timestamp":"2026-08-18T20:32:51.906Z","cwd":"/home/user/project","originator":"codex-tui","cli_version":"0.147.0","source":"cli","thread_source":"user","model_provider":"openai","base_instructions":{"text":"You are Codex."},"history_mode":"paginated","context_window":{"window_id":"01a01693-acbb-7d93-93d6-e3fa4c6570c6"},"git":{"commit_hash":"233745c085a64cacf8a95e47b07b36ce73a2f6ae","branch":"feat/contributor-import-scm-usernames","repository_url":"https://github.com/next-element-inc/monorepo"}}}
```
svarlamov-git-ai: **[Token-usage accuracy review — new finding: fixture sanitization slip]** This capture was otherwise carefully sanitized (zeroed UUIDs, `/home/user/project`, generic base instructions), but the `session_meta.git` block kept what looks like real internal data: `"repository_url":"https://github.com/next-element-inc/monorepo"`, a real 40-hex `commit_hash`, and the branch name `feat/contributor-import-scm-usernames`. This repo is public, so that leaks a private org/repo name, internal branch/feature name, and a commit hash. Please replace the `git` block with placeholder values (e.g. `https://github.com/example/project`, zeroed hash) — the extractor never reads it, so the golden snapshot is unaffected.

src/token_usage/codex.rs
svarlamov-git-ai: **Re-review at a981487e1: park/release chain FIXED; split-burst doc needs one correction.**
- `parked_fork_stays_pending_and_a_later_sweep_pass_releases_it` covers the full worker chain across two real passes — pending_flush persisted, unchanged-bytes file kept as a sweep candidate (pins the `sweep_candidates` pending_flush link), release + emission in pass 2, flag and candidate cleared. Exactly what I asked for. (Nit: the 2.1s wall-clock sleep has a small flake window — if >1s elapses between writing the fixture and pass 1's flush on a loaded CI box, pass 1 releases early and the "stays parked" assert fails. Stamping the fixture's token_count a couple of seconds in the future would remove it.)
- The new accepted-limitation is honest about the misfire's existence, but the bound claim is wrong: "bounded to that one pair per fork" only holds for a 2-event burst. `flush` transitions `replay` to `Done` (codex.rs `std::mem::replace(&mut self.state.replay, ReplayState::Done)`), so when a rewritten burst has N>2 events and a pass lands in the write gap, the parked event is released AND every remaining replayed line then counts through the `Done` arm — worst case the parent's entire history, same magnitude as the restore_state bug we fixed. Either correct the wording ("the parked replay and the remainder of the burst"), or make the stated bound true cheaply: have flush transition to a `FlushedAwaitingNext { first_ts_ms }` state whose next event, if within the burst window of `first_ts_ms`, enters `SkippingBurst` (over-count then truly bounded to the single released event).

tests/fixtures/codex-session-token-count.jsonl
svarlamov-git-ai: Fixed and scrubbed from history — the `session_meta.git` block now carries placeholders (`https://github.com/example/repo`, an all-zero commit hash, branch `main`); the golden snapshot is byte-identical since the git block is never parsed. The fix was amended into the fixture commit and the branch force-pushed, so the leaked values are no longer reachable from the PR branch's history. Caveat: GitHub can retain unreachable objects server-side for a while (old commit SHAs referenced in PR timeline events may still resolve), so if the values are considered sensitive beyond embarrassment, org-level GC/support is the only complete purge.

src/token_usage/codex.rs
svarlamov-git-ai: Both follow-ups addressed — (1) the split-burst hardening you and the round-5 reviewer converged on is in: `flush` arms `SkippingBurst` at the released event's timestamp instead of disarming to `Done`, bounding the misfire to the one released event even for long bursts (test: `late_burst_partners_after_a_flush_release_are_skipped`; the single-turn-fork case is unaffected and now also asserts the own next turn counts). (2) The chain test's flake window is closed: the fixture's token_count is stamped two seconds in the future so a loaded CI box cannot burn through the burst window before pass 1's flush; pass 2's sleep covers the stamp plus window plus the flush clock's second granularity.

PR #2207, opened 2026-08-24
Add token-usage state database with dedup and bucket aggregation

svarlamov-git-ai:
## Summary

Second PR of the TokenUsage stack: `src/token_usage/db.rs`, a single SQLite database (`~/.git-ai/internal/token-usage-db`) holding everything the pipeline needs to stay consistent:

- `tracked_files` — the authoritative read cursor (byte offset) + serialized extractor state per transcript file, plus the quiet-skip size/mtime snapshot, error-backoff state (`processing_errors`/`last_error_at`), and a pending-flush marker. **Keeping the cursor here (not in the streams database) is load-bearing**: `commit_batch` writes entries, extractor state, and the advanced cursor in one transaction, so a crash can never replay transcript lines against post-batch parser state (which would corrupt Codex's cumulative-delta computation).
- `usage_entries` — deduplicated per-entry usage. Per-entry rows (not bucket accumulators) make ccusage's replacement policy exact: a later entry can *lower* a bucket, and the bucket then re-aggregates via SQL. **Dedup is global across sessions** (matching ccusage's whole-files dedup): `claude --resume`/`--continue`/fork copies the parent conversation into a new session file with the original message/request ids, and session-scoped dedup would re-count that history on every resume. When a replacement moves an entry between sessions, `commit_batch` reports the previous owner (and `invalidate_session_files` forces its re-reconciliation). Entries older than the retention cutoff are dropped at insert.
- `bucket_state` — fingerprint, **emission revision (`emit_seq`)**, and timestamp of the last emitted aggregate per `(session_id, model, bucket_ts)`; the fingerprint covers every emitted field so any change — including a drop to zero — triggers re-emission, and the revision orders same-second re-emissions.

Changed buckets are found by **reconciliation** in a single grouped query (`changed_buckets`): one aggregate pass over the session's entries plus its emptied buckets from `bucket_state`, returning only fingerprint mismatches. No pending-emission state lives in memory, so a crash or failed emission is healed by the next pass. Retention pruning is transactional (a partial prune could otherwise strand orphan `bucket_state` rows that re-emit zero over real historical data), and token counts clamp to `i64::MAX` at the DB boundary.

Follows the codebase's SQLite conventions (`open_with_memory_limits`, WAL/NORMAL/MEMORY pragmas, versioned migrations — v2 adds the global dedup indexes and new columns).

## Testing

- 21 unit tests: migrations, cursor/state atomicity, resume/fork global dedup (identical copy not re-counted; cross-session replacement reported), replacement matrix, single-query reconciliation incl. emptied buckets and revision bumps, retention-at-insert, clamping, transactional prune, error/backoff state, snapshot invalidation.

### DIFF
--- a/src/token_usage/db.rs
+++ b/src/token_usage/db.rs
@@ -6,20 +6,30 @@
 //! - `tracked_files`: the authoritative read cursor (byte offset) and
 //!   serialized extractor state per transcript file,
 //! - `usage_entries`: deduplicated per-entry token usage,
-//! - `bucket_state`: the fingerprint of the last emitted aggregate per
-//!   `(session_id, model, bucket_ts)`, so unchanged buckets are never
-//!   re-emitted.
+//! - `bucket_state`: the fingerprint and revision of the last emitted
+//!   aggregate per `(session_id, model, bucket_ts)`, so unchanged buckets are
+//!   never re-emitted.
 //!
 //! Keeping the cursor here (rather than in the streams database) matters:
 //! [`TokenUsageDatabase::commit_batch`] writes the entries, the extractor
 //! state, and the advanced cursor in a single transaction, so a crash can
 //! never replay lines against post-batch parser state.
 //!
+//! Entry deduplication is **global across sessions**, matching ccusage's
+//! whole-files dedup: `claude --resume`/`--continue`/fork writes a new
+//! transcript whose leading lines are copies of the parent conversation with
+//! the original message/request ids, and git-ai tracks that file as a new
+//! session. A session-scoped dedup would re-count the copied history on
+//! every resume. The first-seen row keeps the entry (and its session
+//! attribution); the replacement policy can move a row to the replacing
+//! session, in which case [`TokenUsageDatabase::commit_batch`] reports the
+//! previous session so its buckets can be reconciled too.
+//!
 //! Changed buckets are found by *reconciliation*, not change tracking:
-//! [`TokenUsageDatabase::session_buckets`] enumerates every bucket the
-//! session has entries or emission state for, and the caller re-emits those
-//! whose current aggregate fingerprint differs from the emitted one. Because
-//! no pending-emission state lives in memory, a crash or failed emission is
+//! [`TokenUsageDatabase::changed_buckets`] aggregates every bucket the
+//! session has entries or emission state for in one pass and returns those
+//! whose fingerprint differs from the last emitted one. Because no
+//! pending-emission state lives in memory, a crash or failed emission is
 //! healed by the next pass over the session.
 
 use std::path::Path;
@@ -87,7 +97,25 @@ const MIGRATIONS: &[&str] = &[
 
     INSERT INTO schema_version (version) VALUES (1);
     "#,
+    // Version 2: global dedup indexes (resume/fork dedup crosses sessions),
+    // pending-flush marker + error backoff timestamp on tracked files, and a
+    // per-bucket emission revision.
+    r#"
+    CREATE INDEX IF NOT EXISTS idx_usage_entries_key
+        ON usage_entries(entry_key);
+    CREATE INDEX IF NOT EXISTS idx_usage_entries_message_global
+        ON usage_entries(message_id) WHERE message_id IS NOT NULL;
+
+    ALTER TABLE tracked_files ADD COLUMN pending_flush INTEGER NOT NULL DEFAULT 0;
+    ALTER TABLE tracked_files ADD COLUMN last_error_at INTEGER;
+    ALTER TABLE bucket_state ADD COLUMN emit_seq INTEGER NOT NULL DEFAULT 0;
+
+    INSERT INTO schema_version (version) VALUES (2);
+    "#,
 ];
+
+const TRACKED_FILE_COLUMNS: &str = "session_id, stream_path, tool, byte_offset, state_json, \
+     last_known_size, last_modified, processing_errors, last_error_at, pending_flush";
 
 /// A tracked transcript file: read cursor plus extractor state.
 #[derive(Debug, Clone, PartialEq, Eq)]
@@ -100,6 +128,25 @@ pub struct TrackedFile {
     pub last_known_size: i64,
     pub last_modified: Option<i64>,
     pub processing_errors: i64,
+    pub last_error_at: Option<i64>,
+    /// The extractor reported buffered entries at the end of the last pass;
+    /// the file must be re-processed even if its bytes have not changed.
+    pub pending_flush: bool,
+}
+
+fn read_tracked_file(row: &rusqlite::Row<'_>) -> rusqlite::Result<TrackedFile> {
+    Ok(TrackedFile {
+        session_id: row.get(0)?,
+        stream_path: row.get(1)?,
+        tool: row.get(2)?,
+        byte_offset: row.get::<_, i64>(3)?.max(0) as u64,
+        state_json: row.get(4)?,
+        last_known_size: row.get(5)?,
+        last_modified: row.get(6)?,
+        processing_errors: row.get(7)?,
+        last_error_at: row.get(8)?,
+        pending_flush: row.get(9)?,
+    })
 }
 
 /// The aggregate of one `(session_id, model, bucket_ts)` bucket, i.e. exactly
@@ -136,6 +183,41 @@ impl BucketAggregate {
     }
 }
 
+/// One reconciliation candidate: a bucket whose current aggregate differs
+/// from the last emitted fingerprint.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct ChangedBucket {
+    pub model: String,
+    pub bucket_ts: u32,
+    pub aggregate: BucketAggregate,
+    /// Revision of the last emission for this bucket (0 when never emitted).
+    /// The next emission carries `emit_seq + 1` so the server can order
+    /// same-second re-emissions.
+    pub emit_seq: u64,
+}
+
+/// One extracted batch to persist atomically.
+pub struct BatchCommit<'a> {
+    pub session_id: &'a str,
+    pub stream_path: &'a str,
+    pub entries: &'a [UsageEntry],
+    pub new_offset: u64,
+    pub state_json: Option<&'a str>,
+    /// The extractor still holds buffered entries (see
+    /// `UsageExtractor::has_pending`).
+    pub pending_flush: bool,
+    /// Retention cutoff: entries whose bucket falls before this are dropped
+    /// instead of inserted, so backfill never uploads history the next prune
+    /// would delete.
+    pub min_bucket_ts: u32,
+}
+
+/// Clamp a token count for an INTEGER column (defensive: corrupt transcripts
+/// can carry absurd values).
+fn to_db_i64(value: u64) -> i64 {
+    value.min(i64::MAX as u64) as i64
+}
+
 /// SQLite database for token-usage state.
 pub struct TokenUsageDatabase {
     conn: Arc<Mutex<Connection>>,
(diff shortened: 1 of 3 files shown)


src/token_usage/db.rs
```diff
@@ -0,0 +1,868 @@
...
+        input_tokens INTEGER NOT NULL DEFAULT 0,
+        output_tokens INTEGER NOT NULL DEFAULT 0,
+        cache_read_tokens INTEGER NOT NULL DEFAULT 0,
+        cache_write_tokens INTEGER NOT NULL DEFAULT 0,
+        reasoning_output_tokens INTEGER,
+        total_tokens INTEGER NOT NULL DEFAULT 0,
+        cost_micro_usd INTEGER,
+        is_sidechain INTEGER NOT NULL DEFAULT 0,
+        has_speed INTEGER NOT NULL DEFAULT 0,
+        PRIMARY KEY (session_id, entry_key)
```
svarlamov-git-ai, line 83: **[review finding 1/15 — correctness, high] Per-session dedup double-counts resumed/continued Claude sessions.**

Dedup is strictly session-scoped: `PRIMARY KEY (session_id, entry_key)`, and both `find_dedupe_target` queries filter `session_id = ?1`. ccusage dedups `message_id|request_id` globally across files precisely because `claude --resume`/`--continue`/fork writes a NEW .jsonl whose leading lines are copies of the parent conversation with the original message ids, request ids, and usage. git-ai registers that file as a brand-new session (external_session_id = file stem; `external_parent_session_id` is only set for `subagents/` paths), so the backfill re-inserts the entire copied history under the new session_id and re-emits every historical bucket — the server aggregates across sessions, so a resumed conversation's tokens/cost count twice (N times for N resumes), scaling with conversation length. The port's own doc comment in claude.rs acknowledges the scope was narrowed from ccusage's whole-files dedup. Consider a cross-session dedup for Claude entries (global message_id|request_id identity) or resume-lineage detection that rolls the new file up to the original session.

src/token_usage/db.rs
```diff
@@ -0,0 +1,868 @@
...
+            params![session_id, model, bucket, fingerprint, now_ts],
+        )?;
+        Ok(())
+    }
+
+    /// Retention prune: drop entries and bucket state older than the cutoff
+    /// bucket. Cursors are monotonic, so pruned history is never re-read.
+    pub fn prune_buckets_before(&self, cutoff_bucket_ts: u32) -> Result<usize, GitAiError> {
+        let conn = self.lock();
+        let entries = conn.execute(
```
svarlamov-git-ai: **[review finding 3/15 — correctness/data loss, medium-high] Non-transactional prune can permanently zero real historical buckets server-side.**

`prune_buckets_before` runs two separate autocommit DELETEs (usage_entries, then bucket_state) with no wrapping transaction — unlike `commit_batch`, which uses `conn.transaction()` for exactly this reason. If the daemon dies between them (power loss, force-kill, shutdown-deadline enforcer), orphaned `bucket_state` rows survive with their entries gone. The next pass over any affected session sees those buckets via the `session_buckets` UNION arm, aggregates them to zero, finds a differing fingerprint, and emits zero-value TokenUsage events that the server upserts over the real months-old totals — and since the entries are deleted, the correct values can never be re-emitted. Wrap both DELETEs in one transaction (and consider deleting bucket_state first if kept separate). Secondary: the `usage_entries` DELETE has no index on bare `bucket_ts` (only `(session_id, model, bucket_ts)`), so each 30-minute prune full-scans the table even when there is nothing to delete.

src/token_usage/db.rs
svarlamov-git-ai: Fixed — dedup is now global across sessions, matching ccusage's whole-files semantics (schema v2 adds global `entry_key`/`message_id` indexes). A resumed/forked file's copied history dedups against the original rows, so nothing is re-counted; the first-seen session keeps the attribution. When a copy carries larger totals and the replacement policy moves the row to the new session, `commit_batch` reports the previous owner and the worker invalidates that session's quiet-skip snapshot so its (now lowered/emptied) buckets re-reconcile on their next pass with their own attributes. Covered by `resumed_session_does_not_recount_copied_history`, `cross_session_replacement_reports_the_previous_owner`, and worker-level `resumed_session_copy_is_not_double_counted` / `cross_session_replacement_invalidates_the_previous_owner`.

src/token_usage/db.rs
svarlamov-git-ai: Fixed — both DELETEs now run in one transaction, so a crash can no longer strand orphan `bucket_state` rows that would re-emit zero over real historical data. The prune test also asserts the emptied old bucket is not a reconciliation candidate afterwards.

src/token_usage/db.rs
```diff
@@ -0,0 +1,1191 @@
...
+            "SELECT model, bucket_ts,
+                    COALESCE(SUM(input_tokens), 0),
+                    COALESCE(SUM(output_tokens), 0),
+                    COALESCE(SUM(cache_read_tokens), 0),
+                    COALESCE(SUM(cache_write_tokens), 0),
+                    COUNT(reasoning_output_tokens),
+                    COALESCE(SUM(reasoning_output_tokens), 0),
+                    COALESCE(SUM(total_tokens), 0),
+                    COALESCE(SUM(cost_micro_usd), 0),
+                    COUNT(*)
```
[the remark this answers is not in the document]
svarlamov-git-ai: Good catch — the clamp now targets a sane ceiling (one trillion tokens per entry) instead of i64::MAX, so a bucket full of clamped rows sums far below i64 range and SQLite's SUM can no longer raise an integer-overflow error. The regression test commits two clamped rows into one bucket and asserts both the aggregate and that `changed_buckets` reconciles without erroring.

src/token_usage/db.rs
```diff
@@ -0,0 +1,1203 @@
...
+        last_emitted_at INTEGER NOT NULL,
+        PRIMARY KEY (session_id, model, bucket_ts)
+    );
+
+    INSERT INTO schema_version (version) VALUES (1);
+    "#,
+    // Version 2: global dedup indexes (resume/fork dedup crosses sessions),
+    // pending-flush marker + error backoff timestamp on tracked files, and a
+    // per-bucket emission revision.
+    r#"
```
svarlamov-git-ai, line 111: **[follow-up review — robustness, medium] The v2 migration is neither atomic nor idempotent: a crash mid-batch permanently bricks TokenUsageDatabase::open.**

`migrate()` applies each script with `conn.execute_batch`, which autocommits statement-by-statement — no transaction. This v2 script runs three non-idempotent `ALTER TABLE ... ADD COLUMN` statements before `INSERT INTO schema_version (2)`. A crash or I/O error after the first ALTER but before the version INSERT makes every subsequent `open()` re-run migration 2 and fail forever on 'duplicate column name: pending_flush' — the worker is then never spawned again (daemon.rs logs the open error and continues), silently killing the pipeline until the DB file is deleted by hand. v1 happened to self-heal (CREATE IF NOT EXISTS), so this is new exposure in v2. The codebase convention already exists: the streams DB's multi-statement v4 migration is explicitly wrapped in `BEGIN; ... COMMIT;` (src/streams/db.rs). Wrap this script the same way so the ALTERs and the version row commit atomically.

src/token_usage/db.rs
```diff
@@ -0,0 +1,1203 @@
...
+    session_id: &str,
+    entry: &UsageEntry,
+) -> Result<Option<String>, GitAiError> {
+    let bucket = bucket_ts(entry.ts);
+    let existing = find_dedupe_target(tx, entry)?;
+    match existing {
+        Some(row) => {
+            if should_replace(entry.into(), row.replacement) {
+                tx.execute(
+                    "UPDATE usage_entries SET
```
svarlamov-git-ai, line 748: **[follow-up review — robustness, low-medium] On a v1-upgraded database, a cross-session replacement can violate the (session_id, entry_key) primary key and poison the transcript with an unretryable commit failure.**

Databases written under v1's session-scoped dedup legitimately contain the same `entry_key` under two sessions (exactly the resume duplicates the v2 change targets — the migration keeps those rows). Post-upgrade, when session B replays `m1|r1` with larger totals, `find_dedupe_target`'s global exact lookup (`ORDER BY rowid LIMIT 1`) can pick session A's older row, `should_replace` passes, and this UPDATE sets `session_id = B, entry_key = 'm1|r1'` — colliding with B's own existing `(B, 'm1|r1')` row. SQLite raises a UNIQUE constraint error, `commit_batch` aborts, and every retry (after backoff) re-reads the same line from the unadvanced cursor and fails identically — permanently stalling that file, losing all subsequent usage in it, not just the duplicate. Post-v2-only data cannot produce this (inserts happen only when no global match exists), so the trigger population is upgraded v1 DBs — today's dogfooding machines. Cheap fixes: dedupe legacy duplicate entry_keys in the v2 migration (keep lowest rowid), and/or make the move UPDATE collision-safe (delete the destination-key row in the same transaction, keeping the replacement-policy winner).

src/token_usage/db.rs
svarlamov-git-ai: Fixed — `migrate()` now wraps each migration script in a transaction (`unchecked_transaction` + `execute_batch` + commit), so the v2 ALTERs and the version row land atomically: a crash mid-migration rolls back cleanly and the next open() re-applies from scratch instead of failing forever on a duplicate column.

src/token_usage/db.rs
svarlamov-git-ai: Fixed — the v2 migration now dedups legacy cross-session duplicates (keeping the first-seen row per `entry_key`) before creating the key index, and that index is now UNIQUE, so a cross-session replacement on an upgraded database can no longer collide on the `(session_id, entry_key)` primary key. Regression test `v1_databases_with_legacy_cross_session_duplicates_upgrade_cleanly` builds a real v1 DB with triplicate keys, upgrades it, and exercises the previously-poisonous replacement.

src/token_usage/db.rs
```diff
@@ -0,0 +1,1365 @@
...
+    "#,
+    // Version 2: global dedup (resume/fork dedup crosses sessions),
+    // pending-flush marker + error backoff timestamp + rollup identity on
+    // tracked files, a durable cross-session reconcile flag, and a
+    // per-bucket emission revision. A v1 database may hold cross-session
+    // duplicate keys (its dedup was session-scoped); the first-seen row wins
+    // so the unique index can be created and later replacements cannot
+    // collide on the primary key.
+    r#"
+    DELETE FROM usage_entries WHERE rowid NOT IN (
```
svarlamov-git-ai, line 123: **[round-3 review — correctness, low] The migration's duplicate purge changes session aggregates without flagging them for re-reconciliation, stranding stale emitted fingerprints.**

This DELETE removes the double-counted rows from every session that held a resume/fork copy, but their `bucket_state` rows keep the pre-dedup fingerprints — and nothing schedules a pass: `needs_reconcile` stays 0, quiet-skip size/mtime snapshots still match, and no file changed. Active sessions heal on their next file-driven pass (fingerprint mismatch re-emits), but dormant resumed sessions keep their inflated server-side values indefinitely — the exact staleness the zero-emission/needs_reconcile machinery exists to prevent. Population is limited to pre-release v1 databases, so this is low severity, but the fix is one statement in the same script — `UPDATE tracked_files SET needs_reconcile = 1;` (or scoped to sessions that lost rows via a temp table) — and it establishes the right invariant for any future migration that mutates `usage_entries`: data-changing migrations must schedule reconciliation.

svarlamov-git-ai: Fixed — the v2 script now flags every session that loses rows to the purge (`needs_reconcile = 1`, set before the DELETE in the same atomic migration), so their lowered aggregates re-emit through the normal reconcile path on the next daemon run instead of leaving inflated values on the server. The v1-fixture upgrade test asserts the losing sessions (and only those) come back from `sessions_needing_reconcile`.

PR #2209, opened 2026-08-24
Add token-usage daemon worker with backfill and event emission

svarlamov-git-ai:
## Summary

Fourth PR of the TokenUsage stack: the daemon worker that turns transcript files into TokenUsage metric events, plus wiring and end-to-end tests.

- **Triggers (all off the trace2 ingestion path)**: a non-blocking ping from the stream worker after it successfully processes a transcript, a startup sweep, and a 30-minute ticker (missed ticks skipped, not replayed). Sweeps enumerate the streams database's `transcript` rows, so every session git-ai already knows about is **backfilled** — a newly tracked file starts at byte offset 0 and its full history (within the retention window) is bucketed. Sweep enumeration costs one stat per session compared against the token DB's size/mtime snapshot; settled sessions never reach `spawn_blocking`, DB reads, or repo discovery, and repo discovery is further deferred until events are actually emitted.
- **Queues**: notifications and sweep backfill run on separate queues; the `git-ai await` drain barrier waits only for notification-driven work (with promotion when a notification hits a file queued for backfill), so a first-start historical backfill can never starve `await` into skipping the telemetry flush. Shutdown is observed *during* processing: the worker selects on its own shutdown `Notify` around the in-flight blocking pass.
- **Incremental & quiet**: reads via `read_jsonl_line` from the persisted cursor with the extractor's substring prefilter ahead of any JSON parsing; bounded batches (1000 entries / 4 MiB) commit entries + extractor state + cursor in one token-db transaction; per-file error backoff (5s/30s/5m/30m) keeps permanently failing transcripts from re-reading on every trigger; telemetry backpressure mirrors the stream worker; the retention cutoff is applied at extraction so backfill never uploads history the prune would delete.
- **Emission durability**: changed buckets come from a single grouped reconciliation query; events carry a strictly increasing per-bucket `emitted_seq` (server keeps the highest revision, so same-second re-emissions can't tie); buckets are marked emitted only after the telemetry queue accepted the events, and the quiet-skip snapshot is written only after a fully successful pass — a failed hand-off or crash at any point is healed by the next pass.
- **Session identity**: Claude subagent transcripts roll up to the parent session (ccusage parity); cross-session resume/fork replacements invalidate the previous owner's snapshot so its buckets re-reconcile with their own attributes; forked codex single turns flush at end-of-file once the burst window has passed.
- **Privacy**: repo_url falls back to the agent's cwd inference (persisted back) like the SessionEvent path, closing the exclude-gate fail-open for sweep-backfilled sessions.
- **Feature-flag hygiene**: `token_usage_metrics` gates the worker at startup (like `transcript_streaming`) — when off, no DB is created, no worker/ticker runs, no config re-reads happen, and previously collected data is deleted.
- **Wiring**: spawned next to the stream worker with its own shutdown `Notify` (`request_shutdown` uses `notify_one`, which wakes exactly one waiter); drained in `await_completion` right after the transcript drain; the stream worker pings from both its steady-state and drain success paths.

## Non-negotiable-rules compliance

No git spawns anywhere (repo_url resolves via `resolve_repo_url_from_path` / `infer_cwd`, both gix-based file reads); nothing added to the critical ingestion path (the only touch on an existing path is one non-blocking unbounded-channel send after the stream worker finishes a task); all file/SQLite I/O in `spawn_blocking`.

## Testing

- Worker unit tests with injectable repo resolution + sink: bucket emission with revisions, quiet re-process, incremental append, cross-bucket replacement zeroing, partial trailing lines, retention at extraction, error backoff, codex reasoning flow, fork single-turn flush, cross-session invalidation, resume no-double-count, sweep prefilter/dedup, notification promotion + drain scoping, rollup error keying, backoff schedule.
- Integration tests (`tests/integration/token_usage.rs`) driving the real pipeline via `git-ai checkpoint` hook inputs behind a `git-ai await` barrier (SyncFamily alone does not cover the stream → token → metrics pipeline): claude and codex happy paths, incremental re-emission, and the disabled-flag lifecycle (no DB created, no events).

### DIFF
--- a/src/daemon.rs
+++ b/src/daemon.rs
@@ -2786,6 +2786,9 @@ pub struct ActorDaemonCoordinator {
     stream_worker: Option<crate::daemon::stream_worker::StreamWorkerHandle>,
     token_usage_worker: Option<crate::daemon::token_usage_worker::TokenUsageWorkerHandle>,
     transcript_shutdown_notify: std::sync::OnceLock<Arc<tokio::sync::Notify>>,
+    // Separate from the transcript worker's Notify: notify_one wakes exactly
+    // one waiter, so each worker needs its own.
+    token_usage_shutdown_notify: std::sync::OnceLock<Arc<tokio::sync::Notify>>,
     streams_db: Option<Arc<crate::streams::db::StreamsDatabase>>,
     next_trace_ingest_seq: AtomicUsize,
     queued_trace_payloads: AtomicUsize,
@@ -2917,6 +2920,7 @@ impl ActorDaemonCoordinator {
             stream_worker: None,
             token_usage_worker: None,
             transcript_shutdown_notify: std::sync::OnceLock::new(),
+            token_usage_shutdown_notify: std::sync::OnceLock::new(),
             streams_db: None,
             next_trace_ingest_seq: AtomicUsize::new(0),
             queued_trace_payloads: AtomicUsize::new(0),
@@ -3232,6 +3236,9 @@ impl ActorDaemonCoordinator {
         self.shutdown_notify.notify_waiters();
         if let Some(transcript_shutdown) = self.transcript_shutdown_notify.get() {
             transcript_shutdown.notify_one();
+        }
+        if let Some(token_usage_shutdown) = self.token_usage_shutdown_notify.get() {
+            token_usage_shutdown.notify_one();
         }
         // Hold the condvar mutex so notify_all cannot race with the
         // check-then-wait sequence in daemon_update_check_loop.
@@ -9253,6 +9260,9 @@ pub(crate) async fn run_daemon(config: DaemonConfig) -> Result<DaemonExitAction,
             Ok(streams_db) => {
                 let streams_db = std::sync::Arc::new(streams_db);
                 let shutdown_notify = Arc::new(tokio::sync::Notify::new());
+                // Each worker needs its own shutdown Notify: request_shutdown
+                // uses notify_one, which wakes exactly one waiter.
+                let token_usage_shutdown_notify = Arc::new(tokio::sync::Notify::new());
                 // The token-usage worker is fed by stream-worker completions,
                 // so it lives inside the transcript_streaming gate. Its own
                 // token_usage_metrics flag is re-read per trigger.
@@ -9264,7 +9274,7 @@ pub(crate) async fn run_daemon(config: DaemonConfig) -> Result<DaemonExitAction,
                             streams_db.clone(),
                             std::sync::Arc::new(token_db),
                             telemetry_handle.clone(),
-                            shutdown_notify.clone(),
+                            token_usage_shutdown_notify.clone(),
                         ))
                     }
                     Err(e) => {
@@ -9284,6 +9294,9 @@ pub(crate) async fn run_daemon(config: DaemonConfig) -> Result<DaemonExitAction,
                 let _ = coordinator_inner
                     .transcript_shutdown_notify
                     .set(shutdown_notify);
+                let _ = coordinator_inner
+                    .token_usage_shutdown_notify
+                    .set(token_usage_shutdown_notify);
                 tracing::info!("transcript worker spawned");
             }
             Err(e) => {
--- a/src/daemon.rs
+++ b/src/daemon.rs
@@ -73,6 +73,7 @@ pub mod sweep_coordinator;
 pub mod telemetry_handle;
 pub mod telemetry_worker;
 pub mod test_sync;
+pub mod token_usage_worker;
 pub mod trace_normalizer;
 pub mod transcript_redaction;
 
@@ -2783,6 +2784,7 @@ pub struct ActorDaemonCoordinator {
     trace_ingest_tx: std::sync::OnceLock<mpsc::Sender<Value>>,
     telemetry_worker: Option<crate::daemon::telemetry_worker::DaemonTelemetryWorkerHandle>,
     stream_worker: Option<crate::daemon::stream_worker::StreamWorkerHandle>,
+    token_usage_worker: Option<crate::daemon::token_usage_worker::TokenUsageWorkerHandle>,
     transcript_shutdown_notify: std::sync::OnceLock<Arc<tokio::sync::Notify>>,
     streams_db: Option<Arc<crate::streams::db::StreamsDatabase>>,
     next_trace_ingest_seq: AtomicUsize,
@@ -2913,6 +2915,7 @@ impl ActorDaemonCoordinator {
             trace_ingest_tx: std::sync::OnceLock::new(),
             telemetry_worker: None,
             stream_worker: None,
+            token_usage_worker: None,
             transcript_shutdown_notify: std::sync::OnceLock::new(),
             streams_db: None,
             next_trace_ingest_seq: AtomicUsize::new(0),
@@ -7070,6 +7073,29 @@ impl ActorDaemonCoordinator {
             }
         }
 
+        // Phase 2b: drain the token-usage worker. It is fed by stream-worker
+        // completions, so this must run after the transcript drain above.
+        if !result.timed_out
+            && let Some(worker) = &self.token_usage_worker
+        {
+            let now = Instant::now();
+            if now < deadline {
+                let remaining = deadline - now;
+                maybe_log("token-usage processing");
+                match timeout(remaining, worker.drain()).await {
+                    Ok(Ok(())) => {}
+                    Ok(Err(e)) => {
+                        tracing::warn!(error = %e, "await: token-usage drain failed");
+                    }
+                    Err(_) => {
+                        result.timed_out = true;
+                    }
+                }
+            } else {
+                result.timed_out = true;
+            }
+        }
+
         // Phase 3: flush telemetry and wait for the worker to finish.
         if !result.timed_out
             && let Some(worker) = &self.telemetry_worker
@@ -9227,13 +9253,34 @@ pub(crate) async fn run_daemon(config: DaemonConfig) -> Result<DaemonExitAction,
             Ok(streams_db) => {
                 let streams_db = std::sync::Arc::new(streams_db);
                 let shutdown_notify = Arc::new(tokio::sync::Notify::new());
+                // The token-usage worker is fed by stream-worker completions,
+                // so it lives inside the transcript_streaming gate. Its own
+                // token_usage_metrics flag is re-read per trigger.
+                let token_usage_handle = match crate::token_usage::db::TokenUsageDatabase::open(
+                    config.internal_dir.join("token-usage-db"),
+                ) {
+                    Ok(token_db) => {
+                        Some(crate::daemon::token_usage_worker::spawn_token_usage_worker(
+                            streams_db.clone(),
+                            std::sync::Arc::new(token_db),
+                            telemetry_handle.clone(),
+                            shutdown_notify.clone(),
+                        ))
+                    }
+                    Err(e) => {
+                        tracing::error!(error = %e, "failed to open token-usage database, token-usage worker not started");
+                        None
+                    }
+                };
                 let transcript_handle = crate::daemon::stream_worker::spawn_stream_worker(
                     streams_db.clone(),
                     telemetry_handle.clone(),
                     shutdown_notify.clone(),
+                    token_usage_handle.clone(),
                 );
                 coordinator_inner.streams_db = Some(streams_db);
                 coordinator_inner.stream_worker = Some(transcript_handle);
+                coordinator_inner.token_usage_worker = token_usage_handle;
                 let _ = coordinator_inner
                     .transcript_shutdown_notify
                     .set(shutdown_notify);
--- a/src/daemon.rs
+++ b/src/daemon.rs
@@ -9264,23 +9264,34 @@ pub(crate) async fn run_daemon(config: DaemonConfig) -> Result<DaemonExitAction,
                 // uses notify_one, which wakes exactly one waiter.
                 let token_usage_shutdown_notify = Arc::new(tokio::sync::Notify::new());
                 // The token-usage worker is fed by stream-worker completions,
-                // so it lives inside the transcript_streaming gate. Its own
-                // token_usage_metrics flag is re-read per trigger.
-                let token_usage_handle = match crate::token_usage::db::TokenUsageDatabase::open(
-                    config.internal_dir.join("token-usage-db"),
-                ) {
-                    Ok(token_db) => {
-                        Some(crate::daemon::token_usage_worker::spawn_token_usage_worker(
-                            streams_db.clone(),
-                            std::sync::Arc::new(token_db),
-                            telemetry_handle.clone(),
-                            token_usage_shutdown_notify.clone(),
-                        ))
+                // so it lives inside the transcript_streaming gate, behind
+                // its own startup flag (like transcript_streaming itself).
+                // When the flag is off, nothing runs, no database is created,
+                // and any previously collected data is deleted.
+                let token_usage_db_path = config.internal_dir.join("token-usage-db");
+                let token_usage_handle = if config::Config::get()
+                    .get_feature_flags()
+                    .token_usage_metrics
+                {
+                    match crate::token_usage::db::TokenUsageDatabase::open(&token_usage_db_path) {
+                        Ok(token_db) => {
+                            Some(crate::daemon::token_usage_worker::spawn_token_usage_worker(
+                                streams_db.clone(),
+                                std::sync::Arc::new(token_db),
+                                telemetry_handle.clone(),
+                                token_usage_shutdown_notify.clone(),
+                            ))
+                        }
+                        Err(e) => {
+                            tracing::error!(error = %e, "failed to open token-usage database, token-usage worker not started");
+                            None
+                        }
                     }
-                    Err(e) => {
-                        tracing::error!(error = %e, "failed to open token-usage database, token-usage worker not started");
-                        None
-                    }
+                } else {
+                    crate::token_usage::db::TokenUsageDatabase::remove_database_files(
+                        &token_usage_db_path,
+                    );
+                    None
                 };
                 let transcript_handle = crate::daemon::stream_worker::spawn_stream_worker(
                     streams_db.clone(),
--- a/src/daemon.rs
+++ b/src/daemon.rs
@@ -9248,6 +9248,17 @@ pub(crate) async fn run_daemon(config: DaemonConfig) -> Result<DaemonExitAction,
     crate::daemon::telemetry_worker::set_daemon_internal_telemetry(telemetry_handle.clone());
     coordinator_inner.telemetry_worker = Some(telemetry_handle.clone());
 
+    // With the token-usage flag off, previously collected data is deleted
+    // regardless of the streaming gate below (the spec's "no collected data
+    // is retained" holds even when transcript_streaming is also off).
+    let token_usage_db_path = config.internal_dir.join("token-usage-db");
+    let token_usage_enabled = config::Config::get()
+        .get_feature_flags()
+        .token_usage_metrics;
+    if !token_usage_enabled {
+        crate::token_usage::db::TokenUsageDatabase::remove_database_files(&token_usage_db_path);
+    }
+
     // Spawn the transcript worker BEFORE wrapping coordinator in Arc
     if config::Config::get()
         .get_feature_flags()
@@ -9266,13 +9277,10 @@ pub(crate) async fn run_daemon(config: DaemonConfig) -> Result<DaemonExitAction,
                 // The token-usage worker is fed by stream-worker completions,
                 // so it lives inside the transcript_streaming gate, behind
                 // its own startup flag (like transcript_streaming itself).
-                // When the flag is off, nothing runs, no database is created,
-                // and any previously collected data is deleted.
-                let token_usage_db_path = config.internal_dir.join("token-usage-db");
-                let token_usage_handle = if config::Config::get()
-                    .get_feature_flags()
-                    .token_usage_metrics
-                {
+                // When the flag is off, nothing runs and no database is
+                // created (deletion of old data happens above, outside this
+                // gate).
+                let token_usage_handle = if token_usage_enabled {
                     match crate::token_usage::db::TokenUsageDatabase::open(&token_usage_db_path) {
                         Ok(token_db) => {
                             Some(crate::daemon::token_usage_worker::spawn_token_usage_worker(
@@ -9288,9 +9296,6 @@ pub(crate) async fn run_daemon(config: DaemonConfig) -> Result<DaemonExitAction,
                         }
                     }
                 } else {
-                    crate::token_usage::db::TokenUsageDatabase::remove_database_files(
-                        &token_usage_db_path,
-                    );
                     None
                 };
                 let transcript_handle = crate::daemon::stream_worker::spawn_stream_worker(
--- a/src/daemon/stream_worker.rs
+++ b/src/daemon/stream_worker.rs
@@ -312,6 +312,9 @@ struct StreamWorker {
     sweep_rx: tokio::sync::mpsc::UnboundedReceiver<SweepRequest>,
     drain_rx: tokio::sync::mpsc::UnboundedReceiver<DrainRequest>,
     sweep_trigger_gate: SweepTriggerGate,
+    /// Pinged after a transcript task succeeds so token usage can be
+    /// recomputed for that session (cheap non-blocking send).
+    token_usage_worker: Option<crate::daemon::token_usage_worker::TokenUsageWorkerHandle>,
 }
 
 impl StreamWorker {
@@ -326,6 +329,7 @@ impl StreamWorker {
         sweep_rx: tokio::sync::mpsc::UnboundedReceiver<SweepRequest>,
         drain_rx: tokio::sync::mpsc::UnboundedReceiver<DrainRequest>,
         sweep_trigger_gate: SweepTriggerGate,
+        token_usage_worker: Option<crate::daemon::token_usage_worker::TokenUsageWorkerHandle>,
     ) -> Self {
         let sweep_coordinator =
             crate::daemon::sweep_coordinator::SweepCoordinator::new(streams_db.clone());
@@ -343,6 +347,7 @@ impl StreamWorker {
             sweep_rx,
             drain_rx,
             sweep_trigger_gate,
+            token_usage_worker,
         }
     }
 
@@ -984,7 +989,17 @@ impl StreamWorker {
         // Handle result
         match result {
             Ok(Ok(())) => {
-                // Success - task is done
+                // Success - task is done. Let the token-usage worker pick up
+                // the freshly ingested transcript bytes.
+                if task.stream_kind == "transcript"
+                    && let Some(token_usage) = &self.token_usage_worker
+                {
+                    token_usage.notify_stream_processed(
+                        &task.session_id,
+                        &task.tool,
+                        &task.canonical_path,
+                    );
+                }
             }
             Ok(Err(e)) => {
                 // Error - handle retry logic
@@ -1379,6 +1394,7 @@ pub fn spawn_stream_worker(
     streams_db: Arc<StreamsDatabase>,
     telemetry_handle: DaemonTelemetryWorkerHandle,
     shutdown_notify: Arc<Notify>,
+    token_usage_worker: Option<crate::daemon::token_usage_worker::TokenUsageWorkerHandle>,
 ) -> StreamWorkerHandle {
     let (checkpoint_tx, checkpoint_rx) = tokio::sync::mpsc::unbounded_channel();
     let (sweep_tx, sweep_rx) = tokio::sync::mpsc::unbounded_channel();
@@ -1395,6 +1411,7 @@ pub fn spawn_stream_worker(
         sweep_rx,
         drain_rx,
         sweep_trigger_gate.clone(),
+        token_usage_worker,
     );
 
     tokio::spawn(async move {
@@ -1492,6 +1509,7 @@ mod scheduling_tests {
             sweep_rx,
             drain_rx,
             sweep_trigger_gate,
+            None,
         );
         (temp, worker, shutdown, checkpoint_tx)
     }
@@ -1669,6 +1687,7 @@ mod subagent_sweep_tests {
             sweep_rx,
             drain_rx,
             SweepTriggerGate::new(),
+            None,
         )
     }
 
--- a/src/daemon/stream_worker.rs
+++ b/src/daemon/stream_worker.rs
@@ -989,17 +989,8 @@ impl StreamWorker {
         // Handle result
         match result {
             Ok(Ok(())) => {
-                // Success - task is done. Let the token-usage worker pick up
-                // the freshly ingested transcript bytes.
-                if task.stream_kind == "transcript"
-                    && let Some(token_usage) = &self.token_usage_worker
-                {
-                    token_usage.notify_stream_processed(
-                        &task.session_id,
-                        &task.tool,
-                        &task.canonical_path,
-                    );
-                }
+                // Success - task is done.
+                self.notify_token_usage(&task);
             }
             Ok(Err(e)) => {
                 // Error - handle retry logic
@@ -1357,8 +1348,22 @@ impl StreamWorker {
                 Ok(Err(e)) => {
                     tracing::error!(error = %e, session_id = %task.session_id, "drain task processing error");
                 }
-                Ok(Ok(())) => {}
+                Ok(Ok(())) => {
+                    self.notify_token_usage(&task);
+                }
             }
+        }
+    }
+
+    /// On transcript success, let the token-usage worker pick up the freshly
+    /// ingested bytes (cheap non-blocking send). Called from every task
+    /// success arm — including the drain path — so the await barrier, which
+    /// drains token usage right after transcripts, always sees the ping.
+    fn notify_token_usage(&self, task: &ProcessingTask) {
+        if task.stream_kind == "transcript"
+            && let Some(token_usage) = &self.token_usage_worker
+        {
+            token_usage.notify_stream_processed(&task.session_id, &task.tool, &task.canonical_path);
         }
     }
 
--- a/src/daemon/token_usage_worker.rs
+++ b/src/daemon/token_usage_worker.rs
@@ -48,9 +48,34 @@ use crate::daemon::telemetry_worker::DaemonTelemetryWorkerHandle;
 use crate::error::GitAiError;
 use crate::metrics::{EventAttributes, MetricEvent, PosEncoded, TokenUsageValues};
 use crate::streams::db::{StreamRecord, StreamsDatabase};
-use crate::streams::types::{JsonlLineState, read_jsonl_line};
 use crate::token_usage::db::{BatchCommit, TokenUsageDatabase, TrackedFile};
 use crate::token_usage::extractor_for_tool;
+
+/// One raw JSONL line read as bytes: unlike UTF-8-strict `read_line`, a
+/// single invalid byte cannot wedge the cursor forever (the line decodes
+/// lossily, fails JSON parsing, is skipped, and the cursor advances —
+/// matching upstream ccusage's byte-level reads).
+enum LineRead {
+    Eof,
+    /// No trailing newline: the writer may still be appending.
+    Partial(usize),
+    Complete(usize),
+}
+
+fn read_line_bytes(
+    reader: &mut impl std::io::BufRead,
+    buf: &mut Vec<u8>,
+) -> std::io::Result<LineRead> {
+    buf.clear();
+    let bytes = reader.read_until(b'\n', buf)?;
+    if bytes == 0 {
+        return Ok(LineRead::Eof);
+    }
+    if buf.last() != Some(&b'\n') {
+        return Ok(LineRead::Partial(bytes));
+    }
+    Ok(LineRead::Complete(bytes))
+}
 
 /// Entry/byte bounds of one atomic batch commit.
 const BATCH_MAX_ENTRIES: usize = 1_000;
@@ -513,11 +538,7 @@ fn process_task_blocking(
             repo_url
         },
         &sink,
-    )
-    // Cross-session replacements flagged other sessions during the batch
-    // commits: reconcile them now, DB-only (their files are not needed and
-    // may no longer exist).
-    .and_then(|_| reconcile_flagged_sessions(token_db, &sink));
+    );
     if let Err(e) = &result {
         // Keyed by the rollup session id (subagent transcripts track under
         // their parent), matching the row ensure_file created.
@@ -527,6 +548,16 @@ fn process_task_blocking(
             &e.to_string(),
             now_secs(),
         );
+    }
+    // Cross-session replacements flagged other sessions during the batch
+    // commits: reconcile them now, DB-only. Failures here belong to the
+    // flagged sessions (the durable flag retries them), NOT to this task's
+    // transcript — charging them here would put a healthy file into error
+    // backoff and point diagnostics at the wrong session.
+    if result.is_ok()
+        && let Err(e) = reconcile_flagged_sessions(token_db, &sink)
+    {
+        tracing::warn!(error = %e, "token-usage cross-session reconcile failed; flags retained for retry");
     }
     result
 }
@@ -606,13 +637,19 @@ fn process_file(
     let Some(mut extractor) = extractor_for_tool(&identity.tool) else {
         return Ok(());
     };
-    // A shrunken file was rewritten: restart from scratch. Entry-level dedup
-    // keeps re-extraction idempotent.
+    // A shrunken file was rewritten, and unreadable persisted state (corrupt
+    // or cross-version) means the cursor position is meaningless for the
+    // fresh extractor: both restart from scratch. Entry-level dedup keeps
+    // re-extraction idempotent, whereas continuing mid-file on default state
+    // would book the session's whole cumulative history as one delta.
     let mut offset = tracked.byte_offset;
     if offset > size {
         offset = 0;
-    } else if let Some(state) = tracked.state_json.as_deref() {
-        extractor.restore_state(state);
+    } else if let Some(state) = tracked.state_json.as_deref()
+        && !extractor.restore_state(state)
+    {
+        tracing::warn!(session_id = %identity.session_id, "unreadable extractor state; re-reading from the start");
+        offset = 0;
     }
 
     // Quiet skip: nothing changed since the last completed pass and the
@@ -629,7 +666,7 @@ fn process_file(
     reader.seek(SeekFrom::Start(offset))?;
 
     let cutoff = retention_cutoff_bucket(now);
-    let mut line = String::new();
+    let mut line: Vec<u8> = Vec::new();
     let mut reached_end = false;
     let mut interrupted = false;
     while !reached_end && !interrupted {
@@ -640,21 +677,36 @@ fn process_file(
                 interrupted = true;
                 break;
             }
-            match read_jsonl_line(&mut reader, &mut line)? {
-                JsonlLineState::Eof => {
+            match read_line_bytes(&mut reader, &mut line)? {
+                LineRead::Eof => {
                     reached_end = true;
                     break;
                 }
-                // A partial trailing line is still being appended; re-read it
-                // next pass (the cursor stays before it).
-                JsonlLineState::Partial => {
+                // A trailing line without a newline is usually a write in
+                // progress: leave the cursor before it. But if it already
+                // parses as complete JSON it will never grow a newline once
+                // the writer is gone, and the size/mtime snapshot would
+                // suppress every later pass — count it now (upstream counts
+                // unterminated final segments too).
+                LineRead::Partial(bytes) => {
+                    let text = String::from_utf8_lossy(&line);
+                    let trimmed = text.trim();
+                    if !trimmed.is_empty()
+                        && serde_json::from_str::<serde_json::Value>(trimmed).is_ok()
+                    {
+                        offset += bytes as u64;
+                        if extractor.wants_line(trimmed) {
+                            entries.extend(extractor.extract_line(trimmed));
+                        }
+                    }
                     reached_end = true;
                     break;
                 }
-                JsonlLineState::Complete(bytes) => {
+                LineRead::Complete(bytes) => {
                     offset += bytes as u64;
                     consumed += bytes;
-                    let trimmed = line.trim_end();
+                    let text = String::from_utf8_lossy(&line);
+                    let trimmed = text.trim_end();
                     if extractor.wants_line(trimmed) {
                         entries.extend(extractor.extract_line(trimmed));
                     }
@@ -712,14 +764,27 @@ fn emit_changed_buckets(
     // N+1 may exist in the metrics queue, that revision must never be
     // reused, or a crash before the fingerprint write would re-open the
     // equal-revision tie. A failed sink merely wastes a revision number.
+    // The revision is floored to wall-clock seconds so that losing the
+    // local revision state (flag off deletes the DB; the DB is rebuilt)
+    // cannot restart below revisions the server has already seen — a
+    // re-enabled backfill would otherwise re-emit at revision 1 and lose to
+    // every previously uploaded value under the highest-revision upsert.
+    let seq_floor = now_secs().max(0) as u64;
+    let changed: Vec<(crate::token_usage::db::ChangedBucket, u64)> = changed
+        .into_iter()
+        .map(|bucket| {
+            let next_seq = (bucket.emit_seq + 1).max(seq_floor);
+            (bucket, next_seq)
+        })
+        .collect();
     let reservations: Vec<(String, u32, u64)> = changed
         .iter()
-        .map(|bucket| (bucket.model.clone(), bucket.bucket_ts, bucket.emit_seq + 1))
+        .map(|(bucket, next_seq)| (bucket.model.clone(), bucket.bucket_ts, *next_seq))
         .collect();
     token_db.reserve_emit_seqs(&identity.session_id, &reservations)?;
     let repo_url = resolve_repo_url();
     let mut events = Vec::with_capacity(changed.len());
-    for bucket in &changed {
+    for (bucket, next_seq) in &changed {
         let aggregate = &bucket.aggregate;
         let values = TokenUsageValues::new()
             .bucket_ts(bucket.bucket_ts as u64)
@@ -733,7 +798,7 @@ fn emit_changed_buckets(
             .message_count(aggregate.message_count)
             // Strictly increasing per bucket: the server keeps the highest
             // revision, so same-second re-emissions cannot tie.
-            .emitted_seq(bucket.emit_seq + 1);
+            .emitted_seq(*next_seq);
         let mut attrs = EventAttributes::with_version(env!("CARGO_PKG_VERSION"))
             .session_id(identity.session_id.clone())
             .external_session_id(identity.external_session_id.clone())
@@ -746,13 +811,13 @@ fn emit_changed_buckets(
     }
     sink(&events)?;
     let now_ts = now_secs();
-    for bucket in changed {
+    for (bucket, next_seq) in changed {
         token_db.mark_emitted(
             &identity.session_id,
             &bucket.model,
             bucket.bucket_ts,
             &bucket.aggregate.fingerprint(),
-            bucket.emit_seq + 1,
+            next_seq,
             now_ts,
         )?;
     }
@@ -866,7 +931,11 @@ mod tests {
             assert_eq!(event.event_id, 9);
             assert_eq!(value_u64(event, token_usage_pos::INPUT_TOKENS), Some(100));
             assert_eq!(value_u64(event, token_usage_pos::MESSAGE_COUNT), Some(1));
-            assert_eq!(value_u64(event, token_usage_pos::EMITTED_SEQ), Some(1));
+            // Revisions are floored to wall-clock seconds so a rebuilt DB
+            // can never restart below previously uploaded revisions.
+            assert!(
+                value_u64(event, token_usage_pos::EMITTED_SEQ).unwrap() >= now_secs() as u64 - 60
+            );
             assert!(value_u64(event, token_usage_pos::EST_COST_MICRO_USD).unwrap() > 0);
             let attrs = EventAttributes::from_sparse(&event.attrs);
             assert_eq!(attrs.session_id, Some(Some("s_test".to_string())));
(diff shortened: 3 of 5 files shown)


src/daemon/stream_worker.rs
```diff
@@ -984,7 +989,17 @@ impl StreamWorker {
...
+                // the freshly ingested transcript bytes.
+                if task.stream_kind == "transcript"
+                    && let Some(token_usage) = &self.token_usage_worker
+                {
+                    token_usage.notify_stream_processed(
+                        &task.session_id,
+                        &task.tool,
+                        &task.canonical_path,
+                    );
+                }
```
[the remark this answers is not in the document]
svarlamov-git-ai: Good catch — fixed. The notification is now a shared `notify_token_usage` helper called from both success arms (`process_next_task` and `drain_immediate_tasks`), so transcripts processed inside the drain barrier ping the token-usage worker before the phase 2b token-usage drain runs.

src/daemon/token_usage_worker.rs
```diff
@@ -0,0 +1,878 @@
...
+            .await;
+            match result {
+                Ok(Ok(())) => {}
+                Ok(Err(e)) => {
+                    tracing::warn!(error = %e, session_id = %task.session_id, "token-usage processing failed");
+                    let _ = self.token_db.record_error(
+                        &task.session_id,
+                        &task.stream_path,
+                        &e.to_string(),
+                    );
```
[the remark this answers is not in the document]
svarlamov-git-ai: Fixed — error recording moved into `process_task_blocking`, keyed by the resolved rollup session id (`ctx.session_id`) that `ensure_file` used, with a regression test asserting a subagent transcript failure increments `processing_errors` on the parent-keyed row.

src/daemon/token_usage_worker.rs
```diff
@@ -0,0 +1,915 @@
...
+    /// backfills history for sessions tracked before this feature ran.
+    fn enqueue_sweep_tasks(&mut self) {
+        let streams = match self.streams_db.all_streams() {
+            Ok(streams) => streams,
+            Err(e) => {
+                tracing::warn!(error = %e, "token-usage sweep: failed to list streams");
+                return;
+            }
+        };
+        for stream in streams {
```
svarlamov-git-ai: **[review finding 2/15 — daemon resources, high] Every 30-minute sweep does O(all sessions ever) work before the quiet-skip can bail.**

`enqueue_sweep_tasks` enumerates every transcript row the streams DB has ever recorded (`delete_stream` has no production callers, so the table grows monotonically) and stats each file on the async runtime thread. Then, for every enqueued task, `process_task_blocking` pays a streams-DB `get_stream`, `EmissionContext::from_stream` (git-repo discovery walk + gix-config parse of system/global/repo config), a token-DB `ensure_file` INSERT-OR-IGNORE, and `fs::metadata` — all before `process_file`'s size/mtime quiet-skip is consulted. A workstation with thousands of accumulated sessions pays thousands of spawn_blocking tasks, SQLite round trips, and repo/config parses every 30 minutes forever, even fully idle. Aggravating: the sweep ticker keeps tokio's default `MissedTickBehavior::Burst`, so intervals missed during a long backfill or laptop suspend replay as back-to-back full sweeps + prunes on wake. Cheap fixes: hoist the size/mtime staleness check into `enqueue_sweep_tasks` (compare against the token DB snapshot), resolve/cache repo_url only after a file is known changed, and `set_missed_tick_behavior(Delay)`.

src/daemon/token_usage_worker.rs
```diff
@@ -0,0 +1,915 @@
...
+            .message_count(aggregate.message_count);
+        let mut attrs = EventAttributes::with_version(env!("CARGO_PKG_VERSION"))
+            .session_id(ctx.session_id.clone())
+            .external_session_id(ctx.external_session_id.clone())
+            .tool(&ctx.tool)
+            .model(model);
+        if let Some(url) = &ctx.repo_url {
+            attrs = attrs.repo_url(url.clone());
+        }
+        events.push(MetricEvent::new(&values, attrs.to_sparse()));
```
svarlamov-git-ai, line 833: **[review finding 4/15 — correctness, medium] Same-second re-emissions of a bucket tie on event_ts and a stale value can win permanently.**

The spec relies on the server upserting `(session_id, model, bucket_ts)` with 'newest event_ts winning', but `MetricEvent::new` stamps u32 SECONDS and the wire batch carries no sequence. Two emissions of the same bucket within one second are common (two files of one rollup session processed back-to-back in a sweep; pre/post tool-use passes), and the delivery order actively inverts emission order: `RETRYABLE_METRIC_IDS_SQL` orders `id DESC`, so within one upload flush the correction precedes the stale event, and a failed batch is retried minutes after later batches succeeded. With equal event_ts no server tie-break can be right for both orderings — and once `mark_emitted` stores the corrected fingerprint the client never re-sends, so a tie-broken-wrong bucket is permanently wrong (including the emptied-bucket zero correction). Fix: carry a monotonic per-bucket revision in the event (e.g., a counter in bucket_state bumped on each mark_emitted) and upsert on highest revision.

src/daemon/token_usage_worker.rs
```diff
@@ -0,0 +1,915 @@
...
+                }
+                Some(request) = self.drain_rx.recv() => {
+                    self.handle_drain(request).await;
+                }
+            }
+        }
+        tracing::info!("token-usage worker shutdown complete");
+    }
+
+    async fn handle_drain(&mut self, request: DrainRequest) {
```
svarlamov-git-ai, line 271: **[review finding 5/15 — regression, medium] The await drain barrier processes the entire backfill queue, so `git-ai await` can time out and skip the telemetry flush that used to run.**

`handle_drain` calls `process_queued()` on the whole queue — which after the startup sweep contains every historical transcript backfilling from byte 0. On first daemon start with this feature enabled (debug default: on), a commit-triggered await hits Phase 2b's `timeout(remaining, worker.drain())`, which must finish reading the user's full transcript history; on timeout `result.timed_out = true` and Phase 3 (`if !result.timed_out && telemetry_worker`) — the telemetry flush that main's await always ran — is skipped. Consider bounding the drain to notification-derived work (separate sweep/backfill queue from the notify queue), or draining only tasks enqueued before the barrier.

src/daemon/token_usage_worker.rs
```diff
@@ -0,0 +1,915 @@
...
+        }
+
+        loop {
+            self.process_queued().await;
+            if self.shutdown_flag.load(Ordering::Relaxed) {
+                break;
+            }
+            tokio::select! {
+                _ = self.shutdown_notify.notified() => {
+                    self.shutdown_flag.store(true, Ordering::Relaxed);
```
svarlamov-git-ai, line 238: **[review finding 6/15 — regression, medium] shutdown_flag is only ever set here, so every shutdown check in the blocking path is dead code and shutdown is starved during a backfill.**

The run loop awaits `process_queued()` to completion before re-entering `select!`, and this arm is the only `store(true)`. So while the queue drains (startup backfill over months of transcripts), `request_shutdown`'s `notify_one` just parks a permit: the per-task check in `process_queued` and the per-line check in `process_file` always read false, and the worker chews the entire backlog before observing shutdown. During a daemon restart/self-update, the old process's runtime-drop also blocks on the in-flight spawn_blocking pass. The interruption plumbing (flag threaded through `process_task_blocking`/`process_file`, the `interrupted` mid-batch path and its comments) describes behavior that cannot occur. Either process one task per select iteration (the stream worker's loop shape) so shutdown wins between tasks, or set the shared flag directly from `request_shutdown`.

src/daemon/token_usage_worker.rs
```diff
@@ -0,0 +1,915 @@
...
+/// emit those whose aggregate fingerprint differs from the last emitted one,
+/// marking them emitted only after the sink accepted the events.
+fn emit_changed_buckets(
+    token_db: &TokenUsageDatabase,
+    ctx: &EmissionContext,
+    sink: impl Fn(&[MetricEvent]) -> Result<(), GitAiError>,
+) -> Result<(), GitAiError> {
+    let mut events = Vec::new();
+    let mut emitted = Vec::new();
+    for (model, bucket) in &token_db.session_buckets(&ctx.session_id)? {
```
svarlamov-git-ai: **[review finding 8/15 — daemon resources, medium] Every notification reconciles every bucket the session has ever had — cost grows with session age, not with what changed.**

`emit_changed_buckets` runs `session_buckets` (UNION scan) and then two mutex-locked point queries (`aggregate_bucket` + `emitted_fingerprint`) per `(model, bucket)` — on every stream-worker notification, i.e. roughly every transcript append. A long-lived rollup session (parent + subagents, active for days within the 90-day retention) accumulates hundreds-to-thousands of buckets, so each append re-runs ~2N queries on the blocking pool, sustained all day. The batch just committed knows exactly which buckets it touched: reconcile only those on the notify path and reserve full-session reconciliation for sweeps/error recovery — or at least collapse the loop into one `GROUP BY model, bucket_ts` aggregate plus one bucket_state fetch (two statements total).

src/daemon/token_usage_worker.rs
```diff
@@ -0,0 +1,915 @@
...
+                    &token_db,
+                    &telemetry,
+                    &task_clone,
+                    &shutdown_flag,
+                )
+            })
+            .await;
+            match result {
+                Ok(Ok(())) => {}
+                Ok(Err(e)) => {
```
svarlamov-git-ai: **[review finding 9/15 — daemon resources/correctness, medium] No retry backoff: a permanently failing transcript is re-read in full on every notification and every sweep, forever.**

`BufRead::read_line` returns InvalidData on any non-UTF-8 byte, so one corrupt byte in a transcript makes `process_file` error on every pass: the cursor stays pinned before the bad line, the size/mtime snapshot is never written (no quiet-skip), and every stream notification plus every 30-minute sweep re-opens the file and re-reads up to the poison byte — while all usage after it is never counted. `record_error` increments `processing_errors`, but nothing ever reads it (no backoff, no gating in `enqueue_sweep_tasks`), so the error columns are write-only state. Contrast the stream worker's retry_count/next_retry_at backoff. Suggest consulting `processing_errors` to back off or skip persistently failing files (and consider lossy handling of undecodable lines so the cursor can advance).

tests/integration/token_usage.rs
```diff
@@ -0,0 +1,336 @@
...
+    fs::write(&file_path, "const x = 1;\n").unwrap();
+    checkpoint_with_transcript(
+        &repo,
+        "claude",
+        "sess-token-claude",
+        &transcript_path,
+        &file_path,
+        "const x = 1;\nconst y = 2;\n",
+    );
+    repo.sync_daemon();
```
svarlamov-git-ai: **[review finding 10/15 — tests, medium] The e2e assertions race the async pipeline: sync_daemon() provides no barrier over stream-worker -> token-usage-worker -> metrics DB.**

`TestRepo::sync_daemon` resolves to `ControlRequest::SyncFamily`, whose handler (`sync_family`, src/daemon.rs) waits only for checkpoint admission, trace-ingest, and family-sequencer drains. The transcript notification is a fire-and-forget unbounded-channel send at checkpoint admission (`notify_checkpoint_stream`), and transcript processing, the token worker's queue hop + spawn_blocking pass, and `persist_metrics_blocking` all complete afterwards. The token-usage drain (Phase 2b) exists only in `await_completion` (`git-ai await`), which the daemon_mode.rs test uses but these tests do not. On a loaded CI runner, `assert_eq!(events.len(), ..)` right after sync_daemon flakes, and the negative assertions (unchanged bucket not re-emitted; zero bucket emitted exactly once) pass vacuously whenever the worker simply has not run yet. Use an await-based barrier (or the polling pattern in tests/integration/opencode.rs) before asserting.

src/daemon/token_usage_worker.rs
```diff
@@ -0,0 +1,915 @@
...
+        )?;
+    }
+
+    if interrupted {
+        // Shutdown mid-pass: entries and cursor are committed, and the
+        // size/mtime snapshot stays stale, so the next pass resumes and
+        // reconciles emission.
+        return Ok(());
+    }
+    emit_changed_buckets(token_db, ctx, sink)?;
```
svarlamov-git-ai: **[review finding 11/15 — efficiency, medium] Backfill inserts, emits, and uploads buckets older than the 90-day retention, only for the next sweep to delete them.**

Neither extraction (`upsert_entry` accepts any bucket_ts) nor emission applies the ENTRY_RETENTION_DAYS cutoff, so a first-run backfill over years of transcripts inserts hundreds of thousands of usage_entries rows and emits/uploads TokenUsage events for buckets far older than 90 days — all persisted to the metrics DB and sent over the network — which `prune_buckets_before` deletes up to 30 minutes later. Pure wasted CPU, disk, and upload bandwidth plus a transiently unbounded usage_entries table. Dropping entries older than the retention cutoff in process_file/commit_batch (and skipping emission for pre-cutoff buckets) avoids all of it.

src/daemon/token_usage_worker.rs
```diff
@@ -0,0 +1,915 @@
...
+            Some(parent_ext) => (
+                generate_session_id(parent_ext, &stream.tool),
+                parent_ext.clone(),
+            ),
+            None => (
+                stream.session_id.clone(),
+                stream.external_session_id.clone(),
+            ),
+        };
+        let repo_url = stream
```
svarlamov-git-ai: **[review finding 13/15 — privacy gate, low-medium] Sweep-backfilled sessions without a resolvable repo_url bypass the exclude gate (fail-open).**

`EmissionContext::from_stream` resolves repo_url only from `stream.repo_work_dir` — no `agent.infer_cwd` fallback, unlike the SessionEvent path in stream_worker.rs (task.repo_work_dir -> stream row -> infer_cwd). Rows with NULL repo_work_dir (pre-v3-migration rows, sessions the stream worker has not successfully processed, deleted working dirs) emit TokenUsage events with no repo_url attr, and `is_allowed_repository_with_remotes(None)` skips the exclusion check under an exclude-only config (pinned by test_no_remotes_allowed_when_only_excludes) — so an excluded repo's token usage uploads. The startup/30-min sweeps process exactly these rows independently of the stream worker, and once emitted the fingerprint prevents re-emission. The fail-open semantic matches pre-existing SessionEvent behavior, but the missing infer_cwd fallback is a real gap: add the same fallback (or persist resolution) in from_stream.

src/daemon.rs
```diff
@@ -9211,16 +9244,43 @@ pub(crate) async fn run_daemon(config: DaemonConfig) -> Result<DaemonExitAction,
             Ok(streams_db) => {
                 let streams_db = std::sync::Arc::new(streams_db);
                 let shutdown_notify = Arc::new(tokio::sync::Notify::new());
+                // Each worker needs its own shutdown Notify: request_shutdown
+                // uses notify_one, which wakes exactly one waiter.
+                let token_usage_shutdown_notify = Arc::new(tokio::sync::Notify::new());
+                // The token-usage worker is fed by stream-worker completions,
+                // so it lives inside the transcript_streaming gate. Its own
+                // token_usage_metrics flag is re-read per trigger.
+                let token_usage_handle = match crate::token_usage::db::TokenUsageDatabase::open(
```
svarlamov-git-ai: **[review finding 14/15 — feature-flag hygiene, low-medium] With token_usage_metrics off (release default), the DB is still created/migrated, the worker/ticker still run, and disabling the flag stops the retention prune.**

Three leaks around the per-trigger flag design: (1) every release user with transcript_streaming on gets `~/.git-ai/internal/token-usage-db` (+WAL) created and migrated synchronously at each daemon start, plus the worker task and 30-minute ticker, purely to be discarded at the flag check — the DB open could be deferred until the flag is first seen enabled; (2) `token_usage_enabled()` calls `Config::fresh()` — a full config-file read + parse + env scan — on every stream-worker notification, i.e. several times per second during active streaming even when the flag is off (a short-TTL cache like the existing `Config::fresh_author_cached` pattern fits); (3) `prune_old_entries` only runs when the flag is enabled, so data collected while enabled is retained indefinitely after the flag is turned off — the prune should run regardless.

src/daemon/token_usage_worker.rs
svarlamov-git-ai: Fixed — sweeps now prefilter in the enumeration loop: one `fs::metadata` per session compared against the token DB's size/mtime snapshot (loaded once as a map), so settled sessions never reach `spawn_blocking`, DB reads, or repo discovery. Repo discovery itself (gix-config parse) is additionally deferred inside the pass until events are actually emitted. The ticker uses `MissedTickBehavior::Skip` so suspend/resume triggers one sweep, not a replay of every missed tick.

src/daemon/token_usage_worker.rs
svarlamov-git-ai: Fixed as suggested — `bucket_state` now carries a strictly increasing per-bucket `emit_seq`, bumped on every `mark_emitted`, and events carry it at `token_usage_pos::EMITTED_SEQ`. The server upserts keeping the highest revision, so same-second re-emissions and inverted upload/retry order can no longer make a stale value win permanently. The spec documents the contract.

src/daemon/token_usage_worker.rs
svarlamov-git-ai: Fixed — notifications and sweep backfill are separate queues, and the drain barrier processes only notification-driven work (a notification for a file already queued for backfill promotes it). A first-start historical backfill keeps running in the background loop and can no longer starve `git-ai await` past its deadline or skip the Phase 3 telemetry flush. Covered by `notifications_promote_queued_sweep_tasks_and_drain_skips_backfill`.

src/daemon/token_usage_worker.rs
svarlamov-git-ai: Fixed — `process_task` now selects on the shutdown `Notify` around the in-flight `spawn_blocking` pass: shutdown sets the flag (which the pass polls per line and per backpressure wait) and the loop breaks as soon as the pass returns, so daemon exit is bounded by one batch boundary instead of a whole backfill.

src/daemon/token_usage_worker.rs
svarlamov-git-ai: Fixed — reconciliation is now a single grouped query (`changed_buckets`): one aggregate pass over the session's entries plus its bucket_state rows, under one connection lock, returning only fingerprint mismatches. Per-pass cost is one indexed scan instead of ~2N individual queries.

src/daemon/token_usage_worker.rs
svarlamov-git-ai: Fixed — `tracked_files` gained `last_error_at` (schema v2); a file whose last pass failed backs off 5s/30s/5m/30m (by consecutive `processing_errors`) before it is read again, and a successful commit clears the state. Sweeps still enqueue such files but the pass returns before any file read. Covered by `error_backoff_suppresses_immediate_retries`.

tests/integration/token_usage.rs
svarlamov-git-ai: Fixed — the e2e tests now run behind `git-ai await` (which reaches `await_completion`: stream drain → token-usage drain → telemetry flush, all before it returns) in addition to `sync_daemon`, so the assertions — including the negative ones — are deterministic rather than racing the pipeline. Fixtures also moved to recent timestamps so they stay inside the new retention window.

src/daemon/token_usage_worker.rs
svarlamov-git-ai: Fixed — the retention cutoff is applied at extraction: `commit_batch` drops entries whose bucket falls before it, so first-run backfill never inserts or uploads pre-cutoff history for the next prune to delete. Covered by `entries_before_the_retention_cutoff_are_dropped_at_insert` and `entries_older_than_retention_are_not_emitted`.

src/daemon/token_usage_worker.rs
svarlamov-git-ai: Fixed — repo_url resolution now has the same `infer_cwd` fallback as the SessionEvent path (resolved lazily, only when events are actually emitted, and persisted back to the streams row via `update_repo_work_dir`), so sweep-backfilled sessions carry a repo_url whenever one is resolvable and the exclude gate can act on it. The residual fail-open for genuinely unresolvable sessions matches the pre-existing SessionEvent semantics, as noted.

src/daemon.rs
svarlamov-git-ai: Fixed — went with startup gating rather than per-trigger checks, mirroring `transcript_streaming`: when `token_usage_metrics` is off, the worker and ticker are not spawned, the database is never created/migrated, `Config::fresh()` is never read (the per-trigger helper is gone entirely), and a previously created database (+WAL/SHM) is deleted at startup — so disabling the flag also removes all collected data rather than retaining it un-pruned. Enabling requires a daemon restart, which matches the existing streaming flag's semantics. Covered by the `disabled_flag_spawns_nothing_and_deletes_collected_data` e2e.

src/daemon/token_usage_worker.rs
```diff
@@ -0,0 +1,1308 @@
...
+                _ = sweep_ticker.tick() => {
+                    self.enqueue_sweep_tasks();
+                }
+                Some(task) = self.notify_rx.recv() => {
+                    self.enqueue_notify(task);
+                }
+                Some(request) = self.drain_rx.recv() => {
+                    self.handle_drain(request).await;
+                }
+            }
```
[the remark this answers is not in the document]
svarlamov-git-ai: You're right — the `continue` before the select meant drain requests waited for the whole sweep queue. The loop now uses the stream worker's pattern: ready work is a `_ = async {}, if has_ready_task` select arm, so `drain_rx`, `notify_rx`, the ticker, and shutdown are all serviced between tasks while backfill is queued; pending notifications are ingested before each task is chosen so fresh work still promotes over backfill. A drain during backfill now completes after only the notification-driven queue.

src/daemon/token_usage_worker.rs
```diff
@@ -0,0 +1,1315 @@
...
+            {
+                continue;
+            }
+            self.enqueue_sweep(TokenUsageTask {
+                session_id: stream.session_id,
+                tool: stream.tool,
+                stream_path: stream.stream_path,
+            });
+        }
+    }
```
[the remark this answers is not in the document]
svarlamov-git-ai: Fixed — the enumeration is now a pure `sweep_candidates` function run inside `spawn_blocking` (both SQLite scans and the per-file stats); only the cheap enqueue happens on the async task, consistent with `process_task` and the module's I/O invariant.

src/daemon/token_usage_worker.rs
```diff
@@ -0,0 +1,1348 @@
...
+        // Shutdown mid-pass: entries and cursor are committed, and the
+        // size/mtime snapshot stays stale, so the next pass resumes and
+        // reconciles emission.
+        return Ok(());
+    }
+    // Cross-session replacements changed other sessions' buckets: invalidate
+    // their snapshots so their next pass re-reconciles with their own
+    // attributes.
+    for session in &foreign_sessions {
+        token_db.invalidate_session_files(session)?;
```
svarlamov-git-ai: **[follow-up review — correctness, medium] Cross-session invalidation is best-effort in two ways that can leave permanent server-side over-counts after an entry moves sessions.**

(a) **Not crash/interrupt-durable.** `commit_batch` durably moves the row (and advances the cursor) per batch, but the previous owners accumulate in the in-memory `foreign_sessions` vec and `invalidate_session_files` runs only after the whole-file loop — and is skipped entirely by the `if interrupted { return Ok(()) }` early return. A shutdown mid-pass or a crash between a batch commit and the invalidation loop loses the invalidation forever: the moved line is behind the committed cursor, is never re-read, so the move is never re-reported, session A's snapshot stays settled, and its emptied bucket never re-emits — contradicting the module's own 'a crash at any point is healed by the next pass' invariant.

(b) **A no-op when the previous session's transcripts are gone.** Re-reconciliation of A only ever happens via a pass over one of A's files: `sweep_candidates` skips rows whose `fs::metadata` errors, and notifications only fire for successfully processed files. If A's transcript was deleted (Claude Code prunes after ~30 days; users clear ~/.claude/projects), the invalidation never fires and the server permanently keeps A's pre-move totals while B also carries the moved entry. Realistic trigger: a subagent file ingested before its parent linkage was known (sidechain entries under their own rollup session) whose message is later replaced by the parent's non-sidechain entry under a different rollup session — non-sidechain wins regardless of totals.

Both close with the same deeper fix: reconciliation needs no file read (`changed_buckets` + emit are DB-only), so reconcile foreign sessions directly in the same pass (identity/attrs resolvable from the streams DB), and/or record the invalidation durably inside `commit_batch`'s transaction instead of an in-memory vec.

src/daemon/token_usage_worker.rs
```diff
@@ -0,0 +1,1348 @@
...
+            .session_id(identity.session_id.clone())
+            .external_session_id(identity.external_session_id.clone())
+            .tool(&identity.tool)
+            .model(&bucket.model);
+        if let Some(url) = &repo_url {
+            attrs = attrs.repo_url(url.clone());
+        }
+        events.push(MetricEvent::new(&values, attrs.to_sparse()));
+    }
+    sink(&events)?;
```
svarlamov-git-ai, line 835: **[follow-up review — correctness, low] The emission revision is consumed at sink time but only reserved after it, re-opening an equal-seq tie in the crash window.**

`sink(&events)` durably persists events carrying `emit_seq + 1` into the metrics DB, but `mark_emitted` stores that revision only afterwards. Crash between the two (or `mark_emitted` erroring after a successful sink): the next pass re-reads any newly appended lines first, then reconciles against the stale `bucket_state` and emits the *changed* aggregate under the *same* revision N+1 — while the pre-crash values at N+1 are already queued/uploaded. A server keeping the highest `emitted_seq` treats equal as not-newer, so whichever payload lands first wins and the bucket can pin to the stale value until some later change bumps it to N+2 — never, for a finished session. This is exactly the tie the revision was added to eliminate.

Fix: reserve the revision before handing events to the sink — persist `emit_seq = N+1` (leaving `emitted_fingerprint` untouched until after the sink succeeds). A failed sink then just wastes a revision number (harmless for a strictly-increasing u64), and a crash after reserve-but-before-sink re-emits at N+2, a gap the highest-wins server ignores.

tests/integration/token_usage.rs
```diff
@@ -0,0 +1,411 @@
...
+fn now_secs() -> i64 {
+    std::time::SystemTime::now()
+        .duration_since(std::time::UNIX_EPOCH)
+        .unwrap()
+        .as_secs() as i64
+}
+
+/// Recent RFC3339 timestamps (start of today + offset) so fixture entries
+/// fall inside the retention window.
+fn recent_ts(minute: u32, second: u32) -> String {
```
svarlamov-git-ai, line 52: **[follow-up review — test determinism, low] recent_ts anchors to the start of *today* UTC and recomputes the base per call — two small daily flake windows.**

(a) For the first ~61 seconds after 00:00 UTC, `recent_ts(1, 0)` is in the *future*, and the codex EOF flush requires `now_ms - first_ts_ms > 1000`: the worker unit test `forked_codex_single_turn_is_flushed_at_end_of_file` (which duplicates this helper in src/daemon/token_usage_worker.rs) then gets zero events from its single pass and fails — its comment 'recent_ts is in the past (>1s)' is false in that window. (b) `base = now_secs() - now_secs() % 86_400` is recomputed on every call, so a test whose `recent_ts` fixture write and `bucket_of` assertion straddle midnight UTC compares buckets 86,400s apart (the e2e tests are most exposed — the write-to-assert gap spans daemon sync plus a 60s await).

Cheap deterministic fix for both duplicated helper pairs: compute the anchor once (e.g. `LazyLock` holding `now_secs() - 7_200` floored to a 300s boundary) and share it between `recent_ts` and `bucket_of` — timestamps are always safely in the past and a day rollover cannot split bases, while staying far inside the 90-day retention window.

src/daemon.rs
```diff
@@ -9211,16 +9244,54 @@ pub(crate) async fn run_daemon(config: DaemonConfig) -> Result<DaemonExitAction,
...
+                                token_usage_shutdown_notify.clone(),
+                            ))
+                        }
+                        Err(e) => {
+                            tracing::error!(error = %e, "failed to open token-usage database, token-usage worker not started");
+                            None
+                        }
+                    }
+                } else {
+                    crate::token_usage::db::TokenUsageDatabase::remove_database_files(
```
svarlamov-git-ai: **[follow-up review — flag hygiene, low] The flag-off data deletion only runs inside the transcript_streaming gate, so 'no collected data is retained' does not hold when transcript_streaming is off.**

`remove_database_files` is reachable only when `transcript_streaming` is enabled AND `StreamsDatabase::open` succeeded — it is nested in that match arm. A user who collected token-usage data and later disables transcript_streaming (or whose transcripts-db fails to open) keeps `~/.git-ai/internal/token-usage-db` on disk indefinitely with `token_usage_metrics` off, contradicting the spec's 'a previously created one is deleted — no collected data is retained' and the comment above this block. Hoist the flag-off deletion out of the transcript_streaming gate so it runs unconditionally at daemon startup when the flag is off.

src/daemon/token_usage_worker.rs
svarlamov-git-ai: Fixed with the suggested deeper approach, both halves: (a) `commit_batch` now durably flags the previous owner (`needs_reconcile`, set in the same transaction as the entry move), so the pending correction survives interrupts and crashes — the in-memory vec and the `invalidate_session_files` snapshot trick are gone. (b) Reconciliation of flagged sessions is DB-only: `reconcile_flagged_sessions` re-aggregates and emits with the identity stored on the tracked rows (`external_session_id` is persisted at `ensure_file` time), needing no file — it runs inline after every pass and from sweeps for crash recovery, so it fires even when the previous session's transcripts were deleted. Covered by `cross_session_replacement_reconciles_the_previous_owner_without_its_file` (which deletes the file first) and `failed_reconcile_sink_keeps_the_flag_for_retry`.

src/daemon/token_usage_worker.rs
svarlamov-git-ai: Fixed as suggested — `reserve_emit_seqs` persists `emit_seq = N+1` for every changed bucket *before* events are handed to the sink, leaving the fingerprint untouched (placeholder `''` on first contact) until after the sink succeeds. A crash in the old window now re-emits at N+2; a failed sink just wastes a revision number. Covered by `reserved_revisions_are_never_reused`.

tests/integration/token_usage.rs
svarlamov-git-ai: Fixed — fixtures now anchor to *yesterday's* UTC midnight via a process-wide `OnceLock` (`fixture_base`), so the base never shifts between fixture creation and assertion across a midnight boundary and fixture timestamps are never in the future. Applied to both the worker unit tests and the integration tests.

src/daemon.rs
svarlamov-git-ai: Fixed — the flag-off deletion is hoisted out of the `transcript_streaming` gate and runs unconditionally at daemon startup when `token_usage_metrics` is off, so 'no collected data is retained' holds even when streaming is also disabled.

src/daemon/token_usage_worker.rs
```diff
@@ -0,0 +1,1348 @@
...
+                Ok(tasks) => tasks,
+                Err(e) => {
+                    tracing::error!(error = %e, "token-usage sweep panicked");
+                    return;
+                }
+            };
+        for task in tasks {
+            self.enqueue_sweep(task);
+        }
+    }
```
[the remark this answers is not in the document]
svarlamov-git-ai: Fair for this PR standalone — the prune wiring previously landed one PR up the stack, which made this module's retention docs overstated here. Retention is one concern, so both halves now live in this PR: the sweep calls `prune_buckets_before(retention_cutoff_bucket(now))` in `spawn_blocking` (at startup and every tick) before enumerating, alongside the existing drop-at-insert cutoff.

src/daemon/token_usage_worker.rs
```diff
@@ -0,0 +1,1481 @@
...
+    token_db: &TokenUsageDatabase,
+    sink: &impl Fn(&[MetricEvent]) -> Result<(), GitAiError>,
+) -> Result<(), GitAiError> {
+    for (session_id, external_session_id, tool) in token_db.sessions_needing_reconcile()? {
+        let identity = SessionIdentity {
+            session_id: session_id.clone(),
+            external_session_id,
+            tool,
+        };
+        emit_changed_buckets(token_db, &identity, || None, sink)?;
```
svarlamov-git-ai: **[round-3 review — privacy gate, low-medium] DB-only reconcile corrections carry no repo_url, so they bypass the exclude gate that blocked the session's original events.**

The reconcile path emits with `|| None`, so these events reach `should_deliver_metric_event` with no repo_url attr, and `is_allowed_repository_with_remotes(None)` fails open under an exclude-only config. Sequence: repo X is excluded -> the session's normal emissions (which DO carry repo_url) are correctly dropped at upload -> a cross-session replacement flags the session -> the reconcile correction, carrying the session's remaining real aggregate values (session ids, tool, model, bucket, token counts), uploads because it lacks the attribute the gate keys on. The gate exists precisely so excluded repos' usage never reaches the server; this path sends it. Secondary: for non-excluded repos the correction also loses the repo_url attribute the original emission had (values-only upsert per the code comment, but worth confirming against server attr handling).

Fix options: persist the resolved repo_url (or repo_work_dir) into `tracked_files` alongside `external_session_id` so DB-only reconciliation carries the same attribute, or resolve via the streams DB inside the reconcile (both call sites already have `streams_db`; a session lookup does not need the transcript file).

svarlamov-git-ai: Fixed — the per-task resolver now persists the resolved repo_url on the session's tracked files (`update_session_repo_url`, never erasing a stored value on a transient resolution failure), and `reconcile_flagged_sessions` emits with the stored value from `sessions_needing_reconcile`. Corrections therefore carry the same repo gate attribute as the session's original events; when the originals never resolved a repo_url, corrections match them too. Covered by the extended `cross_session_replacement_reconciles_the_previous_owner_without_its_file` test, which asserts the correction's repo_url attr.

src/daemon/token_usage_worker.rs
```diff
@@ -0,0 +1,1499 @@
...
+    let mut interrupted = false;
+    while !reached_end && !interrupted {
+        let mut entries = Vec::new();
+        let mut consumed = 0usize;
+        loop {
+            if shutdown_flag.load(Ordering::Relaxed) {
+                interrupted = true;
+                break;
+            }
+            match read_jsonl_line(&mut reader, &mut line)? {
```
svarlamov-git-ai: **[Token-usage accuracy review — finding 1/15]** (stack-wide review vs upstream ccusage)

The incremental read layer diverges from upstream ccusage's byte-level reads in two silent-data-loss ways, with no test for either: (a) `read_jsonl_line` uses UTF-8-strict `BufRead::read_line`, so a single invalid-UTF-8 byte in a transcript makes every pass error at the same committed offset forever (error backoff caps at 30min retries, the cursor never advances, and ALL usage after that byte is never counted) whereas upstream (`fs::read` + `byte_lines` + `from_slice`) skips just the bad line; (b) a final line that is complete JSON but lacks a trailing newline is returned as `Partial`, then `update_file_metadata` records the full size, so if the file never grows again the quiet-skip (size+mtime match) suppresses every later pass and that entry's tokens are permanently lost — upstream's `byte_lines` yields and counts the unterminated final segment.

**Failure scenario:** Agent crashes mid-write leaving a truncated multi-byte character; the next append merges it into one invalid-UTF-8 line -> git-ai reports zero usage for everything after that point in the file, forever, with only a daemon log line. Or: agent killed between writing a usage JSON object and its newline -> that message's tokens are never reported. Suggested tests: transcript with an invalid-UTF-8 line followed by valid usage lines (assert later lines counted), and a transcript whose last usage line has no trailing newline (assert eventual counting or pin the intended loss).

src/daemon.rs
```diff
@@ -9199,6 +9232,17 @@ pub(crate) async fn run_daemon(config: DaemonConfig) -> Result<DaemonExitAction,
...
 
+    // With the token-usage flag off, previously collected data is deleted
+    // regardless of the streaming gate below (the spec's "no collected data
+    // is retained" holds even when transcript_streaming is also off).
+    let token_usage_db_path = config.internal_dir.join("token-usage-db");
+    let token_usage_enabled = config::Config::get()
+        .get_feature_flags()
+        .token_usage_metrics;
+    if !token_usage_enabled {
+        crate::token_usage::db::TokenUsageDatabase::remove_database_files(&token_usage_db_path);
```
svarlamov-git-ai, line 9259: **[Token-usage accuracy review — finding 3/15]** (stack-wide review vs upstream ccusage)

Feature-flag off->on permanently desyncs the server: turning the flag off deletes the token DB (including all per-bucket `emit_seq` state), and re-enabling backfills every bucket starting at emitted_seq=1 - but the documented server contract keeps the HIGHEST emitted_seq per bucket, so any bucket previously emitted at seq>=2 silently ignores the re-emitted (possibly corrected) values, while the local DB records the new fingerprint as emitted so the divergence is never healed.

**Failure scenario:** Common path: bucket emitted at seq 1 then re-emitted at seq 2 after an append (the normal case per `appended_usage_reemits_the_bucket_with_bumped_revision`); user toggles the flag off then on; the recomputed aggregate differs (new lines, replacement correction, pricing change) and re-emits at seq 1 -> server keeps the stale seq-2 value forever. No epoch/generation exists in the wire format to disambiguate, and no test covers the off->on transition (only off is tested). Fix directions: persist a monotonic epoch outside the deleted DB, derive the initial seq from wall time, or spec the server to treat a new epoch as authoritative.

tests/integration/token_usage.rs
```diff
@@ -0,0 +1,418 @@
+//! End-to-end tests for TokenUsage metric events (event id 9): checkpoint ->
```
svarlamov-git-ai, line 1: **[Token-usage accuracy review — finding 5/15]** (stack-wide review vs upstream ccusage)

There is no realistic-transcript corpus/golden test anywhere in the suite: every token count in every unit, worker, and e2e test comes from minimal hand-rolled JSON built from the parser's own assumptions, while both extractors silently drop anything they cannot parse AND the committed cursor makes such losses permanent (unlike upstream, which re-reads whole files so a parser fix retroactively repairs history) - the repo already ships real captured transcripts (tests/fixtures/example-claude-code.jsonl with 23 usage lines including genuine streaming re-emits of duplicate message ids, cache_creation breakdowns; codex-session-simple.jsonl with real session_meta/turn_context shapes) that no token-usage test consumes.

**Failure scenario:** Any serde-shape regression or upstream format drift (a field made strict, the `"usage":{` marker broken by a Claude Code serialization change, a Codex field type change) makes extract_line return empty for real transcripts: production token reporting drops to zero with every test green and no drop counter or log to notice. Highest-leverage fix for the stated goal: an extractor-level golden test (real fixture lines -> insta snapshot of per-model bucket sums) over the existing fixtures, plus a parameterized conformance harness over `extractor_for_tool` (prefilter-superset property, state roundtrip == single pass, entry-key stability) so future extractors inherit the invariants.

src/daemon/token_usage_worker.rs
```diff
@@ -0,0 +1,1499 @@
...
+    drain_rx: tokio::sync::mpsc::UnboundedReceiver<DrainRequest>,
+    /// Notification-driven work: drained by the await barrier.
+    notify_queue: VecDeque<TokenUsageTask>,
+    /// Sweep/backfill work: processed only in the background loop.
+    sweep_queue: VecDeque<TokenUsageTask>,
+    queued: HashSet<TokenUsageTask>,
+}
+
+impl TokenUsageWorker {
+    async fn run(mut self) {
```
svarlamov-git-ai, line 216: **[Token-usage accuracy review — finding 9/15]** (stack-wide review vs upstream ccusage)

The durability and identity machinery the spec leans on is tested only below its seams: the run() loop (startup sweep, 30-min ticker, prune ordering, crash-recovery reconcile_flagged) is never driven with real work at any level and `git-ai await` deliberately excludes the sweep queue so no barrier exists to ever test it deterministically; and cross-session dedup / needs_reconcile corrections / subagent-to-parent rollup are never exercised through the real daemon (all e2e tests are single-session; identity flows only through hand-built StreamRecords in unit tests) - the exact seam that already produced a bug in this stack (d90bcbf31: errors recorded under the non-rollup session id).

**Failure scenario:** A select-arm or sweep-filter regression silently kills backfill/healing/prune (users lose all usage written while the daemon was down; crash-left corrections never emit; unbounded DB growth) with green CI; or a change in how the stream worker populates external_parent_session_id re-counts the full conversation history on every `claude --resume` while every unit test holding both sides of the seam constant stays green. Needed: a test-only drain/barrier covering the sweep queue, a daemon-restart backfill e2e, and a two-session resume-copy e2e plus a subagent-transcript e2e asserting rollup attribution and dedup.

src/daemon/token_usage_worker.rs
```diff
@@ -0,0 +1,1499 @@
...
+            .session_id(identity.session_id.clone())
+            .external_session_id(identity.external_session_id.clone())
+            .tool(&identity.tool)
+            .model(&bucket.model);
+        if let Some(url) = &repo_url {
+            attrs = attrs.repo_url(url.clone());
+        }
+        events.push(MetricEvent::new(&values, attrs.to_sparse()));
+    }
+    sink(&events)?;
```
svarlamov-git-ai, line 835: **[Token-usage accuracy review — finding 12/15]** (stack-wide review vs upstream ccusage)

The exactly-once design treats telemetry-queue acceptance as final (`sink(&events)?` then `mark_emitted` records the fingerprint), but the metrics upload layer permanently abandons any record after 6 failed attempts (~20.6h of cumulative backoff; `attempts < 6` in every dequeue query and nothing ever resets attempts) - so corrections and first emissions queued while the machine is offline for a day are dropped by the uploader while the token DB believes they were delivered, and the fingerprint match ensures they are never regenerated.

**Failure scenario:** Dev codes offline for a long weekend with the daemon running; a downward correction (bucket lowered from 50,000 to 200 cache-read tokens) exhausts its 6 attempts -> server keeps the inflated value forever. `failed_sink_leaves_bucket_unmarked_for_retry` covers queue-refusal only; the queue-accepted-then-abandoned seam has no test. Options: exclude TokenUsage events from the attempts cap, or feed upload-abandonment back into bucket_state (clear the fingerprint) so the next pass re-emits.

src/daemon/token_usage_worker.rs
```diff
@@ -0,0 +1,1499 @@
...
+    let size = metadata.len();
+    let modified = modified_secs(&metadata);
+
+    let Some(mut extractor) = extractor_for_tool(&identity.tool) else {
+        return Ok(());
+    };
+    // A shrunken file was rewritten: restart from scratch. Entry-level dedup
+    // keeps re-extraction idempotent.
+    let mut offset = tracked.byte_offset;
+    if offset > size {
```
svarlamov-git-ai, line 665: **[Token-usage accuracy review — finding 13/15]** (stack-wide review vs upstream ccusage)

Transcript rewrite/truncation handling has untested blind spots: only a strict shrink resets the cursor (and even that branch has no test); a file rewritten with different content of size >= the old cursor is read from a mid-content offset with STALE extractor state (codex cumulative subtraction then yields garbage or saturated-to-zero deltas, and session_meta/turn_context lines before the cursor are never seen); entries extracted from replaced content can never be invalidated because usage_entries has no per-file ownership; and a same-size rewrite within the same mtime second is quiet-skipped forever.

**Failure scenario:** A rollout file is rotated/recreated at the same path (or a session id reuses a path): the worker seeks mid-new-content, codex deltas are computed against the old file's prev_totals, and the old content's entries remain counted on top of the new ones - permanently wrong bucket values, no correction, no failing test. At minimum add a shrink-then-reread test pinning idempotent re-extraction, and consider an identity check (first-N-bytes hash or inode) to detect non-append rewrites.

src/daemon/token_usage_worker.rs
```diff
@@ -0,0 +1,1499 @@
...
+        &sink,
+    )
+    // Cross-session replacements flagged other sessions during the batch
+    // commits: reconcile them now, DB-only (their files are not needed and
+    // may no longer exist).
+    .and_then(|_| reconcile_flagged_sessions(token_db, &sink));
+    if let Err(e) = &result {
+        // Keyed by the rollup session id (subagent transcripts track under
+        // their parent), matching the row ensure_file created.
+        let _ = token_db.record_error(
```
svarlamov-git-ai, line 564: **[Token-usage accuracy review — finding 14/15]** (stack-wide review vs upstream ccusage)

A failed cross-session reconcile is charged to the wrong file: `process_file(...).and_then(|_| reconcile_flagged_sessions(...))` means a sink failure while reconciling some OTHER flagged session records the error against the current task's perfectly healthy transcript (whose own pass fully succeeded and whose size/mtime snapshot was already written), putting it into escalating 5s/30s/5m/30m error backoff that defers its future appends and points diagnostics at the wrong transcript.

**Failure scenario:** Telemetry backpressure during reconcile of a flagged session -> the healthy file that happened to trigger the pass accumulates processing_errors on every retry, its fresh usage is delayed up to 30 minutes per append, and last_error blames it for another session's failure. Record reconcile failures separately (or against the flagged session), and add a test that a reconcile sink failure does not back off the triggering file.

src/daemon/token_usage_worker.rs
svarlamov-git-ai: Fixed, both halves — (a) reads are now byte-level (`read_until`) with lossy UTF-8 decoding per line: an invalid byte makes that line fail JSON parsing and be skipped while the cursor advances, matching upstream's byte reads (test: `invalid_utf8_lines_are_skipped_and_the_cursor_advances`). (b) A final line without a newline that already parses as complete JSON is counted before the size/mtime snapshot is written (test: `complete_final_line_without_newline_is_counted`); genuinely half-written JSON still stays pending until completed (test: `truncated_final_json_stays_pending_until_completed`).

src/daemon.rs
svarlamov-git-ai: Fixed — emission revisions are now floored to wall-clock seconds (`max(prev+1, now)`), so a rebuilt state database (flag off/on, or any local loss) resumes at revisions strictly above anything previously uploaded, with no epoch machinery or server change needed. Spec updated; the worker tests assert the floor and strict monotonicity.

tests/integration/token_usage.rs
svarlamov-git-ai: Addressed — `golden_fixture_extraction_is_pinned` runs the extractor over the real captured transcript (tests/fixtures/example-claude-code.jsonl) and insta-snapshots entry count (23), dedup identities (13 — the genuine streaming re-emits collapse), and per-model token sums; `secondary_fixtures_still_extract_usage` covers the thinking/plan/model-not-last captures, and `real_codex_fixture_shapes_parse` runs the codex extractor over a real rollout's session_meta/turn_context shapes (no real token_count capture exists in tree, so codex token math stays pinned by the unit fixtures). Any serde-shape regression against real transcripts now fails loudly.

src/daemon/token_usage_worker.rs
svarlamov-git-ai: Addressed the seams that are cheaply testable end-to-end: `resume_through_daemon_does_not_double_count` (claude --resume copies history with original ids into a new session file; asserts global dedup holds through the real daemon) and `subagent_transcript_rolls_up_to_parent_session` are new e2e tests. Two of the seams you list are deliberate residuals documented in the spec: sweep-discovered tasks intentionally sit outside the `await` drain barrier (matching stream-worker semantics — the drain services notifications; sweeps are eventual), and daemon-restart resume is covered at the worker level (corrupt-state/cursor-reset and idempotent re-read tests) rather than with a kill-and-restart e2e, which our harness can't do hermetically today.

src/daemon/token_usage_worker.rs
svarlamov-git-ai: Documented as an accepted limitation in the spec — TokenUsage rides the shared uploader, so the 6-attempt abandonment applies to it exactly as it does to every other metric kind; the `emitted_seq` revision means a later genuine change to the same bucket re-emits with a higher revision and heals the gap. Changing the abandonment policy is uploader-wide and out of scope for this stack.

src/daemon/token_usage_worker.rs
svarlamov-git-ai: Addressed — file truncation/rewrites now have pinned behavior: shrink resets the cursor and re-reads from 0, dedup absorbs the replay, and `shrunken_file_rereads_and_pins_the_stale_residual` pins the one residual (an entry whose content was rewritten away keeps its old row, since entries are never deleted on re-read). That residual is documented in the spec as accepted: transcripts are append-only in practice; rewrites indicate corruption, and over-counting a corrupted file's history is bounded by 90-day retention.

src/daemon/token_usage_worker.rs
svarlamov-git-ai: Fixed — reconcile failures are decoupled from the triggering file: `reconcile_flagged_sessions` runs after per-file bookkeeping, its failures are logged and leave `needs_reconcile` set for retry (pinned in `failed_reconcile_sink_keeps_the_flag_for_retry`), and never call `record_error` on the notifying file, so an unrelated foreign-session correction can no longer put a healthy live transcript into 30-minute backoff.

src/daemon/token_usage_worker.rs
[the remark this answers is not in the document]
svarlamov-git-ai: Fixed — the enumeration/stat scan was extracted into `sweep_candidates(&streams_db, &token_db)` and `enqueue_sweep_tasks` now awaits it inside `tokio::task::spawn_blocking` (src/daemon/token_usage_worker.rs:309), so the async runtime thread is never blocked by the sweep. Consistent with `process_task` and covered by the sweep scheduling tests, which drive `sweep_candidates` directly.

tests/integration/token_usage.rs
svarlamov-git-ai: **Re-review at c2977628a: PARTIAL (major part fixed).** `golden_fixture_extraction_is_pinned` (insta over the real 23-usage-line capture, per-model sums + distinct keys), `secondary_fixtures_still_extract_usage`, and `real_codex_fixture_shapes_parse` close the Claude side well. Residual: neither codex fixture contains a single `token_count` line (`grep -c token_count tests/fixtures/codex-session-*.jsonl` → 0/0), so the Codex *usage* shape — the accuracy-bearing part — is still pinned only by synthetic lines that share the parser's own assumptions. When a real rollout with `token_count` events can be captured, adding it to the golden set closes the loop. Low priority; not blocking.

src/daemon/token_usage_worker.rs
svarlamov-git-ai: **Re-review at c2977628a: PARTIAL (identity seams fixed, loop still untested).** `resumed_session_through_the_daemon_does_not_double_count` and `subagent_transcript_rolls_up_to_the_parent_session` close the two identity seams — the highest-value part of this finding. Residual: the `run()` loop machinery (startup sweep, 30-min ticker, prune ordering, crash-recovery `reconcile_flagged`) is still never driven with real work at any level, and the spec's new accepted-limitation ("sweep-discovered work sits outside the await barrier") documents the *semantics* but doesn't provide any test hook — a regression in `sweep_candidates` filtering or a select-arm mistake still ships undetected, killing backfill/healing silently. A test-only drain covering the sweep queue plus one daemon-restart-backfill e2e would close it.

src/daemon/token_usage_worker.rs
svarlamov-git-ai: **Re-review at c2977628a: DOC-RESOLVED, but the accepted-limitation text understates the scope.** The spec says "a correction whose bucket never changes again is lost" — but the same mechanism loses FIRST emissions too: a historical backfill performed while uploads fail for >~20h (6 attempts) abandons every settled bucket's first emission, i.e. the entire history is missing server-side, not just corrections, and fingerprints ensure it is never re-sent. Suggest at minimum extending the doc to say "emissions (first or corrective)"; better, a cheap mitigation: for event id 9, cap the retry backoff at 24h instead of abandoning (volume is bounded — one record per bucket-revision), or clear the affected bucket fingerprints when the uploader gives up.

tests/integration/token_usage.rs
svarlamov-git-ai: Closed the residual — tests/fixtures/codex-session-token-count.jsonl is now a sanitized excerpt of a real captured rollout containing 8 production token_count lines (full real shapes: total/last_token_usage with cache_write_input_tokens, rate_limits/credits envelope, real session_meta + turn_context), and `golden_codex_token_count_extraction_is_pinned` snapshots entry count, model resolution, delta sums, and key distinctness over it. Codex usage shapes are no longer pinned only by synthetic JSON.

src/daemon/token_usage_worker.rs
svarlamov-git-ai: Closed — the real `run()` loop is now exercised directly: the sweep interval is an injectable field and the emission sink is injectable in tests (the metrics DB is a process-lifetime singleton, so run()-level tests can't use a real telemetry handle). `run_loop_startup_sweep_backfills_and_ticker_picks_up_appends` covers the startup sweep (no notification ever sent) and the ticker discovering appended lines on its own; `run_loop_restart_backfills_from_the_committed_cursor` covers a fresh worker over the same databases backfilling lines written while no worker ran, without re-emitting handled history.

src/daemon/token_usage_worker.rs
svarlamov-git-ai: Scope corrected in the spec — the limitation now states that abandonment loses any emission, first emissions included (e.g. a historical backfill that ran entirely during a >20h upload outage), not just corrections; a bucket that never changes again stays lost, while any later change re-emits at a higher revision and heals that bucket.

tests/integration/token_usage.rs
svarlamov-git-ai: **Re-review at a981487e1: FIXED — resolving.** `golden_codex_token_count_extraction_is_pinned` runs over `tests/fixtures/codex-session-token-count.jsonl`, which contains 8 genuinely captured `token_count` lines (I verified authenticity: the cumulative `total_token_usage` equals the running sum of `last_token_usage` for every field across all 8 events — internally consistent in a way synthetic data rarely is). I independently recomputed the pinned sums (input 59,288 / output 1,144 / cache_read 183,296 / reasoning 49 / total 243,728) and they match the snapshot, including the input-minus-cached normalization and cached-clamp. Distinct-key and model-resolution assertions included. This closes the Codex golden gap. (One non-accuracy issue with the fixture's sanitization posted separately.)

src/daemon/token_usage_worker.rs
svarlamov-git-ai: **Re-review at a981487e1: FIXED — resolving.** `run_loop_startup_sweep_backfills_and_ticker_picks_up_appends` and `run_loop_restart_backfills_from_the_committed_cursor` drive the real `run()` select loop (injectable `sweep_interval`, cfg(test)-only sink; production path unchanged — `make_sink` falls through to `persist_events`). Startup sweep with no notification, ticker pickup of unannounced appends, and daemon-restart backfill with an exactly-once assertion on the pre-restart bucket are exactly the seams this thread was about. Together with the park-chain test, the worker machinery is now regression-guarded.

src/daemon/token_usage_worker.rs
svarlamov-git-ai: **Re-review at a981487e1: DOC-RESOLVED (defensible) — resolving.** The accepted-limitation now states the true scope: "this loses any emission — first emissions included (e.g. a historical backfill that ran entirely during a >20h upload outage), not just corrections," with the per-bucket healing condition stated precisely. That was the ask; the documented trade-off is a reasonable, explicit decision.

PR #2210, opened 2026-08-24
Token-usage emission-dedup e2e coverage, retention prune, exclude gate, and spec

svarlamov-git-ai:
## Summary

Final PR of the TokenUsage stack: hardening and end-to-end coverage of the emission-dedup contract.

- **E2E tests** for the server-sync guarantees: unchanged buckets stay quiet across re-reads; a sidechain replay corrected by the parent's entry re-emits the bucket with *lower* totals; a bucket emptied by a cross-bucket streaming replacement emits an all-zero event exactly once (and stays quiet afterwards); a deleted transcript is handled without new or zeroing events.
- **Retention prune** wired into the worker sweep: per-entry rows and bucket fingerprints older than 90 days are dropped off the async loop; cursors are monotonic so pruned history is never re-read.
- **Repo exclude gate**: TokenUsage events are transcript-derived like SessionEvents — sweeps keep processing transcripts of sessions tracked before a repo was added to `exclude_repositories` — so `should_deliver_metric_event` now gates both kinds on the event's repo_url at upload time (filtered events are marked delivered without being uploaded, same contract as SessionEvents; covered by a daemon-mode test mirroring the SessionEvent one).
- **`docs/token-usage-events-spec.md`**: pipeline, state database, emission semantics, and ccusage deviations for reviewers.

### DIFF
--- a/docs/token-usage-events-spec.md
+++ b/docs/token-usage-events-spec.md
@@ -17,8 +17,7 @@ transcript file --(incremental read, cursor in token-usage DB)-->
     - prefilter substring check before any JSON parsing
     - per-entry token counts, model, timestamp, transcript costUSD
 --> usage_entries (deduplicated per-entry rows, token-usage DB)
---> dirty (model, bucket) set
---> re-aggregate + fingerprint compare (bucket_state)
+--> reconcile all session buckets: aggregate + fingerprint compare (bucket_state)
 --> TokenUsage MetricEvent --> telemetry queue --> POST /worker/metrics/upload
 ```
 
@@ -63,9 +62,15 @@ The server upserts on `(session_id, model, bucket_ts)` with the newest
 so a re-emitted correction - including one that lowers a bucket or zeroes it
 out entirely - always supersedes the previous value. A bucket is emitted iff
 its aggregate's fingerprint differs from the last emitted fingerprint; an
-emptied bucket therefore emits an all-zero event exactly once. Buckets are
-marked emitted only after the telemetry queue accepted the events, so a
-failed hand-off retries on the next pass.
+emptied bucket therefore emits an all-zero event exactly once. Changed
+buckets are found by reconciling fingerprints across all of the session's
+buckets on every pass (no pending-emission state lives in memory), buckets
+are marked emitted only after the telemetry queue accepted the events, and
+the quiet-skip size/mtime snapshot is written only after a fully successful
+pass - so a failed hand-off or a crash at any point is healed by the next
+pass. At upload time TokenUsage events get the same repo allow/exclude gate
+as SessionEvents (they are transcript-derived, so they keep flowing for
+sessions tracked before a repo was excluded).
 
 Values (`token_usage_pos`): `bucket_ts`, `input_tokens`, `output_tokens`,
 `cache_read_tokens`, `cache_write_tokens`, `total_tokens`,
--- a/docs/token-usage-events-spec.md
+++ b/docs/token-usage-events-spec.md
@@ -1,0 +1,90 @@
+# TokenUsage Events
+
+Token usage and estimated cost per `(session_id, model, bucket_ts)` in
+5-minute UTC buckets, computed from local agent transcripts and emitted as
+metric event id 9 (`TokenUsageValues`, `src/metrics/events.rs`).
+
+The parsing/deduplication logic is ported from
+[ccusage](https://github.com/ccusage/ccusage) (MIT) and adapted to git-ai's
+incremental transcript streaming; per-agent deviations are documented in
+`src/token_usage/{claude,codex}.rs`. v1 covers Claude Code and Codex.
+
+## Pipeline
+
+```
+transcript file --(incremental read, cursor in token-usage DB)-->
+  per-agent extractor (src/token_usage/{claude,codex}.rs)
+    - prefilter substring check before any JSON parsing
+    - per-entry token counts, model, timestamp, transcript costUSD
+--> usage_entries (deduplicated per-entry rows, token-usage DB)
+--> dirty (model, bucket) set
+--> re-aggregate + fingerprint compare (bucket_state)
+--> TokenUsage MetricEvent --> telemetry queue --> POST /worker/metrics/upload
+```
+
+Driven by `TokenUsageWorker` (`src/daemon/token_usage_worker.rs`):
+
+- **Triggers:** a non-blocking ping from the stream worker after it processes
+  a transcript, a startup sweep, and a 30-minute ticker. Sweeps enumerate the
+  streams database's `transcript` rows, so every session git-ai knows about is
+  backfilled (a new file starts at byte offset 0); transcripts that no longer
+  exist are skipped. Nothing runs on the trace2 ingestion path.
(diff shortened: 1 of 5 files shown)


docs/token-usage-events-spec.md
```diff
@@ -0,0 +1,120 @@
...
+--> usage_entries (globally deduplicated per-entry rows, token-usage DB)
+--> reconcile all session buckets in one pass: aggregate + fingerprint
+    compare (bucket_state)
+--> TokenUsage MetricEvent --> telemetry queue --> POST /worker/metrics/upload
+```
+
+Driven by `TokenUsageWorker` (`src/daemon/token_usage_worker.rs`):
+
+- **Triggers:** a non-blocking ping from the stream worker after it processes
+  a transcript (success or failure of that pass), a startup sweep, and a
```
svarlamov-git-ai: **[follow-up review — spec accuracy, low] The spec says notifications fire on 'success or failure of that pass', but the stream worker only notifies on its success arms.**

`notify_token_usage` is called from the two `Ok(Ok(()))` arms in src/daemon/stream_worker.rs (process-next-task and drain paths); the error/panic arms go to retry handling without a ping. So a transcript whose stream pass fails (or exhausts retries) produces no token-usage notification and its usage waits for the next 30-minute sweep — acceptable behavior, but the spec sentence documents a guarantee the code does not provide. Either drop the parenthetical or add the notification to the failure arms (the token pass reads the file independently, so notifying on stream failure would actually be safe and would tighten the await barrier's coverage).

svarlamov-git-ai: Fixed — the spec now says the ping fires after a *successful* stream pass (failed passes are picked up by the next sweep), and its invalidation/revision/deletion sections were updated for the durable `needs_reconcile` flow, pre-sink revision reservation, and streaming-gate-independent flag-off deletion.

src/daemon/telemetry_worker.rs
```diff
@@ -1093,7 +1093,11 @@ where
 }
 
 fn should_deliver_metric_event(config: &Config, event: &MetricEvent) -> bool {
-    if event.event_id != MetricEventId::SessionEvent as u16 {
+    // Transcript-derived events keep flowing for sessions tracked before a
+    // repo was excluded, so they get the same upload-time repo gate.
+    let transcript_derived = event.event_id == MetricEventId::SessionEvent as u16
```
svarlamov-git-ai, line 1172: **[Token-usage accuracy review — finding 10/15]** (stack-wide review vs upstream ccusage)

Repo exclude-gate coverage has two holes: (a) events with NO repo_url bypass an exclude-only config entirely (`is_allowed_repository_with_remotes(None)` skips the exclusion branch, config.rs:445-455), so an excluded repo's token usage leaks whenever resolution fails (work_dir unset + infer_cwd failure), and since the fingerprint is then marked emitted the data is never re-sent gated - while under an allowlist, all no-repo sessions are silently dropped; (b) the only e2e test hand-crafts events and inserts them straight into the metrics DB, bypassing the real checkpoint->worker->upload composition; also (c) corrections filtered while a repo is temporarily excluded are marked delivered AND their fingerprints marked emitted, so un-excluding leaves the server permanently holding pre-exclusion values.

**Failure scenario:** User excludes git@github.com:acme/private (no allowlist); a session's cwd inference fails -> its full token/cost data uploads unfiltered and is never corrected. Or: repo excluded for a week, a sidechain replacement lowers a bucket from 50,000 to 200 cache-read tokens during that window, repo un-excluded -> the inflated value stays server-side forever. Needed: a TokenUsage gate test with repo_url absent (pin either behavior), an e2e through the real pipeline with an excluded remote, and a documented/tested decision on held-vs-dropped corrections during exclusion.

svarlamov-git-ai: Addressed — the gate decisions are now pinned as tests and written down as decisions rather than accidents: `token_usage_without_repo_url_passes_an_exclude_only_gate` pins the fail-open (exclude rules can only match a URL; a repo-less event has nothing to match, same as SessionEvent), `daemon_marks_repository_filtered_token_usage_events_delivered_without_uploading_them` pins the filtered path, and `excluded_repo_token_usage_never_uploads_via_the_real_pipeline` demonstrates defense layer 1 (the checkpoint-time gate refuses tracking for excluded repos, so no transcript stream ever exists to extract from). Corrections that accrue while a repo is excluded are deliberately not replayed on un-exclusion — pinned in the spec's decisions section.

docs/token-usage-events-spec.md
```diff
@@ -0,0 +1,169 @@
...
+  in practice; the shrink path re-reads from byte 0 and dedup absorbs the
+  replay.
+- **Sweep-discovered work sits outside the `git-ai await` drain barrier**
+  (see Pipeline): backfill is eventual, matching stream-worker semantics.
+- **A fork burst split across passes can over-count once.** The end-of-file
+  flush releases a parked first turn after the burst window passes in wall
+  clock. If Codex writes the two replayed burst lines more than a second
+  apart (buffered/laggy serialization carrying sub-second recorded
+  timestamps) and a pass lands in the gap, the parked replay is released as
+  own usage and its burst partner counts too — bounded to that one pair per
```
svarlamov-git-ai: **[round-5 review — doc accuracy / cheap hardening, low] The split-burst residual is not bounded to one pair: after the flush, the state is Done, so the *entire* remaining replay burst counts.**

`flush` transitions the replay state to `ReplayState::Done` (src/token_usage/codex.rs). In the documented gap scenario, the pass releases replay line 1 (delta T1) and disarms the filter entirely; when Codex then writes the rest of the rewritten burst (lines 2..N, cumulative totals T2..TN), the Done arm counts every delta — (T2-T1) + ... + (TN-T(N-1)). Worst case over-count is TN, the fork's whole replayed history, not 'that one pair'; the stated bound only holds for a 2-line burst.

A one-line change makes the documented bound true (and tighter): have `flush` transition to `SkippingBurst { last_ts_ms: first_ts_ms }` instead of `Done`. The remaining burst lines (rewritten timestamps within the window of line 1, chaining per skipped event) are then skipped while `prev_totals` still absorbs their cumulative totals, and a genuine own turn — whose timestamp is necessarily >1s past the released line, since the flush itself required that much wall clock — exits the skip and counts normally. Over-count becomes T1 (just the parked replay) even for long bursts. The single-turn-fork flush case is unaffected: `has_pending()` stays false, the file settles, and a later real event outside the window counts exactly as it does under Done.

svarlamov-git-ai: Taken — implemented the one-line hardening rather than the wording fix: `flush` now transitions to `SkippingBurst {{ last_ts_ms: first_ts_ms }}` instead of `Done`, so late-arriving burst partners land within the window and are skipped while `prev_totals` absorbs their cumulative totals; the over-count is truly bounded to the single released event. Pinned in `late_burst_partners_after_a_flush_release_are_skipped` (release → chained partners skipped → genuine own turn counts with the burst absorbed into the baseline), plus an own-turn-after-flush assertion in the existing single-turn-fork test. The spec bullet is updated to state the tightened bound.

PR #2215, opened 2026-08-25
Throttle transcript passes to a ~30% CPU duty cycle

svarlamov-git-ai:
## Summary

First-run backfill over ~1GB of local codex transcripts pinned a CPU core for over a minute. This PR throttles transcript passes to a ~30% duty cycle of one core: each batch's work `W` is followed by a proportional pause of `W * 7/3` (capped at 5s), so a large backfill takes ~3x longer instead of pinning a core.

- The pause is skipped below a 5ms work floor, so quiet skips, small incremental appends, and the `git-ai await` drain barrier stay fast.
- The pause is also paid after the final batch of a pass, so back-to-back file passes during a multi-file backfill hold the duty cycle across files, not just within one file.
- Pauses sleep in 50ms chunks checking the shutdown flag, so daemon shutdown stays prompt.

## Verification

Verified locally with `task dev` on a machine with ~400MB of tracked transcripts (466 files): the fresh backfill extracted 19,745 entries into 1,533 buckets in ~60s wall clock with 11s of total daemon CPU — ~18% of one core average, no errors in the daemon log.

### DIFF
--- a/src/daemon/token_usage_worker.rs
+++ b/src/daemon/token_usage_worker.rs
@@ -32,11 +32,12 @@
 //! - The `token_usage_metrics` feature flag gates the worker at daemon
 //!   startup (like `transcript_streaming`); when it is off, the token-usage
 //!   database is deleted so no collected data is retained.
-//! - Transcript passes are CPU-throttled to a ~30% duty cycle of one core
-//!   (each batch's work is followed by a proportional pause), so a large
-//!   first-run backfill takes ~3x longer instead of pinning a core. Passes
-//!   below the work floor are not throttled, keeping the notify/drain path
-//!   fast.
+//! - Sweep/backfill passes are CPU-throttled to a ~30% duty cycle of one
+//!   core (each batch's work is followed by a proportional pause), so a
+//!   large first-run backfill takes ~3x longer instead of pinning a core.
+//!   Notification-driven passes are never throttled: they sit on the
+//!   `git-ai await` drain barrier's post-commit critical path, and a task
+//!   promoted from the sweep queue by a notification runs unthrottled too.
 
 use std::collections::{HashMap, HashSet, VecDeque};
 use std::io::{BufReader, Seek, SeekFrom};
@@ -86,14 +87,14 @@ fn read_line_bytes(
 const BATCH_MAX_ENTRIES: usize = 1_000;
 const BATCH_MAX_BYTES: usize = 4 * 1024 * 1024;
 
-/// CPU duty cycle for transcript passes: after each batch, the pass sleeps
-/// long enough that the work occupies at most this share of one core
-/// (work W is followed by a pause of W * (100 - duty) / duty). A multi-
+/// CPU duty cycle for sweep-origin transcript passes: after each batch, the
+/// pass sleeps long enough that the work occupies at most this share of one
+/// core (work W is followed by a pause of W * (100 - duty) / duty). A multi-
 /// gigabyte backfill therefore takes ~3x longer instead of pinning a core.
+/// Notification-origin passes are never throttled (see `TaskOrigin`).
 const THROTTLE_DUTY_PERCENT: u32 = 30;
-/// Passes doing less work than this per batch (quiet skips, small appends)
-/// are not throttled at all, so notification-driven work and the `git-ai
-/// await` drain barrier stay fast.
+/// Batches doing less work than this owe no pause (quiet skips, small
+/// appends swept between notifications).
 const THROTTLE_MIN_WORK: Duration = Duration::from_millis(5);
 /// Upper bound on a single pause (guards pathologically slow batches, and
 /// bounds how long a pass can delay a drain or shutdown).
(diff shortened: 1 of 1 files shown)


src/daemon/token_usage_worker.rs
```diff
@@ -750,6 +788,10 @@ fn process_file(
             pending_flush: extractor.has_pending(),
             min_bucket_ts: cutoff,
         })?;
+        // CPU throttle: pay for this batch's work with a proportional pause
+        // (also after the final batch, so back-to-back file passes during a
+        // large backfill hold the duty cycle too).
+        throttle_sleep(throttle_pause_for(batch_started.elapsed()), shutdown_flag);
```
svarlamov-git-ai: **[review — await latency regression, medium] The throttle applies to notification/drain-path passes too — only the 5ms per-batch floor protects them — so `git-ai await` pays the duty cycle, and a large first-pass via the notify queue can now blow the 30s default timeout.**

Nothing distinguishes task origin here: `handle_drain` -> `process_task` -> `process_file` runs this same call, so every batch of a drain-barrier pass with >5ms of work pays +233% of its work as pause, *inside* Phase 2b of `await_completion`. Three compounding pieces:

1. **The floor only protects small appends.** A long agent turn appending a few MB pays a proportional pause per batch; post-commit await latency for notify work is now ~3.3x its parse+commit time by design.
2. **Queue promotion moves big backfills into the barrier.** `enqueue_notify` promotes a file already queued for sweep backfill into the notify queue on any fresh checkpoint for that session (daemon restarted mid-conversation, resumed old session). By this PR's own measurement (400MB ≈ 60s throttled ≈ 6.7MB/s vs ~18s unthrottled), a 200MB transcript promoted into the drain barrier goes from ~9s to ~30s+ — past `git-ai await`'s default 30s timeout, which sets `result.timed_out` and skips the Phase 3 telemetry flush. The PR goal says the drain barrier 'stays fast'; it only stays fast below the floor.
3. **The final-batch pause is pure dead time on the drain path.** Its cross-file purpose (holding the duty cycle across back-to-back backfill passes) only matters for sweep work; for a drain-path pass it just delays the barrier release by up to `min(7W/3, 5s)` after all work is done.

The queues already encode the right signal: thread the task's origin (notify vs sweep) into `process_task_blocking`/`process_file` and throttle only sweep-origin passes — the drain barrier then never pays. Alternatively, switch to pause-*debt* paid at the start of the NEXT batch/pass instead of after the current one: back-to-back backfill still holds ~30% duty, while the last batch before an idle queue (and every drain-path completion) releases immediately. Minor aggravator either way: `batch_started.elapsed()` is wall time, so slow/contended I/O and the `commit_batch` write inflate 'work' and lengthen pauses beyond the CPU-duty target — conservative for the CPU goal, but additional await latency.

src/daemon/token_usage_worker.rs
```diff
@@ -1722,6 +1764,36 @@ mod tests {
         assert_eq!(identity.external_session_id, "child-ext");
     }
 
+    #[test]
+    fn throttle_pause_holds_the_duty_cycle() {
```
svarlamov-git-ai, line 1830: **[review — test coverage, low] Only the two helper functions are pinned; nothing tests that `process_file` actually throttles, so deleting the call site at line 794 fails no test.**

`throttle_pause_holds_the_duty_cycle` and `throttle_sleep_exits_promptly_on_shutdown` cover the math and the chunked sleep in isolation, but the integration — that a pass over more than the work floor pays a proportional pause, and (per the PR description) that quiet skips and small appends do NOT — has no pin. Given the repo's strict-TDD rule, consider making the sleep injectable the way the sink already is (a `&dyn Fn(Duration)` threaded through `process_task_blocking`/`process_file`, defaulting to `throttle_sleep`) and asserting the recorded pauses for a two-batch fixture pass: proportional pause invoked per batch over the floor, zero invocations for a quiet-skip pass. That avoids wall-clock timing assertions entirely and would also pin whichever drain-path exemption shape you settle on for the other comment.

src/daemon/token_usage_worker.rs
svarlamov-git-ai: Fixed with the origin-gating you recommended — tasks now carry a `TaskOrigin` derived from the queue they pop from: sweep-origin passes throttle, notify-origin passes never do, and a task promoted out of the sweep queue by a notification pops as notify-origin and runs unthrottled (pinned in the promotion test). That removes all three stacked effects from the await barrier: no 3.3x drain passes, no final-batch dead time before the barrier releases, and no promoted backfill paying throttle inside the barrier. The end-to-end run()-loop test (`sweep_passes_are_throttled_and_notify_passes_are_not`) drives both origins over the same file and asserts the notify pass never touches the throttle. On wall-time-as-work: kept deliberately for sweep passes — counting commit/IO time inflates only background pauses, which errs gentle (lower effective duty), never against the drain path now that notify passes are exempt.

src/daemon/token_usage_worker.rs
svarlamov-git-ai: Added — the sleeper is injectable like the sink (`process_file` takes `Option<&dyn Fn(Duration)>`; the worker's `make_throttle` supplies the real chunked sleep, tests supply a recorder). `throttled_passes_pay_a_pause_per_batch_and_quiet_skips_pay_nothing` pins exactly one pause per batch commit on a multi-batch pass (1001 entries → 2 batches → 2 recorded pauses) and zero on a quiet skip, with no wall-clock assertions; deleting the call site now fails it. The run()-level origin test additionally pins that the throttle is wired through `process_task` for sweep tasks and absent for notify tasks.

PR #2232, opened 2026-08-27
feat(pricing): keep models.dev context tiers and fast multipliers in the catalog

svarlamov-git-ai:
The catalog previously flattened models.dev to four flat rates, so
long-context tiers and fast/priority speed pricing were unpriceable
downstream. ModelPricing now carries the model's long-context band
(whole-request above-threshold rates + threshold from models.dev
cost.tiers, lowest context bound wins) and a fast-speed multiplier.

Neither vendors nor models.dev publish fast-tier pricing, and
models.dev's first-party Anthropic entries omit their long-context
premium (it only appears under gateway spellings the provider allowlist
drops), so both are hand-tracked override tables applied at trim time:
fast multipliers ported verbatim from ccusage's overrides (MIT), and
Anthropic's published 200K premium for the 1M-context sonnet/opus
models.

Cache rates become Option so unpublished rates stay distinguishable
from explicit zeros; cache_read_rate()/cache_write_rate() accessors
keep ccusage's models.dev-loader defaults (0.1x / 1.25x input). The
snapshot is regenerated with the new fields; old snapshot/cache JSON
still parses (missing fields default to None / 1.0).

Cost math is unchanged in this commit beyond the 1.25x cache-write
default for unpublished rates (cost-inert for Codex, which reports no
cache writes); consumers of the new tier fields land separately.

---

### DIFF
--- a/src/metrics/local_stats.rs
+++ b/src/metrics/local_stats.rs
@@ -582,8 +582,8 @@ impl CodexSessionAccum {
 fn estimate_cost(acc: &TokenAccum, pricing: &ModelPricing) -> f64 {
     (acc.input as f64 * pricing.input
         + acc.output as f64 * pricing.output
-        + acc.cache_creation as f64 * pricing.cache_write
-        + acc.cache_read as f64 * pricing.cache_read)
+        + acc.cache_creation as f64 * pricing.cache_write_rate()
+        + acc.cache_read as f64 * pricing.cache_read_rate())
         / 1_000_000.0
 }
 
@@ -1419,8 +1419,8 @@ mod tests {
         // 20 cache read, 10 cache creation.
         let expected = (100.0 * pricing.input
             + 50.0 * pricing.output
-            + 20.0 * pricing.cache_read
-            + 10.0 * pricing.cache_write)
+            + 20.0 * pricing.cache_read_rate()
+            + 10.0 * pricing.cache_write_rate())
             / 1_000_000.0;
         assert!(expected > 0.0);
         assert!((stats.tokens.estimated_cost_usd - expected).abs() < 1e-12);
--- a/src/metrics/local_stats.rs
+++ b/src/metrics/local_stats.rs
@@ -579,6 +579,12 @@ impl CodexSessionAccum {
     }
 }
 
+/// Rough display estimate for `git-ai usage`: flat base rates with the
+/// Claude-shape defaults for unpublished cache rates (0.1x input for reads,
+/// 1.25x for writes) for every tool. The TokenUsage event pipeline
+/// (`src/token_usage/cost.rs`) is the authoritative dollars — it prices
+/// per entry with tier, speed, and per-tool cache-rate semantics (e.g. Codex
+/// bills unpublished cache reads at the full input rate).
 fn estimate_cost(acc: &TokenAccum, pricing: &ModelPricing) -> f64 {
     (acc.input as f64 * pricing.input
         + acc.output as f64 * pricing.output
--- a/src/metrics/local_stats.rs
+++ b/src/metrics/local_stats.rs
@@ -611,7 +611,7 @@ fn cost_for_message_slice(entries: impl Iterator<Item = (String, TokenAccum)>) -
     }
     model_totals
         .iter()
-        .filter_map(|(model, acc)| pricing_for(model).map(|p| estimate_cost(acc, p)))
+        .filter_map(|(model, acc)| pricing_for(model).map(|p| estimate_cost(acc, &p)))
         .sum()
 }
 
@@ -648,7 +648,7 @@ fn build_token_summary(
         if let Some(pricing) = pricing_for(&short) {
             *cost_by_day
                 .entry(ts_to_local(ts).date_naive())
-                .or_insert(0.0) += estimate_cost(&acc, pricing);
+                .or_insert(0.0) += estimate_cost(&acc, &pricing);
         }
 
         let entry = model_tokens.entry(short.clone()).or_default();
@@ -686,7 +686,7 @@ fn build_token_summary(
         if let Some(pricing) = pricing_for(&short) {
             *cost_by_day
                 .entry(ts_to_local(acc.last_usage_ts).date_naive())
-                .or_insert(0.0) += estimate_cost(&mapped, pricing);
+                .or_insert(0.0) += estimate_cost(&mapped, &pricing);
         }
 
         let entry = model_tokens.entry(short.clone()).or_default();
@@ -742,7 +742,7 @@ fn build_token_summary(
         summary.cache_read += acc.cache_read;
         summary.cache_creation += acc.cache_creation;
 
-        let cost = pricing_for(&model).map(|p| estimate_cost(&acc, p));
+        let cost = pricing_for(&model).map(|p| estimate_cost(&acc, &p));
         if let Some(c) = cost {
             summary.estimated_cost_usd += c;
         }
--- a/src/metrics/model_pricing.rs
+++ b/src/metrics/model_pricing.rs
@@ -34,6 +34,13 @@ use std::sync::{Mutex, OnceLock};
 const EMBEDDED_SNAPSHOT: &str = include_str!("models_dev_pricing_snapshot.json");
 const MODELS_DEV_API_URL: &str = "https://models.dev/api.json";
 const CACHE_FILE_NAME: &str = "models_dev_pricing.json";
+/// Format version of the on-disk cache. Bump whenever `ModelPricing`'s
+/// serialized shape changes meaning: caches written under another format are
+/// ignored (the embedded snapshot applies) until a refresh rewrites them,
+/// never reinterpreted — a pre-tier cache stored explicit 0.0 for unpublished
+/// cache rates, which the optional-rate schema would read as published-free
+/// pricing, and it carries no tiers or fast multipliers.
+const PRICING_CACHE_VERSION: u32 = 2;
 /// Minimum interval between refresh *attempts* (successful or not), so
 /// offline machines don't stall on the fetch timeout at every invocation.
 const REFRESH_INTERVAL_SECS: u64 = 24 * 3600;
@@ -308,6 +315,7 @@ fn catalog() -> &'static PricingCatalog {
     CATALOG.get_or_init(|| {
         if !use_embedded_only()
             && let Some(cache) = cache_path().and_then(|path| read_json_file::<PricingCache>(&path))
+            && cache.is_current_format()
             && !cache.models.is_empty()
         {
             return PricingCatalog::from_entries(cache.models);
@@ -325,8 +333,17 @@ fn embedded_catalog() -> PricingCatalog {
 /// attempt (used for throttling, so failed attempts are also spaced out).
 #[derive(Serialize, Deserialize)]
 struct PricingCache {
+    /// See [`PRICING_CACHE_VERSION`]; pre-version files default to 0.
+    #[serde(default)]
+    version: u32,
     last_attempt_at: u64,
     models: BTreeMap<String, ModelPricing>,
+}
+
+impl PricingCache {
+    fn is_current_format(&self) -> bool {
+        self.version == PRICING_CACHE_VERSION
+    }
 }
 
 fn cache_path() -> Option<PathBuf> {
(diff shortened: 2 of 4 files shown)


src/metrics/model_pricing.rs
```diff
@@ -56,16 +67,137 @@ const PROVIDER_ALLOWLIST: [&str; 8] = [
...
 
-/// Per-million-token pricing for a model (USD). Cache fields default to 0 —
-/// models.dev omits them for models without prompt caching.
+/// Fast/priority speed multipliers over a request's whole cost, ported from
+/// ccusage's fast-multiplier-overrides.json (MIT License, Copyright (c) 2025
+/// @ryoppippi). No machine-readable source publishes these (OpenAI's Codex
+/// "fast" tier is not its API priority pricing), so they are hand-tracked
+/// against vendor price sheets. Exact entries apply to that catalog id only —
+/// "gpt-5.5-pro" must not inherit gpt-5.5's multiplier.
+const FAST_MULTIPLIER_EXACT: [(&str, f64); 6] = [
```
svarlamov-git-ai: **[code-review] Fallback pricing leaks hand-tracked fast multipliers and long-context tiers to uncataloged model ids**

The doc here says "Exact entries apply to that catalog id only — 'gpt-5.5-pro' must not inherit gpt-5.5's multiplier", but `pricing_for`'s fallbacks return catalog entries verbatim, multiplier and tier included:

- Token-boundary containment (`.filter(|(id, _)| contains_at_token_boundary(&model, id)).max_by_key(...)`): an uncataloged `gpt-5.5-codex` matches the `gpt-5.5` entry and bills fast requests at 2.5x (the codex-family fast multiplier is 2.0x per `gpt-5.3-codex`).
- `family_fallback` median: any unknown opus id (e.g. `claude-opus-4-9`, the doc's own "successors newer than the catalog" scenario) resolves to the median entry `claude-opus-4-7` — `fast_multiplier: 6.0` plus the 200K tier — so every fast-speed request on a new opus bills 6x, and the leaked threshold also flips `is_long_context`, polluting the wire's long-context token splits with conjectured tier data.

Pre-existing fallbacks only leaked base rates; multipliers/tiers riding along is new here. Consider stripping `fast_multiplier`/tier fields (reset to defaults) on non-exact matches, or at least on `family_fallback` results.

src/metrics/model_pricing.rs
```diff
@@ -285,9 +508,31 @@ fn trim_catalog(api_json: &str) -> Result<BTreeMap<String, ModelPricing>, String
             let Some(cost) = model.get("cost") else {
                 continue;
             };
-            if let Ok(pricing) = serde_json::from_value::<ModelPricing>(cost.clone()) {
-                entries.insert(model_id.to_lowercase(), pricing);
-            }
+            let Ok(cost) = serde_json::from_value::<ModelsDevCost>(cost.clone()) else {
```
svarlamov-git-ai, line 563: **[code-review] Strictly-typed tier parsing silently drops whole models the old code kept at flat rates**

`ModelsDevCost.tiers: Vec<ModelsDevTier>` makes a single unexpected tier shape (scalar element, string-encoded number like `"input": "4.0"`, `"size": "272000"`, or `tiers: null`) fail the whole `from_value`, and this `else { continue; }` then drops the model from the fetched catalog with no log. The previous `from_value::<ModelPricing>` ignored `tiers` as an unknown field and kept the model at flat rates. A models.dev format tweak now silently sends a model to the family-median fallback (or $0) even though its flat input/output rates were parseable.

Cheap hardening: deserialize `tiers` as `Vec<serde_json::Value>` (or `#[serde(deserialize_with = ...)]` that skips bad elements) so tier parse failures degrade to flat rates, and/or `debug_log` the skip.

src/metrics/model_pricing.rs
```diff
@@ -285,9 +508,31 @@ fn trim_catalog(api_json: &str) -> Result<BTreeMap<String, ModelPricing>, String
...
+                continue;
+            };
+            let (Some(input), Some(output)) = (cost.input, cost.output) else {
+                continue;
+            };
+            let model_id = model_id.to_lowercase();
+            let long_context = cost
+                .long_context_rates()
+                .or_else(|| prefix_override(&CLAUDE_LONG_CONTEXT_PREFIX, &model_id));
```
svarlamov-git-ai: **[code-review] Overrides are stamped at fetch/trim time, so fetched caches freeze old override tables indefinitely**

`trim_catalog` stamps `FAST_MULTIPLIER_*`/`CLAUDE_LONG_CONTEXT_PREFIX` into the entries it writes to the on-disk cache, and `catalog_with_id` returns `cache.models` verbatim whenever a current-format cache exists. `PRICING_CACHE_VERSION` only guards *shape* changes, so a binary upgrade that edits an override value (a new opus fast multiplier, a corrected tier) is inert on any machine holding a current-format cache until the next successful fetch — and the refresh path runs from `git-ai usage`, so a user who never runs that command keeps the stale multipliers forever while the daemon prices every entry with them (costs are event-time and never repriced). Stamping overrides at catalog-load time (in `PricingCatalog::from_entries` or `catalog_with_id`) would make them binary-versioned with no staleness window, and the cache would only need to carry the raw models.dev data.

src/metrics/model_pricing.rs
svarlamov-git-ai: Fixed in d3ee57782: the family-median fallback now neutralizes the fast multiplier and long-context tier to the median member's base rates (an unknown opus id no longer bills 6x or flips the long-context split). Token-boundary matches deliberately keep the full entry — that is how date-suffixed spellings inherit their base model's data, and ccusage's fuzzy lookup returns full entries (multiplier included) the same way.

src/metrics/model_pricing.rs
svarlamov-git-ai: Fixed in d3ee57782: tier bands are held as raw JSON and parsed per band, so a novel-shaped band (or tiers: null) degrades that model to its flat rates instead of dropping it from the catalog; parseable bands next to a malformed one still apply. Pinned by one_malformed_tier_band_degrades_to_flat_rates_not_a_dropped_model.

src/metrics/model_pricing.rs
svarlamov-git-ai: Fixed in d3ee57782: the override tables now apply when a catalog is LOADED (one choke point for the embedded snapshot and the fetched cache), so a table edit shipped in a new binary takes effect immediately; the trim output and snapshot become pure models.dev data. Application is idempotent over pre-split caches with baked values, and an edited table wins over them.

src/metrics/model_pricing.rs
```diff
@@ -56,16 +67,140 @@ const PROVIDER_ALLOWLIST: [&str; 8] = [
...
-/// models.dev omits them for models without prompt caching.
-#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+/// Fast/priority speed multipliers over a request's whole cost, ported from
+/// ccusage's fast-multiplier-overrides.json (MIT License, Copyright (c) 2025
+/// @ryoppippi). No machine-readable source publishes these (OpenAI's Codex
+/// "fast" tier is not its API priority pricing), so they are hand-tracked
+/// against vendor price sheets. Exact entries apply to that catalog id only —
+/// "gpt-5.5-pro" must not inherit gpt-5.5's multiplier.
+const FAST_MULTIPLIER_EXACT: [(&str, f64); 6] = [
+    ("gpt-5.6-sol", 2.0),
```
svarlamov-git-ai: **[code-review] ccusage's `gpt-5.6` pricing alias is unported: base-id fast usage bills at 1.0x instead of 2.0x**

This table ports ccusage's `fast-multiplier-overrides.json` entries, but not its `pricing_alias` rule that resolves `"gpt-5.6"` to `"gpt-5.6-sol"` before the multiplier lookup (upstream pins this with `gpt_5_6_alias_resolves_to_sol_across_pricing_metadata`). The snapshot's base `gpt-5.6` entry is priced identically to `gpt-5.6-sol`, so a rollout recording `model: "gpt-5.6"` (or a spelling like `gpt-5.6-codex` that the token-boundary fallback resolves to the base entry) with `service_tier = fast` prices at `fast_multiplier` 1.0 — half of what ccusage bills and half of what the `gpt-5.6-sol` spelling bills through this same table.

```suggestion
const FAST_MULTIPLIER_EXACT: [(&str, f64); 7] = [
    ("gpt-5.6", 2.0),
    ("gpt-5.6-sol", 2.0),
```

Worth pinning with a test asserting `pricing_for("gpt-5.6").unwrap().fast_multiplier == 2.0`. (ccusage's other alias, `gpt-5.3-spark` -> `gpt-5.3-codex-spark`, is also unported but has no fast-multiplier consequence today.)

src/metrics/models_dev_pricing_snapshot.json
```diff
@@ -79,261 +79,214 @@
...
+    "cache_read_above": 0.4
   },
-  "deepseek-chat": {
-    "input": 0.14,
-    "output": 0.28,
-    "cache_read": 0.0028,
-    "cache_write": 0.0
-  },
-  "deepseek-reasoner": {
+  "deepseek-v4-flash": {
```
svarlamov-git-ai, line 102: **[code-review] Snapshot regen silently drops `deepseek-chat` / `deepseek-reasoner` pricing to $0**

The regenerated snapshot no longer contains `deepseek-chat` and `deepseek-reasoner` (present on main at 0.14/0.28 per M). Neither id resolves through any remaining tier: no exact match, no surviving catalog id is token-boundary-contained in either spelling, and `FAMILY_FALLBACK_TOKENS` has no `deepseek` token — so `pricing_for` now returns `None` and usage recorded under those ids goes from a real cost on main to est_cost `None`/$0 on this branch, silently. Low reach (router/proxy setups recording those ids), but it is a concrete repricing regression shipped by this regen. If the upstream removal is intentional, consider adding `"deepseek"` to `FAMILY_FALLBACK_TOKENS` so future removals degrade to the family median instead of $0.

src/metrics/local_stats.rs
```diff
@@ -582,8 +582,8 @@ impl CodexSessionAccum {
 fn estimate_cost(acc: &TokenAccum, pricing: &ModelPricing) -> f64 {
     (acc.input as f64 * pricing.input
         + acc.output as f64 * pricing.output
-        + acc.cache_creation as f64 * pricing.cache_write
-        + acc.cache_read as f64 * pricing.cache_read)
+        + acc.cache_creation as f64 * pricing.cache_write_rate()
+        + acc.cache_read as f64 * pricing.cache_read_rate())
```
svarlamov-git-ai, line 592: **[code-review] Codex cache-read fallback here diverges from the authoritative Codex pricing shape (latent), and the deviation is no longer documented**

`cache_read_rate()` defaults an unpublished cache-read price to 0.1x input — the Claude rule — and `estimate_cost` applies it to all accumulators, including Codex sessions, while the stack's authoritative Codex shape (`cost.rs::codex_cost_usd`, added later in this stack) bills unpublished cache reads at the FULL input rate. Latent with today's snapshot (every caching codex model publishes `cache_read`), but a snapshot refresh that ships a caching codex model with a lagging cache-read price would make `git-ai usage` undercount its cache-read-dominated sessions by up to ~10x vs the emitted TokenUsage events. The `cost.rs` module doc enumerates this display path's deviations (tiering, multipliers, 1h split) but no longer names the cache-read one, which main's doc did. Suggest either restoring that one doc line or branching on `pricing.cache_read.is_none()` for codex accumulators.

src/metrics/model_pricing.rs
svarlamov-git-ai: Fixed in 315b4b6d9: added `("gpt-5.6", 2.0)` to `FAST_MULTIPLIER_EXACT` with a comment tying it to ccusage's `pricing_alias` (gpt-5.6 → gpt-5.6-sol). models.dev catalogs the bare id with sol's rates, so only the multiplier was missing. Pinned in `fast_multipliers_attach_exactly_and_by_prefix`. (ccusage's other alias, gpt-5.3-spark → gpt-5.3-codex-spark, maps to an id with no multiplier in ccusage either, so no entry is needed.)

src/metrics/local_stats.rs
svarlamov-git-ai: Documented in 315b4b6d9: `estimate_cost` now states it is a rough display estimate that keeps Claude-shape cache defaults for every tool, and that the TokenUsage event pipeline (`src/token_usage/cost.rs`) is the authoritative dollars with per-tool cache-rate semantics. Aligning the display estimate's shape per tool is left out deliberately — it would duplicate the per-entry pricing paths for a UI approximation.

src/metrics/models_dev_pricing_snapshot.json
svarlamov-git-ai: Upstream data change, not a trim regression: models.dev itself dropped the `deepseek-chat`/`deepseek-reasoner` alias ids (its current API lists only `deepseek-v4-*`; verified against a fresh fetch). main's next weekly snapshot refresh would pick up the same drop. Usage recorded under a dropped id degrades to unpriced (tokens still counted, cost absent) — the documented behavior for uncataloged models; inventing an alias mapping models.dev no longer publishes isn't something ccusage does either (its coverage comes from its embedded LiteLLM catalog, a source we evaluated and rejected in this stack's design).

PR #2233, opened 2026-08-27
feat(token-usage): extract request speed / Codex service tier per entry

svarlamov-git-ai:
UsageEntry's has_speed dedup marker becomes speed: Option<Speed> plus a
speed_inferred flag, and entries carry which of ccusage's two cost
formulas prices them (pricing_shape), so the cost side can apply fast
multipliers and per-shape tiering. Replacement-policy semantics are
unchanged (a present marker still breaks ties).

Codex now records the service tier: thread_settings_applied events set
a sticky tier exactly like ccusage's parser (an event without a
service_tier says nothing; a present-but-unknown value resets rather
than inheriting stale state; default/standard and fast/priority map as
spelling pairs). Unmarked entries resolve through an injected fallback
read from the rollout's codex-home config.toml (ccusage's auto speed
policy) — resolved by the worker per pass and cached by config mtime,
never per line — else standard, with speed_inferred recording that the
tier did not come from the transcript. Unlike ccusage, resolution
happens at extraction time and is stored, so a later config change
never retroactively reprices history.

The token database's v3 schema is a pre-release rebuild adding the
per-entry pricing dimensions (speed, speed_inferred, the 1h cache-write
split, transcript-vs-catalog cost provenance, and columns for the
long-context decision and catalog id that the tiered cost change will
write). Dropping the cursors makes the next pass re-extract every
transcript under the new rules; dedup keeps that idempotent.

---

### DIFF
--- a/src/token_usage/claude.rs
+++ b/src/token_usage/claude.rs
@@ -8,17 +8,16 @@
 //! - Deduplication and the replacement policy run in the token-usage database
 //!   (entries stream in across runs), via [`should_replace_entry`] and the
 //!   `message_id` fallback; ccusage dedups in memory over whole files.
-//! - Fast-speed entries keep their base model name (no "-fast" suffix or fast
-//!   pricing multiplier); `usage.speed` remains the replacement tie-breaker.
+//! - Fast-speed entries keep their base model name (no "-fast" suffix);
+//!   `usage.speed` carries the fast pricing multiplier and remains the
+//!   replacement tie-breaker.
 //! - Entries whose model is missing or `<synthetic>` are attributed to
 //!   [`UNKNOWN_MODEL`] instead of carrying no model, so tokens aren't lost.
-//! - Long-context tiered pricing is not applied (git-ai's models.dev catalog
-//!   has flat rates); the 1h-ephemeral cache multiplier is (see `cost.rs`).
 
 use serde::Deserialize;
 
 use super::extractor::UsageExtractor;
-use super::types::{TokenCounts, UsageEntry, parse_rfc3339_secs};
+use super::types::{PricingShape, Speed, TokenCounts, UsageEntry, parse_rfc3339_secs};
 
 /// Model recorded for entries with no usable model name.
 pub const UNKNOWN_MODEL: &str = "unknown";
@@ -52,7 +51,7 @@ impl From<&UsageEntry> for ReplacementCandidate {
         Self {
             is_sidechain: entry.is_sidechain,
             token_total: entry.dedupe_token_total(),
-            has_speed: entry.has_speed,
+            has_speed: entry.speed.is_some(),
         }
     }
 }
@@ -103,19 +102,12 @@ struct RawUsage {
     cache_creation_input_tokens: u64,
     #[serde(default)]
     cache_read_input_tokens: u64,
+    /// ccusage `Speed`: unknown values fail the line's parse, matching
+    /// ccusage's strict schema.
     #[serde(default)]
     speed: Option<Speed>,
     #[serde(default)]
     cache_creation: Option<RawCacheCreation>,
-}
-
-/// ccusage `Speed`: unknown values fail the line's parse, matching ccusage's
-/// strict schema.
-#[derive(Deserialize, Clone, Copy)]
-#[serde(rename_all = "lowercase")]
-enum Speed {
-    Standard,
-    Fast,
 }
 
 #[derive(Deserialize, Default, Clone, Copy)]
@@ -230,7 +222,9 @@ fn usage_entry(
             .filter(|cost| cost.is_finite() && *cost >= 0.0)
             .map(super::cost::micro_usd),
         is_sidechain,
-        has_speed: usage.speed.is_some(),
+        speed: usage.speed,
+        speed_inferred: false,
+        pricing_shape: PricingShape::Claude,
     }
 }
 
@@ -447,7 +441,9 @@ mod tests {
         assert_eq!(e.cache_write_1h, 0);
         assert_eq!(e.transcript_cost_micro_usd, None);
         assert!(!e.is_sidechain);
-        assert!(!e.has_speed);
+        assert_eq!(e.speed, None);
+        assert!(!e.speed_inferred);
+        assert_eq!(e.pricing_shape, PricingShape::Claude);
     }
 
     #[test]
@@ -566,7 +562,7 @@ mod tests {
 
         // Tie: speed marker wins.
         let mut with_speed = base.clone();
-        with_speed.has_speed = true;
+        with_speed.speed = Some(Speed::Standard);
         assert!(should_replace_entry(&with_speed, base));
         assert!(!should_replace_entry(base, &with_speed));
         assert!(!should_replace_entry(&base.clone(), base));
@@ -582,9 +578,13 @@ mod tests {
     fn speed_marker_is_extracted() {
         let line = r#"{"timestamp":"2026-01-01T00:00:00Z","message":{"id":"m","model":"x","usage":{"input_tokens":1,"output_tokens":1,"speed":"fast"}},"requestId":"r"}"#;
         let e = &extract(line)[0];
-        assert!(e.has_speed);
+        assert_eq!(e.speed, Some(Speed::Fast));
+        assert!(!e.speed_inferred);
         // Model name keeps its base form (deviation from ccusage's "-fast").
         assert_eq!(e.model, "x");
+
+        let standard = r#"{"timestamp":"2026-01-01T00:00:00Z","message":{"id":"m","model":"x","usage":{"input_tokens":1,"output_tokens":1,"speed":"standard"}},"requestId":"r"}"#;
+        assert_eq!(extract(standard)[0].speed, Some(Speed::Standard));
     }
 
     #[test]
--- a/src/token_usage/claude.rs
+++ b/src/token_usage/claude.rs
@@ -9,8 +9,7 @@
 //!   (entries stream in across runs), via [`should_replace_entry`] and the
 //!   `message_id` fallback; ccusage dedups in memory over whole files.
 //! - Fast-speed entries keep their base model name (no "-fast" suffix);
-//!   `usage.speed` carries the fast pricing multiplier and remains the
-//!   replacement tie-breaker.
+//!   `usage.speed` remains the replacement tie-breaker.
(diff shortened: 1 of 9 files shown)


src/token_usage/claude.rs
```diff
@@ -8,17 +8,16 @@
 //! - Deduplication and the replacement policy run in the token-usage database
 //!   (entries stream in across runs), via [`should_replace_entry`] and the
 //!   `message_id` fallback; ccusage dedups in memory over whole files.
-//! - Fast-speed entries keep their base model name (no "-fast" suffix or fast
-//!   pricing multiplier); `usage.speed` remains the replacement tie-breaker.
+//! - Fast-speed entries keep their base model name (no "-fast" suffix);
+//!   `usage.speed` carries the fast pricing multiplier and remains the
+//!   replacement tie-breaker.
```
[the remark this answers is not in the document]
svarlamov-git-ai: Good catch — fixed in fc00d9f10: this PR's doc now claims only the tie-breaker role, and the multiplier claim moves to the tiered-cost PR (#2234) where the pricing actually lands.

src/token_usage/codex.rs
```diff
@@ -379,10 +426,45 @@ fn make_entry(event: PendingEvent) -> UsageEntry {
...
+        _ => None,
     }
 }
 
+/// Speed fallback implied by a Codex `config.toml`: `Some(Fast)` when any
+/// `service_tier` key holds `fast`/`priority`, comment-stripped and
+/// quote-trimmed (ccusage `codex_config_requests_fast_service_tier`). A
+/// configured `standard` resolves like an absent key — the auto policy
+/// already defaults to standard.
+pub fn config_fallback_speed(config_toml: &str) -> Option<Speed> {
```
svarlamov-git-ai, line 457: **[code-review] `config_fallback_speed` line-scans TOML, so scoped/inert `service_tier` values read as a global fast tier**

This hand parser (`line.split('#').next()` + `split_once('=')`) has no table-header or string awareness:

- `[profiles.turbo]` followed by `service_tier = "fast"` matches even when that profile is not active — while src/streams/model_extraction.rs parses the same file with the `toml` crate and resolves profile-scoped keys properly (`config.get("profiles")?.get(profile)?...`).
- A multi-line TOML string whose body contains `service_tier = "fast"` also matches.

One such config poisons every unmarked entry under that codex home: they bucket as `speed=1` and bill at the model's fast multiplier (2x-6x), inflating `est_cost_micro_usd`, with `speed_inferred=1` the only hint. If byte-for-byte ccusage parity is the goal here, consider documenting that the toml crate was deliberately rejected; otherwise reuse the existing `toml` parsing (active-profile-aware) from model_extraction.rs.

src/token_usage/claude.rs
```diff
@@ -230,7 +221,9 @@ fn usage_entry(
             .filter(|cost| cost.is_finite() && *cost >= 0.0)
             .map(super::cost::micro_usd),
         is_sidechain,
-        has_speed: usage.speed.is_some(),
+        speed: usage.speed,
+        speed_inferred: false,
```
svarlamov-git-ai: **[code-review] `speed_inferred: false` for unmarked Claude entries contradicts the emitted field's documented semantics**

docs/token-usage-events-spec.md defines `speed_inferred` as "any entry's tier was not recorded in the transcript — resolved from configuration or the standard default", and events.rs position 12 says "any entry's tier unrecorded (config/default)". A Claude line with no `usage.speed` is exactly that case — the entry lands in the speed-0 bucket via `COALESCE(speed, 0)` because of the standard default — yet it hardcodes `speed_inferred: false`, while the identical unmarked situation on Codex reports `speed_inferred: true`.

A consumer using the flag to distinguish recorded-standard from assumed-standard data gets opposite answers per tool for the same condition. Either set `speed_inferred: usage.speed.is_none()` here, or narrow the wire docs/spec to say the flag is Codex-only.

src/token_usage/db.rs
```diff
@@ -130,6 +130,80 @@ const MIGRATIONS: &[&str] = &[
 
     INSERT INTO schema_version (version) VALUES (2);
     "#,
+    // Version 3: per-entry pricing dimensions (speed/service tier, tier
```
svarlamov-git-ai, line 133: **[code-review] First destructive migration + no future-version guard makes downgrades fail silently at runtime**

`migrate()` applies nothing when `schema_version` is newer than the binary's `MIGRATIONS` (the loop just skips), and until now that was survivable because shipped v2 was additive. v3/v5 are the first migrations that DROP columns old binaries query: a rolled-back daemon opens the v5 database fine, then every pass errors with "no such column: has_speed" (old main's UPDATE/INSERT/SELECT all reference it), churning processing_errors/backoff per file with no startup diagnostic naming the version mismatch. Blast radius is moderated by `token_usage_metrics` being release-off, but a one-line guard in `migrate()` (error or reset when `current_version > MIGRATIONS.len()`) would turn silent per-pass failure into an explicit signal.

src/token_usage/codex.rs
svarlamov-git-ai: Fixed in 951cae1a1: the fallback now parses config.toml properly (toml crate, same as streams::model_extraction) — the active profile's service_tier wins, else the top-level key, so a fast tier inside an inactive [profiles.*] table no longer marks the whole home fast. Documented as an intentional deviation from ccusage's table-blind line scan; covered by config_fallback_respects_profile_scoping.

src/token_usage/claude.rs
svarlamov-git-ai: Fixed in 951cae1a1: unmarked Claude entries now report speed_inferred=true — the flag uniformly means "tier not recorded in the transcript" across tools.

src/token_usage/db.rs
svarlamov-git-ai: Fixed in 951cae1a1: migrate() now fails closed when the database's schema_version is newer than the binary's migrations, with an actionable message, instead of opening and churning per-pass column errors after a downgrade.

src/token_usage/types.rs
```diff
@@ -46,9 +66,14 @@ pub struct UsageEntry {
     pub transcript_cost_micro_usd: Option<u64>,
     /// Claude sidechain (subagent) entry; loses to non-sidechain duplicates.
     pub is_sidechain: bool,
-    /// Entry carried an explicit `usage.speed` marker; wins ties on
-    /// replacement.
-    pub has_speed: bool,
+    /// Request speed. `None` when the transcript carried no marker (Claude
+    /// without `usage.speed`); a present marker wins ties on replacement.
+    pub speed: Option<Speed>,
```
svarlamov-git-ai, line 71: **[code-review] `speed: Option<Speed>` + `speed_inferred` double-encode "no marker" and push NULL-handling into load-bearing SQL**

Claude constructs `speed_inferred = speed.is_none()` and Codex never produces `None`, so the Option adds no information beyond `speed_inferred` — but persisting the nullable column forces the "unmarked = standard" rule to be restated as `COALESCE(speed, 0)` verbatim at three load-bearing SQL sites (the bucket expression index, `changed_buckets`' GROUP BY, `aggregate_bucket`'s WHERE; SQLite expression-index matching is structural, so any future query writing plain `speed` silently loses the index, and a plain `speed = ?` filter would wrongly drop unmarked-Claude rows from standard buckets), while a fourth site decodes NULL-ness back into marker presence (`speed IS NOT NULL` -> `has_speed`). Simpler: resolve to a NOT NULL `speed` at insert time (plain column index; v1's `has_speed` shows the precedent) and derive marker presence from `NOT speed_inferred`. That swap is drop-in for pricing/bucketing; it only changes the `should_replace` tie-break for Codex entries, which is unreachable today (`message_id: None`, content-derived keys) — and the changed direction better matches the documented "a present marker wins ties" rule.

src/token_usage/codex.rs
```diff
@@ -379,10 +426,45 @@ fn make_entry(event: PendingEvent) -> UsageEntry {
...
+
+/// Speed fallback implied by a Codex `config.toml`: the `service_tier` of
+/// the active profile (`profiles.<profile>.service_tier`, matching Codex's
+/// own config resolution and `streams::model_extraction`'s model lookup),
+/// falling back to the top-level key. Only `fast`/`priority` produce a
+/// fallback — a configured `standard` resolves like an absent key, since
+/// the auto policy already defaults to standard. Deviation from ccusage,
+/// whose table-blind line scan would mark the whole home fast for a
+/// `service_tier = "fast"` inside an *inactive* profile table.
+pub fn config_fallback_speed(config_toml: &str) -> Option<Speed> {
```
svarlamov-git-ai, line 457: **[code-review] Duplicates the codex config profile-resolution chain from `streams::model_extraction`**

`config_fallback_speed` re-implements the active-profile-else-top-level TOML resolution that `extract_model_from_codex_config` (src/streams/model_extraction.rs) already contains for the `model` key — the same `config.get("profile") -> config.get("profiles")?.get(profile)?.get(<key>)` chain with top-level fallback, each with its own `toml::from_str` parse of the same file. Any fix to profile-resolution semantics (or a Codex config format change) now needs the same edit in both places. A shared `codex_config_profile_value(config: &toml::Value, key: &str)` helper in `streams::model_extraction` — which the worker already reaches for `read_codex_config`/`codex_home_from_transcript_path` — keeps one copy.

src/token_usage/types.rs
svarlamov-git-ai: Considered and kept: NULL-ness is the replacement-policy marker (ccusage's `has_speed` — an entry with a recorded tier beats one without), and `speed_inferred` is a *bucket-level* wire fact aggregated with MAX, so collapsing them changes replacement semantics, not just storage. The COALESCE coupling is real but confined to the three aggregation sites, all covered by the expression index and pinned by the fingerprint/index tests; a NOT NULL rework would touch every extractor and the wire aggregation for a shape win we'd rather take post-1.0 if at all.

src/token_usage/codex.rs
svarlamov-git-ai: Same conclusion as the earlier review passes (this was raised and answered there too): a naive shared helper changes edge behavior — `extract_model_from_codex_config` tolerates malformed profile tables differently and returns the raw string, while the speed path needs the exact-string tier mapping with unknown-resets. The duplication is two ~15-line functions against a documented ccusage-ported spec; unifying them couples two call sites that evolve with different upstreams (codex config schema vs ccusage speed detection).

PR #2234, opened 2026-08-27
feat(token-usage): price entries with long-context tiers and fast multipliers

svarlamov-git-ai:
cost.rs now implements both of ccusage's catalog cost formulas, selected
by the entry's pricing shape:

- Claude (calculate_cost_from_pricing): whole-request tier selection —
  once uncached input + cache reads + cache writes exceed the model's
  threshold (strict), every rate switches to its above-threshold value,
  falling back per rate to the base; 1h cache writes bill at 2x the
  selected tier's input rate. Unpublished cache rates keep the
  0.1x/1.25x input-derived defaults.
- Codex (calculate_codex_bucket_cost, collapsed per entry): the tier is
  selected by the raw per-turn input (uncached + cached; output never
  selects), and an unpublished cache-read rate bills at the full input
  rate of the selected tier rather than the Claude default. No cache
  write terms (Codex reports none).

Fast-speed entries bill the whole request at the model's fast
multiplier on top of either formula (ccusage prices the fast bucket as
standard + cost x (multiplier - 1); identical per entry). Transcript
costUSD still wins outright and the $10k clamp stays last.

price_entry() replaces the bare cost lookup at upsert time and also
yields the request's tier decision and the id of the catalog that
priced it (embedded:<version> or modelsdev:<content-hash>); both are
stored on the entry row so bucket emission can carry the token splits
and provenance a different pricing sheet needs to recompute cost.
Recomputed costs change bucket fingerprints, so affected buckets
re-emit through the existing reconciliation.

The marginal-at-200K arm ccusage keeps for LiteLLM above_200k data
without a threshold is documented as unreachable (models.dev tiers
always carry one) and intentionally not ported.

---

### DIFF
--- a/src/token_usage/cost.rs
+++ b/src/token_usage/cost.rs
@@ -59,7 +59,10 @@ pub fn price_entry(entry: &UsageEntry) -> EntryPricing {
         };
     };
     EntryPricing {
-        cost_micro_usd: Some(cost_from_pricing(entry, &pricing)),
+        // The same tier decision selects the billed rates AND lands in the
+        // recorded long_context column the wire splits are built from, so
+        // the two can never diverge.
+        cost_micro_usd: Some(tiered_cost_from_pricing(entry, &pricing, long_context)),
         long_context,
         pricing_catalog: Some(pricing_catalog_id()),
     }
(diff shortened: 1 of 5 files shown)


src/token_usage/cost.rs
```diff
@@ -22,33 +23,155 @@ const CACHE_WRITE_1H_INPUT_MULTIPLIER: f64 = 2.0;
...
+    /// Catalog in effect when the cost was computed; `None` when the cost
+    /// came from the transcript or the model has no pricing.
+    pub pricing_catalog: Option<&'static str>,
+}
+
+/// Price one entry (ccusage "auto" mode): the transcript's own `costUSD`
+/// wins, otherwise the catalog formula for the entry's [`PricingShape`],
+/// with fast-speed entries billing the whole request at the model's fast
+/// multiplier.
+pub fn price_entry(entry: &UsageEntry) -> EntryPricing {
```
svarlamov-git-ai, line 44: **[code-review] The long-context decision is computed twice; pass one decision down to keep the emitted splits provably consistent with the billed cost**

`price_entry` computes `is_long_context(entry, pricing)` for the recorded `long_context` flag, then `cost_from_pricing` -> `claude_cost_usd`/`codex_cost_usd` each recompute it to pick rates. The emitted long-context token splits and the actual billed tier are only consistent because the same pure function runs twice on the same inputs — the repricing contract these fields exist for deserves a structural guarantee. Computing the tier once in `price_entry` and passing the bool into `cost_from_pricing` removes the duplicate work and makes divergence impossible.

svarlamov-git-ai: Fixed in 1d5150b3a: price_entry computes the tier decision once and threads it into the cost formulas, so the recorded long_context flag (which the wire splits are built from) can never diverge from the tier the cost billed at. The test-only wrapper that recomputes it is #[cfg(test)].

src/token_usage/cost.rs
```diff
@@ -22,33 +23,158 @@ const CACHE_WRITE_1H_INPUT_MULTIPLIER: f64 = 2.0;
...
+}
+
+/// The billed rate for one token class: the above-threshold rate in a
+/// long-context request (falling back per rate to the base), the base rate
+/// otherwise. Shared by both cost shapes so the fallback rule cannot drift.
+fn tier_rate(long_context: bool, base: f64, above: Option<f64>) -> f64 {
+    if long_context {
+        above.unwrap_or(base)
+    } else {
+        base
```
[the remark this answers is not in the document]
svarlamov-git-ai: Intentional, matching ccusage and current source data: ccusage's own `ModelPricing` carries a single above-threshold band (its models.dev loader reads one `above_200k`-style tier), and no model on the 8-provider allowlist publishes more than one context tier in models.dev today — the multi-band shape has no real data to exercise it. A second band would extend `ModelPricing` and the wire splits (repricing contract), which is follow-up material if a vendor ever publishes one; pricing would degrade gracefully until then (everything above the first threshold bills at that band's rates, exactly like ccusage).

PR #2235, opened 2026-08-27
feat(token-usage): speed-keyed buckets and repricing fields on the wire

svarlamov-git-ai:
Fast and standard usage of one model bill at different rates, so speed
(0 standard, 1 fast) joins the bucket identity: bucket_state is keyed by
(session_id, model, speed, bucket_ts) (schema v4 rebuilds just that
table; entries and cursors survive) and aggregation groups by the
entry's speed, with unmarked entries falling into the standard bucket.

TokenUsage events gain positions 11-19 so the server can recompute a
bucket's cost under a different pricing sheet: speed and whether it was
inferred from configuration, the 1h cache-write split, the long-context
token splits per class (tokens of requests that selected their model's
long-context tier; whole-request selection already applied client-side,
base-tier tokens are the totals minus these), and the transcript-priced
portion of the cost, which is fixed under repricing (its tokens remain
in the totals, so such buckets reprice approximately). The pricing
catalog id rides custom_attributes on catalog-priced buckets.

The bucket fingerprint covers every new emitted value, so any split
changing re-emits the bucket at a higher revision; the catalog id alone
is deliberately excluded (an id change without numeric change means
identical rates). The spec's value table, bucket identity, and pricing
sections are updated accordingly.

---

### DIFF
--- a/src/daemon/token_usage_worker.rs
+++ b/src/daemon/token_usage_worker.rs
@@ -1001,8 +1001,9 @@ fn emit_changed_buckets(
             attrs = attrs.repo_url(url.clone());
         }
         if let Some(catalog) = &aggregate.pricing_catalog {
-            attrs = attrs
-                .custom_attributes(serde_json::json!({ "pricing_catalog": catalog }).to_string());
+            // A dedicated attribute: custom_attributes stays reserved for
+            // the org-configured map every event type carries.
+            attrs = attrs.pricing_catalog(catalog.clone());
         }
         events.push(MetricEvent::new(&values, attrs.to_sparse()));
     }
@@ -1569,16 +1570,21 @@ mod tests {
         assert_eq!(value_u64(standard, token_usage_pos::SPEED), Some(0));
         assert_eq!(
             value_u64(standard, token_usage_pos::SPEED_INFERRED),
-            Some(0)
+            Some(1),
+            "no usage.speed marker was recorded on the standard entry"
         );
         assert_eq!(
             value_u64(standard, token_usage_pos::TRANSCRIPT_COST_MICRO_USD),
             Some(0)
         );
-        // Catalog-priced: the pricing catalog id rides custom_attributes.
+        // Catalog-priced: the pricing catalog id rides its dedicated
+        // attribute (custom_attributes stays free for the org map).
         let attrs = EventAttributes::from_sparse(&standard.attrs);
-        let custom = attrs.custom_attributes.unwrap().unwrap();
-        assert!(custom.contains("pricing_catalog"), "{custom}");
+        assert_eq!(
+            attrs.pricing_catalog.flatten().as_deref(),
+            Some(crate::metrics::model_pricing::pricing_catalog_id())
+        );
+        assert_eq!(attrs.custom_attributes, None);
 
         let fast = &events[1];
         assert_eq!(value_u64(fast, token_usage_pos::SPEED), Some(1));
@@ -1600,7 +1606,7 @@ mod tests {
             Some(500_000)
         );
         let attrs = EventAttributes::from_sparse(&fast.attrs);
-        assert!(attrs.custom_attributes.flatten().is_none());
+        assert!(attrs.pricing_catalog.flatten().is_none());
     }
 
     #[test]
--- a/src/daemon/token_usage_worker.rs
+++ b/src/daemon/token_usage_worker.rs
@@ -963,6 +963,10 @@ fn emit_changed_buckets(
         .collect();
     token_db.reserve_emit_seqs(&identity.session_id, &reservations)?;
     let repo_url = resolve_repo_url();
+    // The org-configured attribute map every event type carries (the
+    // checkpoint and session emitters attach the same one).
+    let config = crate::config::Config::fresh();
+    let custom_attributes = config.custom_attributes();
     let mut events = Vec::with_capacity(changed.len());
     for (bucket, next_seq) in &changed {
         let aggregate = &bucket.aggregate;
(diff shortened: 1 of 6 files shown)


src/daemon/token_usage_worker.rs
```diff
@@ -984,6 +1000,10 @@ fn emit_changed_buckets(
         if let Some(url) = &repo_url {
             attrs = attrs.repo_url(url.clone());
         }
+        if let Some(catalog) = &aggregate.pricing_catalog {
+            attrs = attrs
+                .custom_attributes(serde_json::json!({ "pricing_catalog": catalog }).to_string());
```
svarlamov-git-ai: **[code-review] `pricing_catalog` hijacks the org-configured `custom_attributes` channel**

Every other event emitter populates `custom_attributes` from the user/org-configured map (`Config::custom_attributes()` — see src/daemon/rewrite_metrics.rs:213-215, src/daemon/checkpoint.rs, src/authorship/post_commit.rs), so the wire position has established semantics: "the org's configured attribute map". This line writes a synthetic `{"pricing_catalog": ...}` object into the same field instead. Consequences: TokenUsage events can never carry the configured org attributes (org-attribute filtering/reporting over cost silently excludes them), and a server consumer parsing the field as the org map sees a fabricated `pricing_catalog` key that collides outright with a user-defined attribute of the same name.

Consider merging the catalog id into the configured map (like `apply_rewrite_metric_custom_attributes` does for rewrite metrics), or giving the catalog id its own attribute.

src/token_usage/db.rs
```diff
@@ -550,65 +602,55 @@ impl TokenUsageDatabase {
...
-                    COALESCE(SUM(cost_micro_usd), 0),
-                    COUNT(*)
+        let mut stmt = conn.prepare(&format!(
+            "SELECT model, COALESCE(speed, 0), bucket_ts, {AGGREGATE_COLUMNS}
              FROM usage_entries
              WHERE session_id = ?1
-             GROUP BY model, bucket_ts",
-        )?;
-        let aggregates: Vec<(String, u32, BucketAggregate)> = stmt
+             GROUP BY model, COALESCE(speed, 0), bucket_ts",
```
svarlamov-git-ai, line 653: **[code-review] `GROUP BY model, COALESCE(speed, 0), bucket_ts` forces a temp b-tree sort on every `changed_buckets` call**

Verified with EXPLAIN QUERY PLAN on the v5 schema: the old `GROUP BY model, bucket_ts` streamed off `idx_usage_entries_bucket (session_id=?)`; the new query adds `USE TEMP B-TREE FOR GROUP BY`. This runs after every processed file pass and reconcile. Note when fixing: a plain `(session_id, model, speed, bucket_ts)` index does NOT restore streaming (SQLite can't prove COALESCE order); either use an expression index on `(session_id, model, COALESCE(speed, 0), bucket_ts)` or `GROUP BY model, speed, bucket_ts` and coalesce NULL to 0 in Rust. Materiality is modest (~10ms at 20k rows, background thread), but the fix is one line.

src/daemon/token_usage_worker.rs
svarlamov-git-ai: Fixed in 24069befb: the catalog id now rides a dedicated attribute position (31) and custom_attributes stays reserved for the org-configured map. The server ingest was updated in lockstep to read the new attribute.

src/token_usage/db.rs
svarlamov-git-ai: Fixed in 24069befb: v4 adds an expression index matching the GROUP BY exactly — (session_id, model, COALESCE(speed, 0), bucket_ts) — and the v5 rebuild carries it, so reconciliation streams off the index again instead of a temp B-tree.

src/metrics/events.rs
```diff
@@ -2358,18 +2358,35 @@ pub mod token_usage_pos {
...
-/// model attribute is the bucket's model.
+/// One event per (session_id, model, speed, bucket_ts): the deduplicated
+/// token usage and estimated cost of a 5-minute UTC bucket. The server
+/// upserts on that key keeping the highest emitted_seq (a strictly
+/// increasing per-bucket revision), so a bucket is re-emitted whenever its
+/// aggregate changes (including corrections downward and drops to zero) and
+/// same-second re-emissions cannot tie on the u32-second event_ts. Uses
+/// EventAttributes for standard metadata (repo_url, tool, model, session
+/// ids, etc.); the model attribute is the bucket's model, and the pricing
+/// catalog id rides custom_attributes.
```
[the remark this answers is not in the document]
svarlamov-git-ai: Good catch — fixed in 3d2223df2; the doc now names the dedicated pricing_catalog attribute.

src/daemon/token_usage_worker.rs
```diff
@@ -984,6 +1000,11 @@ fn emit_changed_buckets(
         if let Some(url) = &repo_url {
             attrs = attrs.repo_url(url.clone());
         }
+        if let Some(catalog) = &aggregate.pricing_catalog {
+            // A dedicated attribute: custom_attributes stays reserved for
+            // the org-configured map every event type carries.
+            attrs = attrs.pricing_catalog(catalog.clone());
```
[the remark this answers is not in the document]
svarlamov-git-ai: Fixed (commit on this branch): `emit_changed_buckets` now attaches the org-configured map via `custom_attributes_map(Config::fresh().custom_attributes())`, same as the checkpoint and session emitters; the e2e `fast_long_context_codex_usage_emits_tier_and_speed_fields` now pins attr 30 end-to-end through the daemon (config patch → emitted event).

PR #2236, opened 2026-08-27
feat(token-usage): match forked Codex sessions against their parent's usage prefix

svarlamov-git-ai:
A forked rollout replays its parent's history with timestamps rewritten
to the fork instant. The rewritten-burst heuristic (skip leading events
spaced <= 1s apart) misidentifies a real fork whose own first turn
follows the replay after a longer pause — observed in the wild as a
16,508-token replayed event counted as genuine usage 6.975s before the
child's first own turn.

Port ccusage's parent-prefix matching into the incremental extractor:
a fork now parks in AwaitingParent until the worker resolves the parent
rollout and answers with its pre-fork usage signatures (per-event
deltas, timestamps excluded; collected without replay filtering so a
prefix the parent itself replayed stays included, truncated at the
child's session_meta timestamp). The child's leading events consume the
matching front of that prefix; a mismatch after any match means the
replayed history ended, and a mismatch with nothing matched — or an
unresolvable parent — falls back to the burst heuristic exactly as
before (never retried, matching ccusage's unavailable-parent path).
The unmatched remainder persists in the extractor state, so the parent
is read once per fork, not per pass or restart.

The worker serves prefix requests after restoring state and after every
consumed line. Parent resolution tries files tracked under the parent's
external session id, then a bounded scan of the codex sessions tree,
always confirming a candidate by its recorded session_meta id (ccusage
locates parents by id, never filename) — a bounded single-file read in
the async worker.

The spec's fork section is rewritten and the split-burst accepted
limitation narrows to the unresolvable-parent fallback path.

---

### DIFF
--- a/src/daemon/token_usage_worker.rs
+++ b/src/daemon/token_usage_worker.rs
@@ -792,10 +792,10 @@ fn resolve_parent_prefix(
         .find(|path| records_session_id(path, &request.parent_id))
         .or_else(|| scan_sessions_tree(Path::new(child_path), &request.parent_id))?;
     let file = std::fs::File::open(&parent).ok()?;
-    Some(crate::token_usage::codex::collect_parent_prefix(
+    crate::token_usage::codex::collect_parent_prefix(
         BufReader::with_capacity(128 * 1024, file),
         request.fork_ts_ms,
-    ))
+    )
 }
 
 /// Whether the rollout's first line records `session_id` as its own id.
@@ -812,18 +812,35 @@ fn records_session_id(path: &Path, session_id: &str) -> bool {
 }
 
 /// Fallback parent discovery for rollouts the token database has not tracked
-/// (yet): walk the child's enclosing codex `sessions` tree — and its sibling
-/// `archived_sessions`, where Codex moves archived rollouts — for a filename
-/// carrying the parent id (codex rollout filenames embed the session uuid),
-/// confirming by the recorded `session_meta` id.
+/// (yet): walk the codex home's `sessions` and `archived_sessions` trees for
+/// a filename carrying the parent id (codex rollout filenames embed the
+/// session uuid), confirming by the recorded `session_meta` id. The home is
+/// resolved like `streams::model_extraction` (configured home, else the
+/// nearest `.codex` ancestor) so archived or custom-layout children resolve
+/// too; a `sessions`/`archived_sessions` ancestor is the fallback anchor.
 fn scan_sessions_tree(child_path: &Path, parent_id: &str) -> Option<PathBuf> {
-    let sessions_root = child_path
-        .ancestors()
-        .find(|dir| dir.file_name().and_then(|name| name.to_str()) == Some("sessions"))?;
-    let mut pending = vec![sessions_root.to_path_buf()];
-    if let Some(codex_home) = sessions_root.parent() {
-        pending.push(codex_home.join("archived_sessions"));
-    }
+    let roots: Vec<PathBuf> =
+        match crate::streams::model_extraction::codex_home_from_transcript_path(child_path) {
+            Some(home) => vec![home.join("sessions"), home.join("archived_sessions")],
+            None => {
+                let anchor = child_path.ancestors().find(|dir| {
+                    matches!(
+                        dir.file_name().and_then(|name| name.to_str()),
+                        Some("sessions" | "archived_sessions")
+                    )
+                })?;
+                let parent = anchor.parent();
+                vec![
+                    anchor.to_path_buf(),
+                    match parent {
+                        Some(dir) if anchor.ends_with("sessions") => dir.join("archived_sessions"),
+                        Some(dir) => dir.join("sessions"),
+                        None => return None,
+                    },
+                ]
+            }
+        };
+    let mut pending = roots;
     let mut visited = 0usize;
     while let Some(dir) = pending.pop() {
         let Ok(entries) = std::fs::read_dir(&dir) else {
--- a/src/daemon/token_usage_worker.rs
+++ b/src/daemon/token_usage_worker.rs
@@ -782,6 +782,14 @@ fn resolve_parent_prefix(
     child_path: &str,
     request: &ParentPrefixRequest,
 ) -> Option<Vec<UsageSignature>> {
+    // The trait hook is provider-neutral, but everything below is codex
+    // machinery (rollout layout, session_meta confirmation, codex delta
+    // parsing). Another tool's request resolves to "parent unavailable"
+    // (its burst fallback) instead of misparsing a foreign transcript as a
+    // rollout; extend with per-tool dispatch when a second tool needs it.
+    if tool != "codex" {
+        return None;
+    }
     let tracked = token_db
         .stream_paths_for_external_session(&request.parent_id, tool)
         .unwrap_or_default();
@@ -800,13 +808,20 @@ fn resolve_parent_prefix(
 
 /// Whether the rollout's first line records `session_id` as its own id.
 fn records_session_id(path: &Path, session_id: &str) -> bool {
+    use std::io::Read;
     let Ok(file) = std::fs::File::open(path) else {
         return false;
     };
+    // A rollout's session_meta line is small; a candidate whose first line
+    // is larger is not one, and must not be buffered wholly into memory
+    // (the scan feeds this any file whose name carries the parent id).
+    const MAX_FIRST_LINE_BYTES: u64 = 64 * 1024;
     let mut line = Vec::new();
-    if read_line_bytes(&mut BufReader::new(file), &mut line).is_err() {
+    let mut reader = BufReader::new(file.take(MAX_FIRST_LINE_BYTES));
+    if read_line_bytes(&mut reader, &mut line).is_err() {
         return false;
     }
+    // A line truncated at the cap is not valid JSON and fails the id parse.
     crate::token_usage::codex::session_meta_id(&String::from_utf8_lossy(&line)).as_deref()
         == Some(session_id)
 }
@@ -859,7 +874,7 @@ fn scan_sessions_tree(child_path: &Path, parent_id: &str) -> Option<PathBuf> {
             } else if path
                 .file_name()
                 .and_then(|name| name.to_str())
-                .is_some_and(|name| name.contains(parent_id))
+                .is_some_and(|name| name.contains(parent_id) && name.ends_with(".jsonl"))
                 && path != child_path
                 && records_session_id(&path, parent_id)
             {
--- a/src/daemon/token_usage_worker.rs
+++ b/src/daemon/token_usage_worker.rs
@@ -55,6 +55,7 @@ use crate::error::GitAiError;
 use crate::metrics::{EventAttributes, MetricEvent, PosEncoded, TokenUsageValues};
 use crate::streams::db::{StreamRecord, StreamsDatabase};
 use crate::token_usage::db::{BatchCommit, TokenUsageDatabase, TrackedFile};
+use crate::token_usage::extractor::{ParentPrefixRequest, UsageSignature};
 use crate::token_usage::{Speed, extractor_for_tool};
 
 /// One raw JSONL line read as bytes: unlike UTF-8-strict `read_line`, a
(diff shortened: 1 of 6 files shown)


src/token_usage/codex.rs
```diff
@@ -225,6 +249,34 @@ impl UsageExtractor for CodexUsageExtractor {
...
+                fork_ts_ms,
+            } => Some(ParentPrefixRequest {
+                parent_id: parent_id.clone(),
+                fork_ts_ms: *fork_ts_ms,
+            }),
+            _ => None,
+        }
+    }
+
+    fn provide_parent_prefix(&mut self, prefix: Option<Vec<UsageSignature>>) {
```
svarlamov-git-ai, line 273: **[code-review] `Some(vec![])` — a resolved parent with zero pre-fork usage — takes the burst fallback and can drop the child's real leading turns**

With an empty prefix, `MatchingParent { remaining: [], matched: false }` means the child's first real delta mismatches (`remaining.front()` is `None`) and is parked as `AwaitingSecond { pending: Some(real event) }`; a second usage event within 1s transitions to `SkippingBurst` and both real events (plus any chain spaced <= 1s) are permanently dropped, and even the single-turn case defers emission to the wall-clock flush. But an empty prefix is the one case where the parent rollout was resolved and provably contained nothing to replay — `ReplayState::Done` is strictly safer than the burst heuristic here. Trigger: a fork/`thread_spawn` taken before the parent's first `token_count` (collect_parent_prefix truncates at `fork_ts_ms` and returns `[]`). The comment claims ccusage parity, but parity with a heuristic is a weak reason to skip usage we can prove is genuine.

src/daemon/token_usage_worker.rs
```diff
@@ -761,6 +762,97 @@ fn codex_config_fallback_speed(stream_path: &Path) -> Option<Speed> {
...
+    crate::token_usage::codex::session_meta_id(&String::from_utf8_lossy(&line)).as_deref()
+        == Some(session_id)
+}
+
+/// Fallback parent discovery for rollouts the token database has not tracked
+/// (yet): walk the child's enclosing codex `sessions` tree — and its sibling
+/// `archived_sessions`, where Codex moves archived rollouts — for a filename
+/// carrying the parent id (codex rollout filenames embed the session uuid),
+/// confirming by the recorded `session_meta` id.
+fn scan_sessions_tree(child_path: &Path, parent_id: &str) -> Option<PathBuf> {
```
svarlamov-git-ai, line 838: **[code-review] Anchor on an ancestor literally named `sessions` cannot serve archived/relocated children, and re-encodes Codex's on-disk layout**

`child_path.ancestors().find(|dir| ... == Some("sessions"))?` returns `None` for any child path without a `sessions` ancestor (e.g. a rollout processed from `archived_sessions`, or a custom layout), so the parent is unresolvable and the extractor permanently falls back to the burst heuristic — re-opening the replayed-history overcount this PR eliminates. Today the trigger is narrow (checkpoints fire on live rollouts under `sessions/`), but the sibling knowledge ("archived_sessions" lives next to "sessions", filenames embed the session uuid) is now encoded here *and* in `CodexAgent::find_rollout_path_for_session_in_home` / model_extraction — three copies that can drift. Anchoring via `codex_home_from_transcript_path` (already imported in this file, and it handles archived paths) instead of the literal `sessions` ancestor would remove both the gap and one copy of the layout knowledge.

src/token_usage/codex.rs
```diff
@@ -432,6 +509,119 @@ fn make_entry(event: PendingEvent) -> UsageEntry {
...
+/// replay filtering, so a prefix the parent itself replayed from *its*
+/// parent stays included, exactly as the child's copy contains it. Events
+/// whose timestamps don't parse are dropped, mirroring extraction (Codex
+/// never emits such lines; ccusage keeps a present-but-malformed timestamp's
+/// usage in the prefix — a deviation only reachable through corruption, and
+/// note the child's replayed *copy* of such an event is rewritten to the
+/// fork instant, so its delta would go unmatched either way here);
+/// collection stops at the first usage event past `fork_ts_ms` (usage the
+/// parent recorded after the fork was never replayed).
+pub fn collect_parent_prefix(
```
svarlamov-git-ai, line 589: **[code-review] No bound on the parent prefix: whole-parent read plus an unbounded signature deque persisted in state_json**

`collect_parent_prefix` reads the parent rollout to the fork instant and returns one `UsageSignature` per non-zero delta, and the unmatched remainder persists as a serialized `VecDeque` in `tracked_files.state_json` on every batch commit while the child is in `MatchingParent`. A fork taken at the end of a months-long parent (tens of thousands of turns) means a full parent-file read on the notification path plus a multi-MB state_json row rewritten per pass until the replay resolves. `PARENT_SCAN_MAX_ENTRIES` caps directory scanning but nothing caps this. A cap on collected signatures (falling back to the burst heuristic beyond it, like an unresolvable parent) would bound both.

src/token_usage/codex.rs
svarlamov-git-ai: Fixed in 7dc11c39d: a successfully read parent with zero pre-fork usage now goes straight to Done — nothing was replayed, so the child's real back-to-back first turns count instead of being parked by the burst heuristic. Documented as an intentional deviation from ccusage, which shares the fallback hole; the heuristic stays reserved for genuinely unresolvable parents.

src/daemon/token_usage_worker.rs
svarlamov-git-ai: Fixed in 7dc11c39d: scan roots now come from the codex home via the shared codex_home_from_transcript_path resolution (covering sessions and archived_sessions from either side, so an archived child can resolve an untracked parent), with the ancestor-name anchor kept only as the fallback for homes that can't be resolved.

src/token_usage/codex.rs
svarlamov-git-ai: Fixed in 7dc11c39d: collect_parent_prefix is capped at 50k signatures — beyond that the parent counts as unresolvable (burst fallback) instead of bloating every state_json write. The whole-parent read itself matches ccusage and stays a one-time bounded read per fork (the shrinking remainder persists in state, so it is never re-read across passes).

src/token_usage/codex.rs
```diff
@@ -432,6 +513,130 @@ fn make_entry(event: PendingEvent) -> UsageEntry {
...
+    fork_ts_ms: i64,
+) -> Option<Vec<UsageSignature>> {
+    let mut prefix = Vec::new();
+    let mut prev_totals: Option<CodexTotals> = None;
+    let mut line: Vec<u8> = Vec::new();
+    loop {
+        line.clear();
+        let Ok(bytes) = reader.read_until(b'\n', &mut line) else {
+            return Some(prefix);
+        };
```
svarlamov-git-ai, line 604: **[code-review] Mid-read I/O error is treated as a complete parent prefix**

`collect_parent_prefix` maps a mid-read I/O error to the same arm as EOF, returning `Some(prefix)` — a "complete" (possibly empty) pre-fork prefix. `provide_parent_prefix` then treats `Some(vec![])` as proof nothing was replayed (`ReplayState::Done`, whose contract is a *successfully read* parent), so the fork's replayed history is emitted as the child's own usage; an error later in the file truncates the prefix, and the replay tail past the truncation point is likewise counted once `MatchingParent` is exhausted with `matched: true`. This is asymmetric with the open failure in `resolve_parent_prefix` (`File::open(&parent).ok()?`), which correctly yields `None` and falls back to the rewritten-burst heuristic — and the wrong `Done`/short-prefix state persists via `state_json` and is never retried (`parent_request` only fires from `AwaitingParent`).

Trigger: EIO / stale NFS handle / cloud-placeholder eviction between `records_session_id`'s first-line probe and this second read, or a device error mid-file. Effect: replayed tokens double-count in both token totals and est_cost.

```suggestion
        let Ok(bytes) = reader.read_until(b'\n', &mut line) else {
            return None;
        };
```

(Also update the doc comment above, which currently says `None` is only for the `MAX_PARENT_PREFIX_EVENTS` cap.)

src/token_usage/codex.rs
```diff
@@ -182,8 +203,11 @@ impl UsageExtractor for CodexUsageExtractor {
         };
         match raw.entry_type.as_deref() {
             Some("session_meta") => {
-                if raw.payload.as_ref().is_some_and(is_forked_session) {
-                    self.state.replay = ReplayState::AwaitingFirst;
+                if let Some(parent_id) = raw.payload.as_ref().and_then(fork_parent_id) {
+                    self.state.replay = ReplayState::AwaitingParent {
```
svarlamov-git-ai, line 215: **[code-review] Repeated fork markers re-arm `AwaitingParent`, discarding replay progress and re-triggering full parent resolution**

This assignment is unconditional: any later `session_meta` line carrying a fork marker overwrites whatever replay state exists (`Done`, mid-`MatchingParent`, burst arms). Real Codex writes exactly one `session_meta` (line 1), so this needs corrupt/crafted input — but then (a) each of N fork-marker lines costs one full, memoization-free parent resolution in the worker (DB lookup, candidate first-line probes, sessions-tree walk up to the 50k-entry cap, whole parent-file read), unthrottled on notify-origin passes; and (b) each re-arm resets matching, so real usage events following a spurious mid-file fork marker can be mis-classified as replayed history and dropped. One-shot per file (the byte cursor consumes the lines), so self-DoS only; cheap hardening is to re-arm only from the initial state (`ReplayState::Done` with no prior fork seen) or to memoize resolution per `(parent_id, fork_ts_ms)` within a pass.

src/daemon/token_usage_worker.rs
```diff
@@ -761,6 +762,114 @@ fn codex_config_fallback_speed(stream_path: &Path) -> Option<Speed> {
...
+        .or_else(|| scan_sessions_tree(Path::new(child_path), &request.parent_id))?;
+    let file = std::fs::File::open(&parent).ok()?;
+    crate::token_usage::codex::collect_parent_prefix(
+        BufReader::with_capacity(128 * 1024, file),
+        request.fork_ts_ms,
+    )
+}
+
+/// Whether the rollout's first line records `session_id` as its own id.
+fn records_session_id(path: &Path, session_id: &str) -> bool {
```
svarlamov-git-ai, line 810: **[code-review] First-line probe buffers a newline-free file wholly into memory; scan has no filename/size filter**

`records_session_id`'s probe reads the first line with an uncapped buffer: for a file with no `\n`, `read_until`/`read_line_bytes` returns the entire file as one "line" (plus a `from_utf8_lossy` copy, ~2x transient RSS) before the `session_meta` check can reject it. Unlike the existing rollout discovery in `src/streams/agents/codex.rs` (requires the `rollout-` prefix and `.jsonl` extension), `scan_sessions_tree` feeds this probe any file under `sessions/`/`archived_sessions` whose name merely contains the parent uuid, with no size gate — so one large newline-free file there spikes daemon memory during fork resolution. Cheap hardening: apply the same `rollout-*.jsonl` filename filter in the scan, and/or cap the probe with `Read::take(64 * 1024)` — a real `session_meta` first line is far smaller.

src/daemon/token_usage_worker.rs
```diff
@@ -761,6 +762,114 @@ fn codex_config_fallback_speed(stream_path: &Path) -> Option<Speed> {
...
+}
+
+/// Fallback parent discovery for rollouts the token database has not tracked
+/// (yet): walk the codex home's `sessions` and `archived_sessions` trees for
+/// a filename carrying the parent id (codex rollout filenames embed the
+/// session uuid), confirming by the recorded `session_meta` id. The home is
+/// resolved like `streams::model_extraction` (configured home, else the
+/// nearest `.codex` ancestor) so archived or custom-layout children resolve
+/// too; a `sessions`/`archived_sessions` ancestor is the fallback anchor.
+fn scan_sessions_tree(child_path: &Path, parent_id: &str) -> Option<PathBuf> {
```
svarlamov-git-ai, line 838: **[code-review] Third codex sessions-tree walker; consolidate with `streams::agents::codex`**

`scan_sessions_tree` is now the third implementation that encodes the codex `sessions/` + `archived_sessions/` rollout layout, next to `CodexAgent::find_rollout_path_for_session_in_home` and `scan_rollout_recursive` in `src/streams/agents/codex.rs` (the existing helper covers `archived_sessions` too). Direct reuse wasn't possible — this site needs codex-home anchoring from the child path, recorded `session_meta`-id confirmation with child-path exclusion, and the 50k-entry cap — but the layout and filename knowledge can now drift across three copies (filename: `rollout-*<id>*.jsonl` there vs any name containing the id here; selection: newest-by-mtime vs first-confirmed). Consider one bounded walker in `streams::agents::codex`, parameterized by filename predicate + confirmation callback, shared by all three (CLAUDE.md rule 5).

src/daemon/token_usage_worker.rs
```diff
@@ -761,6 +762,114 @@ fn codex_config_fallback_speed(stream_path: &Path) -> Option<Speed> {
...
+) -> Option<Vec<UsageSignature>> {
+    let tracked = token_db
+        .stream_paths_for_external_session(&request.parent_id, tool)
+        .unwrap_or_default();
+    let parent = tracked
+        .into_iter()
+        .map(PathBuf::from)
+        .filter(|path| path != Path::new(child_path))
+        .find(|path| records_session_id(path, &request.parent_id))
+        .or_else(|| scan_sessions_tree(Path::new(child_path), &request.parent_id))?;
```
svarlamov-git-ai, line 801: **[code-review] Provider-neutral trait hooks answered with unconditionally codex machinery**

`parent_request`/`provide_parent_prefix` are documented as generic `UsageExtractor` hooks, but the worker answers them with codex-only resolution: `codex::collect_parent_prefix` parses the parent file as a codex rollout, `codex::session_meta_id` confirms identity, and `scan_sessions_tree` walks the codex home layout. Today only the codex extractor returns a request, so this is latent — but the first non-codex extractor that implements the documented-generic hook gets its parent transcript silently misparsed as a codex rollout (empty/wrong prefix, no error). Cheap guard: gate resolution on the task's `tool == "codex"` (or route parent-file prefix collection through the extractor itself), so a future implementer hits a loud gap instead of silent misparsing.

src/token_usage/codex.rs
svarlamov-git-ai: Fixed in e46b9e44e: a mid-read error now resolves to `None` — the burst-heuristic fallback, like a parent that failed to open — instead of verifying a truncated prefix as complete. Pinned by `a_parent_read_error_means_the_prefix_is_unknown_not_empty`.

src/token_usage/codex.rs
svarlamov-git-ai: Fixed in e46b9e44e: a fork marker arms replay matching only from the pristine state (replay `Done` and no usage counted), so repeated `session_meta` lines can no longer discard matching progress, re-trigger parent resolution per marker, or mis-skip genuine usage. Pinned by `a_repeated_fork_marker_does_not_rearm_replay_matching`.

src/daemon/token_usage_worker.rs
svarlamov-git-ai: Fixed in e46b9e44e: the scan now considers only `.jsonl` candidates, and `records_session_id` reads at most 64KiB for the first-line probe (a line truncated at the cap fails the JSON id parse, and rollout `session_meta` lines are far smaller).

src/daemon/token_usage_worker.rs
svarlamov-git-ai: Fixed in e46b9e44e: `resolve_parent_prefix` answers only `tool == "codex"` requests — another tool's request resolves as parent-unavailable (its burst fallback) instead of misparsing a foreign transcript as a rollout, with a comment directing per-tool dispatch when a second tool implements the hook.

src/daemon/token_usage_worker.rs
svarlamov-git-ai: Design call, kept separate: the three walkers select differently — `find_rollout_path_for_session_in_home` wants newest-by-mtime among `rollout-*<id>*.jsonl`, this one wants first-candidate-confirmed-by-recorded-id (ccusage's rule: recorded id, never filename), with its own entry cap. A predicate-parameterized shared walker would carry selection mode, confirmation callback, cap, and root-derivation options — a bigger surface than the ~30 duplicated lines. The .jsonl filter added in e46b9e44e narrows the drift risk (all three now agree on candidate shape).

PR #2241, opened 2026-08-27
feat(token-usage): attribute subagent usage to the subagent's own session

svarlamov-git-ai:
Usage is now leaf-attributed for every tool: subagent and fork
transcripts own their usage under their own session, carrying the
parent as the parent_session_id/external_parent_session_id attributes
— the same identity model SessionEvents already use, so TokenUsage and
SessionEvents finally agree about which session a transcript is. The
parent-inclusive total becomes a derived rollup over the parent
relationship instead of the stored identity, keeping per-agent cost
visible; recording everything under the parent was lossy in the other
direction.

Previously the worker rolled every stream with an
external_parent_session_id into its parent. For Codex forks that
contradicted both the spec ("sessions are per rollout file") and
ccusage, hiding fork children entirely; for Claude sidechains it
matched ccusage but differed from git-ai's own SessionEvents. This is
now a single standardized rule, with the Claude difference from
ccusage documented (per-session groupings differ; aggregate totals
match). Cross-file dedup of sidechain replays is unaffected — entry
dedup is global, not session-scoped.

tracked_files persists the parent id so DB-only reconciliation carries
the same relationship attributes; schema v5 is a pre-release reset
(existing rows are keyed by the rolled-up parent and don't record
their source rollout, so they cannot be reassigned in place). The
server already ingests both parent attributes into
token_usage_v2_events columns; no wire change.

---

### DIFF
--- a/src/daemon/token_usage_worker.rs
+++ b/src/daemon/token_usage_worker.rs
@@ -23,10 +23,12 @@
 //!   error backoff so a permanently failing transcript is not re-read on
 //!   every trigger. The size/mtime quiet-skip snapshot is written only after
 //!   a fully successful pass.
-//! - Claude subagent transcripts roll up to their parent session (matching
-//!   ccusage), which also lets sidechain replays dedup against the parent's
-//!   entries. Entry dedup itself is global across sessions (resume/fork
-//!   copies); when a replacement moves an entry between sessions, the
+//! - Usage is leaf-attributed: subagent/fork transcripts own their usage
+//!   under their own session (like SessionEvents), carrying the parent as
+//!   the parent_session_id/external_parent_session_id attributes. Entry
+//!   dedup is global across sessions (resume/fork copies; sidechain replays
+//!   of parent messages dedup against the parent's entries wherever they
+//!   live); when a replacement moves an entry between sessions, the
 //!   previous owner's files are invalidated so its buckets re-reconcile on
 //!   the next pass with their own attributes.
 //! - The `token_usage_metrics` feature flag gates the worker at daemon
@@ -547,29 +549,26 @@ fn sweep_candidates(
     tasks
 }
 
-/// Rollup identity of the session a transcript belongs to: subagent
-/// transcripts attribute to their parent session (matching ccusage).
+/// Identity of the session a transcript belongs to: every agent execution
+/// owns its own usage, so subagent/fork transcripts attribute to their own
+/// session and carry the parent as a relationship — the same leaf identity
+/// SessionEvents use. (Deviation from ccusage, which rolls Claude sidechains
+/// into the parent session: totals still match, but git-ai groups every
+/// tool the same way and keeps the per-agent view; the parent-inclusive
+/// total stays derivable through the parent attributes.)
 struct SessionIdentity {
     session_id: String,
     external_session_id: String,
+    external_parent_session_id: Option<String>,
     tool: String,
 }
 
 impl SessionIdentity {
     fn from_stream(stream: &StreamRecord) -> Self {
-        let (session_id, external_session_id) = match &stream.external_parent_session_id {
-            Some(parent_ext) => (
-                generate_session_id(parent_ext, &stream.tool),
-                parent_ext.clone(),
-            ),
-            None => (
-                stream.session_id.clone(),
-                stream.external_session_id.clone(),
-            ),
-        };
         Self {
-            session_id,
-            external_session_id,
+            session_id: stream.session_id.clone(),
+            external_session_id: stream.external_session_id.clone(),
+            external_parent_session_id: stream.external_parent_session_id.clone(),
             tool: stream.tool.clone(),
         }
     }
@@ -722,6 +721,7 @@ fn reconcile_flagged_sessions(
         let identity = SessionIdentity {
             session_id: session.session_id.clone(),
             external_session_id: session.external_session_id,
+            external_parent_session_id: session.external_parent_session_id,
             tool: session.tool,
         };
         emit_changed_buckets(token_db, &identity, || session.repo_url, sink)?;
@@ -769,8 +769,8 @@ const PARENT_SCAN_MAX_ENTRIES: usize = 50_000;
 
 /// Locate and read a forked child's parent rollout, returning its pre-fork
 /// usage prefix. Candidates come from the token database (files tracked
-/// under the parent's external session id — which, for rolled-up children,
-/// includes the child itself) and a bounded scan of the codex sessions tree
+/// under the parent's own external session id) and a bounded scan of the
+/// codex sessions tree
 /// for filenames carrying the parent id; a candidate counts only when its
 /// first line is a `session_meta` recording that id (ccusage locates parents
 /// by recorded id, never by filename). `None` — the parent is unresolvable —
@@ -904,6 +904,7 @@ fn process_file(
         stream_path,
         &identity.tool,
         &identity.external_session_id,
+        identity.external_parent_session_id.as_deref(),
     )?;
     let now = now_secs();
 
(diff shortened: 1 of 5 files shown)


src/daemon/token_usage_worker.rs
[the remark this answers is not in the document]
svarlamov-git-ai: Fixed — the comment now describes leaf-attribution keying.

src/daemon/token_usage_worker.rs
```diff
@@ -547,29 +549,26 @@ fn sweep_candidates(
...
-            ),
-            None => (
-                stream.session_id.clone(),
-                stream.external_session_id.clone(),
-            ),
-        };
         Self {
-            session_id,
-            external_session_id,
+            session_id: stream.session_id.clone(),
```
svarlamov-git-ai, line 569: **[code-review] Leaf attribution silently drops subagent cost from `git-ai analyze`**

`enrich_token_usage` (src/commands/analyze/sessions.rs:941) joins the token-usage cube on exact `session_id` (`dimensions: vec![format!("{}.session_id", TOKEN_USAGE_CUBE)]`), and the default sessions listing excludes subagent sessions (`let parent_filter = if include_subagents { None } else { Some("") };`, line 315). On main, subagent usage rolled up to the parent session id, so it surfaced in the parent's `cost_usd`/token columns. With this change it lands under leaf session ids that the default listing never pulls, and no parent-rollup was added on either the client or the query side — subagent tokens/cost vanish from the default analyze output (and even with `--include-subagents` they are not attributed to the parent).

Suggested fix: roll child usage into parents in `enrich_token_usage` (group by `parent_session_id` when present, or query the cube with an IN filter over parent+children and aggregate client-side).

src/daemon/token_usage_worker.rs
```diff
@@ -1107,6 +1108,17 @@ fn emit_changed_buckets(
...
+        // Subagent/fork sessions carry their parent as a relationship (leaf
+        // attribution), exactly like SessionEvents.
+        if let Some(parent_ext) = identity
+            .external_parent_session_id
+            .as_deref()
+            .filter(|parent_ext| !parent_ext.is_empty())
+        {
+            attrs = attrs
+                .external_parent_session_id(parent_ext.to_string())
+                .parent_session_id(generate_session_id(parent_ext, &identity.tool));
```
svarlamov-git-ai, line 1159: **[code-review] Plain Codex forks get leaf attribution but never carry the parent relationship the spec promises**

`CodexAgent::detect_subagent_parent` (src/streams/agents/codex.rs:153) requires `payload.thread_source == "subagent"` before it reads `forked_from_id`, so a plain fork (`session_meta` with `forked_from_id` but no `thread_source`) produces a stream record with `external_parent_session_id = None` — and these attributes are never emitted. Meanwhile the extractor itself happily extracts the same `forked_from_id` for parent-prefix matching, and docs/token-usage-events-spec.md now states usage is leaf-attributed "Codex fork/subagent rollouts included — with the parent carried as the `parent_session_id`/`external_parent_session_id` attributes" and that the parent-inclusive total is "derivable through the parent relationship".

For plain forks that derived rollup permanently misses the fork's usage. The new integration test `forked_codex_session_counts_only_its_own_usage` exercises exactly this shape and its emitted event has no parent linkage. Either widen `detect_subagent_parent` to cover plain `forked_from_id` (matching `fork_parent_id` in src/token_usage/codex.rs) or scope the spec claim to subagent spawns.

src/token_usage/db.rs
```diff
@@ -223,11 +223,89 @@ const MIGRATIONS: &[&str] = &[
 
     INSERT INTO schema_version (version) VALUES (4);
     "#,
+    // Version 5: leaf session attribution. Usage is owned by the transcript's
```
svarlamov-git-ai, line 235: **[code-review] Table rebuild leaves previously-emitted parent-attributed server buckets permanently unzeroed (double count for flag-enabled installs)**

Dropping `bucket_state` deletes the only record that buckets were already emitted, and emptied buckets are re-emitted as zero *only* from surviving `bucket_state` rows. Any install that had `token_usage_metrics` enabled (config/env on a release build, or any debug build — the flag is debug-on) emitted subagent usage under `generate_session_id(parent_ext, tool)`; after this migration the same usage re-emits under leaf session ids while the old parent-attributed server rows are never zeroed — double counted in any parent-inclusive rollup. The "pre-release reset" comment is accurate for release defaults, but debug-build emitters (every dev machine) hit this deterministically. Worth confirming the server side can tolerate or purge the orphaned rows.

src/daemon/token_usage_worker.rs
svarlamov-git-ai: Not a regression from this PR: git-ai analyze reads the v1 public_v1_token_usage cube, which is populated by the SessionEvent ETL — and SessionEvents have always used leaf session ids (stream.session_id with parent attrs), unchanged here. The v2 TokenUsage events this PR re-attributes have no consumer in analyze (or any dashboard) yet. The underlying product gap — default parent listings not rolling subagent cost up — therefore predates this stack and applies to the v1 path; worth a follow-up that aggregates by the parent relationship (now uniformly available on both event families), tracked separately.

src/daemon/token_usage_worker.rs
svarlamov-git-ai: Intentional, and now stated in the spec (9d2ba3411): a plain user-initiated fork (forked_from_id without a subagent thread_source) is a new top-level conversation, not a sub-task — linking it to its parent would also hide it from default session listings, which filter on parent_session_id. Subagent spawns carry the parent attributes; the replayed prefix is removed either way, so no usage is double-counted.

src/token_usage/db.rs
svarlamov-git-ai: Accepted by design: the v3/v5 rebuilds are explicitly pre-release resets (the flag is release-off; the project owner decided existing data — including anything a debug build uploaded — is disposable and will be cleared server-side). The migration comments document this; production-grade identity migrations would need the source-rollout ownership the finding describes, which is exactly why we are not attempting an in-place reassignment.

src/token_usage/db.rs
```diff
@@ -232,11 +232,89 @@ const MIGRATIONS: &[&str] = &[
...
+    // own session — subagent/fork transcripts no longer roll up into their
+    // parent — with the parent carried as a relationship
+    // (tracked_files.external_parent_session_id, emitted as the
+    // parent_session_id/external_parent_session_id attributes, matching
+    // SessionEvents). Pre-release reset like v3: existing rows are keyed by
+    // the rolled-up parent session and cannot be reassigned (they don't
+    // record their source rollout), so the tables are rebuilt and the next
+    // pass re-extracts everything under leaf identities.
+    r#"
```
svarlamov-git-ai, line 244: **[code-review] v5 restates ~70 lines of DDL where `ALTER TABLE` + `DELETE` is equivalent — and the restatement already drifted once**

v5's only structural delta over v4 is one nullable column (`tracked_files.external_parent_session_id TEXT`) plus the pre-release data reset, but it re-declares all three tables and four indexes. That copy must stay in sync with v3+v4 by hand, and it already failed once on this branch: v5 initially recreated v3's plain bucket index instead of v4's `COALESCE(speed, 0)` expression index, silently putting the reconciliation GROUP BY back on a temp B-tree (fixed in c2a04e749). A delete-based reset can't have that bug because it preserves v4's definitions by construction, and migrations run strictly sequentially so v4's schema is fixed at this point (v2 already uses `ALTER TABLE ... ADD COLUMN`):

```sql
ALTER TABLE tracked_files ADD COLUMN external_parent_session_id TEXT;
DELETE FROM tracked_files;
DELETE FROM usage_entries;
DELETE FROM bucket_state;
INSERT INTO schema_version (version) VALUES (5);
```

svarlamov-git-ai: Acknowledged — the restatement did drift once, which is why the v4/v5 index parity is now pinned by tests (expression-index presence + the forward-version guard). Kept as a rebuild because v5's point is the pre-production data reset (leaf attribution re-keys sessions): DELETE FROM x3 + ALTER expresses the same end state with the same risk surface, and this schema's convention (v3, v5) is reset-by-rebuild. Post-1.0, migrations will be additive ALTERs.

src/daemon/token_usage_worker.rs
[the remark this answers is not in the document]
svarlamov-git-ai: Fixed in 7d3410647: the cap is now 1MiB — codex `session_meta` embeds the session's instructions (AGENTS.md contents) and can legitimately exceed 64KiB, so the probe bounds memory without rejecting real parents (a line truncated at the cap still fails the JSON id parse rather than matching).

PR #2255, opened 2026-08-30
fix(daemon): never run family side effects inline on the trace ingest worker

### LINKED ISSUE #2252: git write ops slow down or hang for 45-70s while the daemon grinds write-op side effects; checkpoint ingress stops being admitted (v1.6.24)
## Summary

On v1.6.24 (macOS arm64), git write ops in a large monorepo (~100k files) intermittently degrade severely. Two related symptom classes on the same machine:

1. **Slowdown** (timed, most recent): a `git checkout` took 73 s wall with only 26 s CPU -- ~47 s spent waiting, no user intervention.
2. **Full hangs** (two earlier incidents): a `gt submit` produced no output for 3+ minutes and a `git checkout` hung; in both cases killing the git-ai daemon made the command complete immediately.

Around the timed slowdown, the daemon shows a ~2-minute logging gap covering the checkout, and immediately afterwards its checkpoint ingress **stops being admitted**: checkpoints are "received into bounded ingress" but no admission or processing follows for 30+ seconds (baseline minutes earlier: 0-123 ms). This is distinct from #2244's memory-balloon abort loop (no memory emergencies, no flush overruns, no errors in the logs below) and is not addressed by the #2245-#2251 stack.

## Production evidence (v1.6.24, macOS arm64, 3 incidents in 2 days on one machine)

**Slowdown incident (timed):** `time command git checkout dev/<user>/<branch>` in the ~100k-file monorepo:

```
2.60s user 23.24s system 35% cpu 1:13.32 total
```

73 s wall, 26 s CPU — ~47 s waiting on something outside the process. (This run completed on its own; no daemon kill.) The user's daemon log for that window (fresh daemon, healthy, zero errors):

```
20:07:02-20:08:26  heavy agent-checkpoint traffic, seq 17 -> 76, ALL received -> admitted ->
                   processed within ms (receipt_to_admission_ms=0..123, every status="ok")
20:08:26.342       last log line before the incident
                   <-- the timed 73 s checkout runs ~20:09:10-20:10:23; the daemon logs
                       NOTHING for the ~2 minutes covering it -->
20:10:34.913  INFO checkpoint received into bounded ingress receipt_seq=77 retained_bytes=42991
20:10:35.355  INFO checkpoint received into bounded ingress receipt_seq=78 retained_bytes=43323
20:11:04.843  INFO checkpoint received into bounded ingress receipt_seq=79 retained_bytes=42974
20:11:04.908  INFO checkpoint received into bounded ingress receipt_seq=80 retained_bytes=43273
20:11:06.200  INFO checkpoint received into bounded ingress receipt_seq=81 retained_bytes=42924
20:11:06.254  INFO checkpoint received into bounded ingress receipt_seq=82 retained_bytes=43299
20:11:06.660  INFO checkpoint received into bounded ingress receipt_seq=83 retained_bytes=42950
20:11:06.706  INFO checkpoint received into bounded ingress receipt_seq=84 retained_bytes=43115
```

Two corroborating signals around the timed slowdown: (1) a daemon that had been logging checkpoint turnarounds every few hundred ms goes **completely silent for the ~2 minutes covering the checkout** (20:08:26 -> 20:10:34) -- consistent with its workers busy in a long non-logging pass; (2) when traffic resumes immediately after the checkout, seq 77-84 are received but **never admitted or processed** through end-of-log 30+ seconds later, against a baseline of receipt_to_admission_ms=0..123. An active Cursor agent session was running on the machine (source of the ~43 KB checkpoints). No errors, no memory emergencies anywhere in the log.

**Earlier incidents, same machine:** a `gt submit` "running for over 3 min with no output — I killed git-ai and output got printed right after that" (three daemon replacements in 13 minutes visible in fleet telemetry), and a hung `git checkout` the day before, also resolved instantly by killing the daemon. Related self-heal evidence from the same logs: `daemon health check failed ... processing_stalled=true queued_payloads=245` and `socket health restart deferred while accepted checkpoints remain ... consecutive_deferrals=2`.

## What we could and could not reproduce synthetically

Deterministic (isolated daemon, `git_path` pointed at a 3 s-sleep git wrapper simulating monorepo-scale side-effect latency): one traced commit's write-op processing takes the daemon **~90 s** — the family-actor grind precondition is easy to hit. Harness attached below.

Not reproduced synthetically: the user-visible slowdown/hang. In our isolated setup, trace2 emission and `git-ai checkpoint claude` submissions never blocked (the CLI checkpoint even succeeds with no daemon running, so the synchronous leg we observe in production evidently runs through a path our synthetic session does not exercise — presumably the agent-session checkpoint flow that produced seq 77-84). On the affected production machine the degradation reproduces reliably: large-diff branch switch in the monorepo with an active Cursor session.

Question for maintainers: which client path waits synchronously on checkpoint admission (or on the family actor) such that a concurrent git write command stalls? Given the daemon's 2-minute logging gap covering the timed slowdown, the never-admitted seq 77-84 right after it, and kill-instantly-unblocks in the two hang incidents, something in the traced-git or agent-hook flow waits while admission is queued behind side effects.

## Repro harness (grind precondition)

<details><summary>harness_adm.sh — isolated daemon + slow-git side effects + timed traced ops</summary>

```bash
#!/usr/bin/env bash
# Repro: synchronous checkpoint admission queues behind family write-op side
# effects -- a second traced git command blocks while the daemon grinds the
# first one's side effects. slow-git (sleep per daemon-side git spawn) stands
# in for monorepo-scale side-effect latency.
set -uo pipefail

BIN="${1:-$HOME/.git-ai/bin/git-ai}"
WORK="$2"
SLOW_S="${3:-3}"     # seconds each daemon-side git invocation sleeps

mkdir -p "$WORK"
H="$WORK/home"; S="$WORK/s"; R="$WORK/repo"
mkdir -p "$H/.git-ai" "$S" "$R"

# slow git used ONLY by the daemon (config git_path); user git is the real one
cat > "$WORK/slow-git" <<EOF
#!/bin/bash
sleep $SLOW_S
exec /opt/homebrew/bin/git "\$@"
EOF
chmod +x "$WORK/slow-git"

# stalled telemetry backend (accepts, never responds)
/usr/bin/python3 - "$WORK/port.txt" <<'PY' &
import socket, sys, threading, time
srv = socket.socket(); srv.bind(("127.0.0.1", 0)); srv.listen(64)
open(sys.argv[1], "w").write(str(srv.getsockname()[1]))
def hold(c):
    try: time.sleep(3600)
    finally: c.close()
while True:
    c, _ = srv.accept()
    threading.Thread(target=hold, args=(c,), daemon=True).start()
PY
MOCK_PID=$!
for _ in $(seq 1 50); do [[ -s "$WORK/port.txt" ]] && break; sleep 0.1; done
PORT=$(cat "$WORK/port.txt")

printf '[user]\n\tname = Repro\n\temail = repro@example.com\n[init]\n\tdefaultBranch = main\n[trace2]\n\teventTarget = af_unix:stream:%s\n\teventNesting = 5\n' "$S/t.sock" > "$H/.gitconfig"
cat > "$H/.git-ai/config.json" <<CFG
{
  "git_path": "$WORK/slow-git",
  "telemetry_oss": "off",
  "disable_version_checks": true,
  "disable_auto_updates": true,
  "allow_repositories": ["*"],
  "api_base_url": "http://127.0.0.1:$PORT/api/git-ai",
  "api_key": "test",
  "notes_backend": { "kind": "http", "backend_url": "http://127.0.0.1:$PORT/api/git-ai" },
  "feature_flags": { "async_mode": true, "rewrite_stash": true }
}
CFG

E=( "HOME=$H" "GIT_CONFIG_GLOBAL=$H/.gitconfig" "GIT_CONFIG_NOSYSTEM=1"
    "GIT_AI_DAEMON_HOME=$H" "GIT_AI_DAEMON_CONTROL_SOCKET=$S/c.sock"
    "GIT_AI_DAEMON_TRACE_SOCKET=$S/t.sock" )
g() { env "${E[@]}" timeout 120 git "$@"; }
gq() { env "${E[@]}" GIT_TRACE2_EVENT=0 timeout 120 git "$@"; }

# repo: modest size, two branches with a real diff
( cd "$R"
  gq init -q .
  gq remote add origin git@github.com:rhmarkets/repro-admission.git
  for i in $(seq 1 300); do mkdir -p "d$((i % 20))"; echo "content $i" > "d$((i % 20))/f$i.txt"; done
  gq add -A; gq commit -qm base
  gq checkout -qb feature
  for i in $(seq 1 300); do echo "changed $i" >> "d$((i % 20))/f$i.txt"; done
  gq add -A; gq commit -qm feature-change
  gq checkout -q main
) >/dev/null 2>&1

env "${E[@]}" "$BIN" bg run > "$WORK/daemon.stdout" 2>&1 &
DP=$!
for _ in $(seq 1 100); do [[ -S "$S/c.sock" && -S "$S/t.sock" ]] && break; sleep 0.1; done
[[ -S "$S/t.sock" ]] || { echo "FATAL: no daemon sockets"; kill $MOCK_PID 2>/dev/null; exit 2; }
sleep 2

t() { /usr/bin/python3 -c 'import time; print(f"{time.time():.2f}")'; }
dur() { /usr/bin/python3 -c "print(f'{$2-$1:.2f}')"; }

# tiny claude-format transcript for hook checkpoints
printf '{"role":"user","message":{"content":[{"type":"text","text":"repro"}]}}\n' > "$H/t.jsonl"
ckpt() {
  ( cd "$R" && env "${E[@]}" timeout 120 "$BIN" checkpoint claude \
      --hook-input "{\"cwd\":\"$R\",\"transcript_path\":\"$H/t.jsonl\",\"session_id\":\"repro-adm\"}" ) >/dev/null 2>&1
}

echo "=== control: both ops UNTRACED (daemon blind) ==="
( cd "$R"; echo x >> d1/f1.txt; )
s=$(t); ( cd "$R"; gq add -A; gq commit -qm untraced-warm ) >/dev/null 2>&1; e=$(t)
echo "untraced commit: $(dur $s $e)s"
s=$(t); ( cd "$R"; gq checkout -q feature; gq checkout -q main ) >/dev/null 2>&1; e=$(t)
echo "untraced checkout roundtrip: $(dur $s $e)s"

echo "=== baseline: hook checkpoint with idle daemon ==="
s=$(t); ckpt; e=$(t)
echo "checkpoint (idle): $(dur $s $e)s"

echo "=== phase A: traced commit starts the family side-effect grind ==="
( cd "$R"; echo y >> d2/f2.txt; )
s=$(t); ( cd "$R"; g add -A; g commit -qm traced-op-A ) >/dev/null 2>&1; e=$(t)
echo "traced commit A: $(dur $s $e)s"
# let the daemon ingest A and begin its (slow-git) side-effect pass
sleep 8
echo "-- ingestion check --"
grep -hc "git write op" "$H/.git-ai/internal/daemon/logs/"*.log 2>/dev/null || echo 0

echo "=== phase B: traced ops DURING the grind (expect blocking) ==="
for i in 1 2 3 4 5 6; do
  ( cd "$R"; echo "b$i" >> d3/f3.txt; )
  s=$(t); ( cd "$R"; g add -A; g commit -qm "traced-B$i" ) >/dev/null 2>&1; e=$(t)
  echo "traced commit B$i: $(dur $s $e)s"
done
s=$(t); ckpt; e=$(t)
echo "checkpoint (during grind) 1: $(dur $s $e)s"
s=$(t); ckpt; e=$(t)
echo "checkpoint (during grind) 2: $(dur $s $e)s"
s=$(t); ( cd "$R"; g checkout -q feature ) >/dev/null 2>&1; rcb=$?; e=$(t)
echo "traced checkout B: $(dur $s $e)s rc=$rcb"
s=$(t); ( cd "$R"; g checkout -q main ) >/dev/null 2>&1; e=$(t)
echo "traced checkout B2: $(dur $s $e)s"

echo "=== phase C: same checkouts untraced again (recovery control) ==="
sleep 1
s=$(t); ( cd "$R"; gq checkout -q feature; gq checkout -q main ) >/dev/null 2>&1; e=$(t)
echo "untraced checkout roundtrip: $(dur $s $e)s"

sleep 10
kill $DP $MOCK_PID 2>/dev/null
sleep 1; pkill -f "GIT_AI_DAEMON_HOME=$H" 2>/dev/null

echo "=== daemon checkpoint admission timeline ==="
grep -hE "checkpoint (received|prepared|processing completed)|git write op" "$H/.git-ai/internal/daemon/logs/"*.log 2>/dev/null \
  | sed -E 's/component\S+ //; s/phase\S+ //; s/trace_id\S+ //; s/family=\S+ //' | tail -30

```

</details>

## Suggested directions

- Never let checkpoint admission (or any client-visible ack) queue behind a family write-op side-effect pass; bound the pass or make admission independent of it.
- Deadline on any client-blocking daemon interaction: after N ms, accept-and-defer (or refuse) instead of keeping a git process or agent hook waiting.
- Emit an explicit log line when ingress admission is delayed beyond a threshold (seq 77-84 above had to be inferred from the absence of admission lines).

We can run any diagnostic build on the affected machine and have fleet telemetry to confirm a fix within a day. Related: #2244 (memory-balloon abort loop, same machine), PRs #2245-#2251.

svarlamov-git-ai:
## Problem

Fixes #2252.

On v1.6.24, git write ops in a large monorepo intermittently degrade severely (73s checkout with ~47s of waiting; multi-minute hangs resolved instantly by killing the daemon), and the daemon's checkpoint ingress stops being admitted: checkpoints are logged "received into bounded ingress" but no admission or processing follows (seq 77–84 in the report), against a healthy baseline of `receipt_to_admission_ms=0..123`.

## Root cause

The serial trace-ingest worker executed family side-effect passes **inline**:

- `apply_trace_payload_to_state` → `replace_pending_root_entry` / `append_ready_command_entry` took the per-family `side_effect_exec_lock` and ran `drain_ready_family_sequencer_entries_locked` in place — i.e. the full write-op side-effect pipeline (post-commit authorship, rewrite-note migration, working-log migration: unbounded-duration git work, ~90s per write op at monorepo scale per the issue's harness) ran on the ingest worker.
- The non-sequencer `Applied` branch of `ingest_trace_payload_fast` ran `maybe_apply_side_effects_for_applied_command` inline as well.

While one family ground, `processed_trace_ingest_seq` froze. Checkpoint admission (`prepare_checkpoint_admission`) waits on that watermark (`wait_for_trace_ingest_seq(trace_ingest_target)`), so admission stalled **for every repository family**, exactly matching the report:

- seq 77–84 received but never admitted;
- queued trace payloads backing up (`queued_payloads=245`, `processing_stalled=true`);
- socket-health restarts deferred "while accepted checkpoints remain" — the deferral was waiting on the very checkpoints the stall blocked, producing the observed restart churn (three daemon replacements in 13 minutes).

## Fix

Trace ingestion is now **sequencing-only**:

- `replace_pending_root_entry` and `append_ready_command_entry` are constant-time sequencer-map mutations — no exec lock, no inline drain.
- Drains run on detached tasks (`schedule_family_drain` / `schedule_ready_family_drains_after_root_cleared`), still strictly serialized per family by the existing `side_effect_exec_lock`, and still gated by the existing `unadmitted_checkpoints == 0` / open-root fences, so ordering semantics are unchanged.
- Side effects of commands that do not participate in the sequencer (`git am`, `filter-branch`, …) are sequenced too, as `FamilySequencerEntry::AppliedSideEffects` entries ordered by the command's start time, so they keep the exact per-family ordering the inline worker used to guarantee (and can no longer overlap a same-family drain, closing a pre-existing window). Commit-time file-timestamp snapshots still start on the worker before the entry is queued.

Because these passes were previously covered by the ingest watermark, three guarantees are restored explicitly:

- **Bounded concurrency**: side-effect passes do blocking git work on the 4-thread daemon runtime; with detached drains, simultaneously grinding families could otherwise occupy every worker and starve the runtime. Command/applied passes are bounded by a dedicated 2-permit `command_side_effect_semaphore`; checkpoint side effects keep their own existing semaphore (they run on the blocking pool via `spawn_blocking`, so long command passes can never starve checkpoint processing). Lock ordering is uniform — family exec lock, then permit — so a permit holder never waits on another family's exec lock and no deadlock is possible.
- **`sync.family`** now drains every family's ready entries and waits for in-flight detached passes globally (`inflight_effects_by_family`, registered on the worker *before* the watermark advances). Side effects can write into *other* repositories — `git push <path>` pushes authorship notes into the destination repo — and before detachment the global watermark implied all already-ingested passes had completed, which the notes-sync regression suite depends on (this is what the first CI round caught). Entries fail-closed behind a still-open trace root stay queued without blocking the fence, exactly as pre-change, so sync still does not hang on another repo's in-flight interactive command.
- **Graceful shutdown** (`drain_accepted_checkpoints`) also fences on in-flight effects, so the shutdown response cannot be sent — and the process cannot exit — while a detached pass is mid-write on entries it already popped from the sequencer.
- In-flight registrations are RAII (`FamilyEffectGuard`), created **before** entries are popped from the sequencer and released on every exit path (early returns, panics, dropped tasks) after all completion bookkeeping — so the fences can neither miss an imminent pass nor be left blocked by a failed one. The update/uptime restart paths defer while a pass is executing, like graceful shutdown.

## Tests (TDD — each failed before the corresponding change)

All grind tests gate on a daemon-side signal (the delay hook logs `test side-effect delay started` when the pass begins executing) instead of fixed sleeps, so loaded runners can't make them pass vacuously or fail spuriously.

- `family_side_effect_grind_does_not_stall_independent_checkpoint_admission` — a 10s side-effect grind in one family (existing `GIT_AI_TEST_DELAY_SIDE_EFFECT_MS_FOR_COMMAND` hook); a checkpoint for an independent family must be admitted **and processed** well before the grind ends. On `main` it stays unadmitted for the whole grind.
- `checkpoint_admission_proceeds_while_same_family_side_effects_grind` — the headline symptom: "received into bounded ingress" must be followed by "prepared for family admission" while the same family's side effects are still running; on `main` the admission line never appears until the grind completes.
- `sync_family_waits_for_detached_non_sequencer_side_effects` — `sync.family` must not return while a detached `git am` pass is in flight.
- `sync_family_covers_cross_repo_side_effects_of_other_families` — deterministic repro of the notes-sync CI failure: `sync.family` on a push destination must not return before the push side-effect pass finishes pushing authorship notes into it.
- `graceful_shutdown_waits_for_detached_side_effect_passes` — graceful shutdown must not complete while a detached command pass is still running.

### DIFF
--- a/src/daemon.rs
+++ b/src/daemon.rs
@@ -104,6 +104,14 @@ const DAEMON_SOCKET_PROBE_TIMEOUT: Duration = Duration::from_millis(100);
 const CHECKPOINT_INGRESS_REQUEST_LIMIT: usize = 1_024;
 const CHECKPOINT_INGRESS_BYTE_LIMIT: usize = 64 * 1024 * 1024;
 const CHECKPOINT_FAMILY_DRAIN_CONCURRENCY: usize = 2;
+/// Global bound on concurrently executing command side-effect passes. They
+/// do blocking git work directly on the daemon runtime's 4 worker threads,
+/// and with drains running on detached tasks an unbounded number of
+/// simultaneously grinding families could otherwise occupy every worker and
+/// starve the runtime (#2252). Checkpoint side effects keep their own
+/// semaphore: they run on the blocking pool and must not be starved by
+/// long command passes.
+const COMMAND_SIDE_EFFECT_CONCURRENCY: usize = 2;
 // Trace2 frames are written synchronously by Git to the daemon's Unix socket.
 // With small kernel socket buffers (macOS defaults to ~8 KiB), a bursty trace2
 // stream can fill the buffer and block the raw `git` process in `write()` until
@@ -2648,6 +2656,14 @@ fn read_checkpoint_body<R: BufRead>(
 enum FamilySequencerEntry {
     PendingRoot,
     ReadyCommand(Box<crate::daemon::domain::NormalizedCommand>),
+    /// A command already applied to family state (it did not participate in
+    /// the sequencer, e.g. `git am`) whose side-effect pass is still pending.
+    /// Sequencing the pass keeps it ordered with, and serialized against,
+    /// the family's other passes (#2252).
+    AppliedSideEffects {
+        applied: Box<crate::daemon::domain::AppliedCommand>,
+        commit_file_timestamp_snapshots: CommitFileTimestampSnapshotHandles,
+    },
     Checkpoint {
         request: Box<CheckpointRequest>,
         receipt_seq: u64,
@@ -2681,6 +2697,30 @@ type CommitFileTimestampSnapshotHandles = HashMap<String, CommitFileTimestampSna
 const COMMIT_FILE_TIMESTAMP_SNAPSHOT_WAIT: Duration = Duration::from_millis(500);
 const SESSION_EVENT_RECOVERY_PREFLIGHT_WAIT: Duration = Duration::from_secs(2);
 const SESSION_EVENT_RECOVERY_PREFLIGHT_POLL: Duration = Duration::from_millis(100);
+
+/// RAII registration of an in-flight family side-effect pass; see
+/// [`ActorDaemonCoordinator::begin_family_effect_guarded`].
+struct FamilyEffectGuard<'a> {
+    coordinator: &'a ActorDaemonCoordinator,
+    family: String,
+}
+
+impl Drop for FamilyEffectGuard<'_> {
+    fn drop(&mut self) {
+        let _ = self.coordinator.end_family_effect(&self.family);
+    }
+}
+
+/// Extracts a printable message from a `catch_unwind` panic payload.
+fn panic_payload_message(panic_payload: &(dyn std::any::Any + Send)) -> String {
+    if let Some(message) = panic_payload.downcast_ref::<String>() {
+        message.clone()
+    } else if let Some(message) = panic_payload.downcast_ref::<&str>() {
+        (*message).to_string()
+    } else {
+        "unknown panic".to_string()
+    }
+}
 
 fn run_blocking_side_effect<T>(operation: impl FnOnce() -> T) -> T {
     if tokio::runtime::Handle::try_current()
@@ -2768,7 +2808,11 @@ pub struct ActorDaemonCoordinator {
         Mutex<HashMap<String, VecDeque<RecentReplayPrerequisite>>>,
     side_effect_errors_by_family: Mutex<HashMap<String, BTreeMap<u64, String>>>,
     side_effect_exec_locks: Mutex<HashMap<String, Arc<AsyncMutex<()>>>>,
+    /// Families with a scheduled-or-running coalesced drain task; see
+    /// [`Self::schedule_family_drain`].
+    scheduled_family_drains: Mutex<HashSet<String>>,
     checkpoint_side_effect_semaphore: Semaphore,
+    command_side_effect_semaphore: Semaphore,
     checkpoint_ingress_quota: Arc<CheckpointIngressQuota>,
     checkpoint_ingress_tx: std::sync::OnceLock<mpsc::Sender<AcceptedCheckpoint>>,
     next_checkpoint_receipt_seq: AtomicUsize,
@@ -2865,7 +2909,6 @@ impl DaemonExitAction {
 
 enum TracePayloadApplyOutcome {
     None,
-    Applied(Box<crate::daemon::domain::AppliedCommand>),
     QueuedFamily,
 }
 
@@ -2892,7 +2935,9 @@ impl ActorDaemonCoordinator {
             recent_replay_prerequisites_by_family: Mutex::new(HashMap::new()),
             side_effect_errors_by_family: Mutex::new(HashMap::new()),
             side_effect_exec_locks: Mutex::new(HashMap::new()),
+            scheduled_family_drains: Mutex::new(HashSet::new()),
             checkpoint_side_effect_semaphore: Semaphore::new(CHECKPOINT_FAMILY_DRAIN_CONCURRENCY),
+            command_side_effect_semaphore: Semaphore::new(COMMAND_SIDE_EFFECT_CONCURRENCY),
             checkpoint_ingress_quota: Arc::new(CheckpointIngressQuota::new(
                 CHECKPOINT_INGRESS_REQUEST_LIMIT,
                 CHECKPOINT_INGRESS_BYTE_LIMIT,
@@ -3295,6 +3340,58 @@ impl ActorDaemonCoordinator {
         Ok(())
     }
 
+    /// Registers an in-flight family side-effect pass for the guard's
+    /// lifetime. Sync, await, and graceful shutdown fence on this
+    /// registration, so it must be released on every exit path — early
+    /// returns, panics, and dropped tasks included — which only a Drop
+    /// impl can guarantee.
+    fn begin_family_effect_guarded<'a>(&'a self, family: &str) -> FamilyEffectGuard<'a> {
+        let _ = self.begin_family_effect(family);
+        FamilyEffectGuard {
+            coordinator: self,
+            family: family.to_string(),
+        }
+    }
+
+    /// Attribution work an automatic restart would abandon mid-flight:
+    /// accepted checkpoints, queued trace payloads, actionable sequencer
+    /// entries, and executing side-effect passes. PendingRoot placeholders
+    /// are deliberately excluded so an idle interactive command (e.g. a
+    /// rebase waiting on an editor) cannot defer restarts forever (#2252).
+    fn has_pending_attribution_work(&self) -> bool {
+        if self.outstanding_checkpoint_state().0 > 0 {
+            return true;
+        }
+        if self.queued_trace_payloads.load(Ordering::Relaxed) > 0 {
+            return true;
+        }
+        if self.has_inflight_family_effects() {
+            return true;
+        }
+        if let Ok(map) = self.family_sequencers_by_family.lock()
+            && map.values().any(|state| {
+                state
+                    .entries
+                    .values()
+                    .any(|entry| !matches!(entry, FamilySequencerEntry::PendingRoot))
+            })
+        {
+            return true;
+        }
+        false
+    }
+
+    /// Whether any detached side-effect pass is currently in flight, in any
+    /// family. Passes register via `begin_family_effect` before the trace
+    /// ingest watermark advances, so this is a valid completion fence for
+    /// work the watermark no longer covers (#2252).
+    fn has_inflight_family_effects(&self) -> bool {
+        self.inflight_effects_by_family
+            .lock()
+            .map(|map| !map.is_empty())
+            .unwrap_or(false)
+    }
+
     fn end_family_effect(&self, family: &str) -> Result<(), GitAiError> {
         let mut map = self
             .inflight_effects_by_family
@@ -3525,35 +3622,34 @@ impl ActorDaemonCoordinator {
         self.append_pending_root_entry(&family, root_sid, started_at_ns)
     }
 
-    async fn append_ready_command_entry(
+    /// Appends an entry to the family sequencer, ordered by the originating
+    /// command's start time. The caller is responsible for scheduling a
+    /// drain of the family afterwards — this must stay a constant-time map
+    /// insert because it runs on the serial trace ingest worker, whose
+    /// watermark checkpoint admission waits on (#2252).
+    fn append_family_sequencer_entry(
         &self,
         family: &str,
-        command: crate::daemon::domain::NormalizedCommand,
+        started_at_ns: u128,
+        entry: FamilySequencerEntry,
     ) -> Result<(), GitAiError> {
-        let exec_lock = self.side_effect_exec_lock(family)?;
-        let _guard = exec_lock.lock().await;
-        {
-            let mut sequencers = self.family_sequencers_by_family.lock().map_err(|_| {
-                GitAiError::Generic("family sequencer map lock poisoned".to_string())
-            })?;
-            let state =
-                sequencers
-                    .entry(family.to_string())
-                    .or_insert_with(|| FamilySequencerState {
-                        next_ordinal: 1,
-                        entries: BTreeMap::new(),
-                    });
-            let order = FamilySequencerOrder {
-                started_at_ns: command.started_at_ns,
-                ordinal: state.next_ordinal,
-            };
-            state.next_ordinal = state.next_ordinal.saturating_add(1);
-            state
-                .entries
-                .insert(order, FamilySequencerEntry::ReadyCommand(Box::new(command)));
-        }
-        self.drain_ready_family_sequencer_entries_locked(family)
-            .await
+        let mut sequencers = self
+            .family_sequencers_by_family
+            .lock()
+            .map_err(|_| GitAiError::Generic("family sequencer map lock poisoned".to_string()))?;
+        let state = sequencers
+            .entry(family.to_string())
+            .or_insert_with(|| FamilySequencerState {
+                next_ordinal: 1,
+                entries: BTreeMap::new(),
+            });
+        let order = FamilySequencerOrder {
+            started_at_ns,
+            ordinal: state.next_ordinal,
+        };
+        state.next_ordinal = state.next_ordinal.saturating_add(1);
+        state.entries.insert(order, entry);
+        Ok(())
     }
 
     async fn drain_ready_family_sequencer_entries(&self, family: &str) -> Result<(), GitAiError> {
@@ -3586,18 +3682,111 @@ impl ActorDaemonCoordinator {
         Ok(())
     }
 
-    async fn drain_ready_family_sequencers_after_root_cleared(
-        &self,
-        family: Option<String>,
-    ) -> Result<(), GitAiError> {
+    /// Schedules drains for sequencer entries unblocked by a cleared trace
+    /// root. Drains run detached: side-effect passes are unbounded-duration
+    /// git work and must never execute inline on the trace ingest worker,
+    /// whose watermark checkpoint admission waits on (#2252).
+    fn schedule_ready_family_drains_after_root_cleared(self: &Arc<Self>, family: Option<String>) {
         if let Some(family) = family {
-            self.drain_ready_family_sequencer_entries(&family).await
+            self.schedule_family_drain(family);
         } else {
-            self.drain_all_ready_family_sequencers().await
+            self.schedule_all_ready_family_drains();
         }
     }
 
-    async fn replace_pending_root_entry(
+    /// Schedules a detached drain for one family, coalescing to at most one
+    /// scheduled-or-running drain task per family. The marker is released
+    /// only after a pass that ends with no actionable front entry (checked
+    /// atomically with the release), so an entry appended before this call
+    /// is always covered: either the running task's next pass pops it, or
+    /// this call spawns a fresh task.
+    fn schedule_family_drain(self: &Arc<Self>, family: String) {
+        {
+            let Ok(mut scheduled) = self.scheduled_family_drains.lock() else {
+                return;
+            };
+            if !scheduled.insert(family.clone()) {
+                return;
+            }
+        }
+        let coordinator = Arc::clone(self);
+        tokio::spawn(async move {
+            loop {
+                if let Err(error) = coordinator
+                    .drain_ready_family_sequencer_entries(&family)
+                    .await
+                {
+                    tracing::error!(
+                        component = "daemon",
+                        phase = "checkpoint_processing",
+                        reason = "family_drain_failed",
+                        %family,
+                        %error,
+                        "failed draining family sequencer"
+                    );
+                    if let Ok(mut scheduled) = coordinator.scheduled_family_drains.lock() {
+                        scheduled.remove(&family);
+                    }
+                    return;
+                }
+                // Deregister atomically with the emptiness check: an entry
+                // appended after the pass but before deregistration must
+                // either be seen here (loop again) or by the fresh task its
+                // own schedule call spawns after we release the marker.
+                let Ok(mut scheduled) = coordinator.scheduled_family_drains.lock() else {
+                    return;
+                };
+                if !coordinator.family_has_actionable_front_entry(&family) {
+                    scheduled.remove(&family);
+                    return;
+                }
+            }
+        });
+    }
+
+    /// Whether the family's sequencer front would be popped by a drain right
+    /// now — mirrors the gates of `drain_ready_family_sequencer_entries_locked`
+    /// (unadmitted checkpoints, PendingRoot front, prior-open-root fencing).
+    /// Gated-but-present entries return false: the event that lifts their
+    /// gate (admission completion, root clear) schedules its own drain.
+    fn family_has_actionable_front_entry(&self, family: &str) -> bool {
+        if self.unadmitted_checkpoints.load(Ordering::Acquire) > 0 {
+            return false;
+        }
+        let Ok(map) = self.family_sequencers_by_family.lock() else {
+            return false;
+        };
+        let Some(state) = map.get(family) else {
+            return false;
+        };
+        let Some((order, entry)) = state.entries.first_key_value() else {
+            return false;
+        };
+        if matches!(entry, FamilySequencerEntry::PendingRoot) {
+            return false;
+        }
+        let entry_root_sid = match entry {
+            FamilySequencerEntry::ReadyCommand(command) => Some(command.root_sid.as_str()),
+            FamilySequencerEntry::AppliedSideEffects { applied, .. } => {
+                Some(applied.command.root_sid.as_str())
+            }
+            _ => None,
+        };
+        !self
+            .family_entry_blocked_by_prior_open_trace_root(
+                family,
+                order.started_at_ns,
+                entry_root_sid,
+            )
+            .unwrap_or(true)
+    }
+
+    /// Replaces a root's PendingRoot sequencer entry with `replacement` and
+    /// returns the family whose sequencer changed. The caller is responsible
+    /// for scheduling a drain of that family afterwards — this must stay a
+    /// constant-time map mutation because it runs on the serial trace ingest
+    /// worker, whose watermark checkpoint admission waits on (#2252).
+    fn replace_pending_root_entry(
         &self,
         root_sid: &str,
         replacement: FamilySequencerEntry,
@@ -3606,8 +3795,6 @@ impl ActorDaemonCoordinator {
             return Ok(None);
         };
         let family = slot.family.clone();
-        let exec_lock = self.side_effect_exec_lock(&family)?;
-        let _guard = exec_lock.lock().await;
         {
             let mut sequencers = self.family_sequencers_by_family.lock().map_err(|_| {
                 GitAiError::Generic("family sequencer map lock poisoned".to_string())
@@ -3636,8 +3823,6 @@ impl ActorDaemonCoordinator {
                 }
             }
         }
-        self.drain_ready_family_sequencer_entries_locked(&family)
-            .await?;
         Ok(Some(family))
     }
 
@@ -3842,20 +4027,27 @@ impl ActorDaemonCoordinator {
             .lock()
             .map_err(|_| GitAiError::Generic("trace ingress state lock poisoned".to_string()))?;
         for root_sid in roots {
-            if let Some(count) = ingress.root_open_connections.get_mut(root_sid) {
-                if *count > 1 {
-                    *count -= 1;
-                    continue;
-                }
-                ingress.root_open_connections.remove(root_sid);
+            if let Some(count) = ingress.root_open_connections.get_mut(root_sid)
+                && *count > 1
+            {
+                *count -= 1;
+                continue;
             }
             if !Self::trace_root_needs_close_marker(&ingress, root_sid) {
+                ingress.root_open_connections.remove(root_sid);
                 Self::clear_trace_ingress_root_locked(&mut ingress, root_sid);
                 continue;
             }
             if ingress.root_close_markers_enqueued.contains(root_sid) {
                 continue;
             }
+            // Keep the root registered as open until the ingest worker has
+            // processed its queued frames and this close marker
+            // (clear_trace_root_tracking removes the registration then). The
+            // reader runs ahead of the worker; clearing here would drop the
+            // open-root fence while the root's own command is still queued,
+            // letting a detached drain execute a later command's pass first
+            // and invert per-family side-effect order (#2252).
             ingress.root_close_markers_enqueued.insert(root_sid.clone());
             close_marker_candidates.push(root_sid.clone());
         }
@@ -4109,14 +4301,7 @@ impl ActorDaemonCoordinator {
                                 Err(error)
                             }
                             Err(panic_payload) => {
-                                let panic_msg =
-                                    if let Some(s) = panic_payload.downcast_ref::<String>() {
-                                        s.clone()
-                                    } else if let Some(s) = panic_payload.downcast_ref::<&str>() {
-                                        s.to_string()
-                                    } else {
-                                        "unknown panic".to_string()
-                                    };
+                                let panic_msg = panic_payload_message(panic_payload.as_ref());
                                 tracing::error!(
                                     component = "daemon",
                                     phase = "trace_ingest_worker",
@@ -4251,14 +4436,7 @@ impl ActorDaemonCoordinator {
                         }
                     }
                     Err(panic_payload) => {
-                        let panic_msg =
-                            if let Some(message) = panic_payload.downcast_ref::<String>() {
-                                message.clone()
-                            } else if let Some(message) = panic_payload.downcast_ref::<&str>() {
-                                message.to_string()
-                            } else {
-                                "unknown panic".to_string()
-                            };
+                        let panic_msg = panic_payload_message(panic_payload.as_ref());
                         tracing::error!(
                             component = "daemon",
                             phase = "checkpoint_ingress_worker",
@@ -4810,6 +4988,10 @@ impl ActorDaemonCoordinator {
         &self,
         family: &str,
     ) -> Result<(), GitAiError> {
+        // Register the in-flight pass BEFORE popping entries: completion
+        // fences must never observe an empty sequencer with no registered
+        // pass while popped entries are about to execute (#2252).
+        let _family_effect = self.begin_family_effect_guarded(family);
         let mut ready: Vec<(u64, FamilySequencerEntry)> = Vec::new();
         let mut progressed = false;
         {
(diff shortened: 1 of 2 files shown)


src/daemon.rs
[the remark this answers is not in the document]
svarlamov-git-ai: Fixed in the latest push. The in-flight registration is now RAII: `begin_family_effect_guarded` returns a `FamilyEffectGuard` whose `Drop` releases the registration on every exit path — the semaphore-acquire `?` returns, any other early return in the entry loop, panics unwinding through, and dropped tasks. The detached Applied task adopts its worker-side registration the same way (`adopt_family_effect` at the top of `run_detached_side_effects_for_applied_command`), and the worker now performs only infallible steps between `begin_family_effect` and the spawn that adopts it. On popped entries when execution fails: the only remaining pre-completion return is the semaphore-closed error, which is unreachable (the semaphore is never closed) and matches the pre-existing checkpoint-arm behavior.

src/daemon.rs
```diff
@@ -6837,44 +6872,118 @@ impl ActorDaemonCoordinator {
...
+                        "unknown panic".to_string()
+                    };
+                    Err(GitAiError::Generic(format!(
+                        "daemon command side effect panic: {}",
+                        panic_msg
+                    )))
+                }
+            }
+        };
+        let _ = self.end_family_effect(&family);
```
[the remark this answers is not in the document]
svarlamov-git-ai: Fixed in the latest push. `run_detached_side_effects_for_applied_command` no longer calls `end_family_effect` manually — it adopts the registration into a `FamilyEffectGuard` declared as the first local, so it drops last: after `append_command_completion_log` and all other completion bookkeeping, and after the exec-lock guard and semaphore permit release. Sync/await/shutdown can no longer observe the family as quiescent before the completion log write finishes.

src/daemon.rs
```diff
@@ -4854,7 +4919,7 @@ impl ActorDaemonCoordinator {
             return Ok(());
         }
 
-        let _ = self.begin_family_effect(family);
+        let _family_effect = self.begin_family_effect_guarded(family);
```
svarlamov-git-ai: **[P1] Race: entries are popped from the sequencer before the family effect is registered, so the completion fences can miss an in-flight pass**

In `drain_ready_family_sequencer_entries_locked`, ready entries are removed under the `family_sequencers_by_family` mutex (released at the end of the pop block), and only then does `begin_family_effect_guarded(family)` register the pass under the separate `inflight_effects_by_family` mutex. Between the two lock scopes the drain task holds only the family exec lock — which no waiter touches for this family, because `drain_all_ready_family_sequencers` filters to families with non-empty entries.

Interleaving (threads, no await needed):
1. Detached drain pops the entries (sequencer now empty), is preempted before `begin_family_effect_guarded`.
2. `sync.family` loop runs: `drain_all_ready_family_sequencers` skips the family (empty), `unadmitted_checkpoints == 0`, `has_inflight_family_effects()` false → breaks and returns status before the pass (working-log write / post-commit authorship) runs.
3. Same window in `drain_accepted_checkpoints` (graceful shutdown): for `ReadyCommand` entries the ingress quota is 0, so shutdown returns Ok and process exit kills the pass mid-write — exactly the hazard the comment at the shutdown check describes. (`Checkpoint` entries are still covered there by the ingress-quota reservation; `ReadyCommand` entries are not. `has_pending_daemon_work` is exposed the same way.)

This violates the invariant stated on `has_inflight_family_effects` ("passes register via `begin_family_effect` before the trace ingest watermark advances") — true for the non-sequencer fast path, but the sequencer drain registers only after popping.

Fix: register the guard before popping (or inside the same critical section), and drop it early when `ready` ends up empty.

src/daemon.rs
```diff
@@ -6837,44 +6897,119 @@ impl ActorDaemonCoordinator {
...
-                    {
-                        let _ = self.record_side_effect_error(&family, applied.seq, &error);
-                        tracing::error!(
-                            %error,
-                            %family,
-                            seq = applied.seq,
-                            "async completion log write failed"
-                        );
-                    }
+                    self.begin_family_effect(&family)?;
```
svarlamov-git-ai: **[P2] Detached side-effect passes lose per-family ordering that the inline worker used to guarantee**

Before this change, `Applied`-path side effects ran inline on the serial ingest worker and sequencer drains ran inline behind them, so per family every pass completed in exact ingestion order. Now a non-sequencer command's pass (`run_detached_side_effects_for_applied_command`) and a later command's drain are both detached tasks serialized only by the family exec lock — which provides mutual exclusion, not order. Tokio's `AsyncMutex` grants by first-poll order, so a later-spawned drain can win; `family_entry_blocked_by_prior_open_trace_root` cannot help because `clear_trace_root_tracking` removes the root before the spawn.

Most non-sequencer commands' passes are no-ops, but a non-FF HEAD move by one (e.g. `git am --abort` after a mid-conflict checkpoint) runs `handle_non_fast_forward_rewrite_with_operation` + `rename_working_log`. If a same-family `git commit` drain wins the race, `post_commit_from_working_log_with_recovery_timestamps` runs against a base whose working log has not been migrated yet → attributions dropped from the note, and the late rename strands/clobbers the working log. Trigger needs an ingest backlog (both terminal payloads queued together, so the worker goes spawn(A)→spawn(B) without yielding and tokio's LIFO slot favors B) — narrow, but the ordering invariant loss is real and nothing re-establishes it.

Worth either routing Applied passes through the same per-family queue as sequencer entries, or documenting why order-sensitive effects (non-FF rewrites) cannot occur on the Applied path.

src/daemon.rs
```diff
@@ -6974,7 +7118,11 @@ impl ActorDaemonCoordinator {
             self.wait_for_trace_ingest_processed_through().await;
             self.drain_all_ready_family_sequencers().await?;
 
-            if self.outstanding_checkpoint_state().0 == 0 {
+            // Detached side-effect passes (drains and non-sequencer
+            // commands) are invisible to the sequencer map once their
+            // entries are popped; exiting while one is in flight would let
+            // process teardown kill it mid-write (#2252).
+            if self.outstanding_checkpoint_state().0 == 0 && !self.has_inflight_family_effects() {
```
svarlamov-git-ai, line 7216: **[P2] Update/uptime restarts are not fenced on in-flight detached passes — only this graceful-shutdown path is**

This check correctly keeps `ControlRequest::Shutdown` from completing while a detached pass is in flight. But `daemon_update_check_loop` defers restarts only on `outstanding_checkpoint_state().0 > 0` (src/daemon.rs ~9345 and ~9369) and then calls `request_restart_after_update()` / `request_restart()` → `request_shutdown()` directly, bypassing `drain_accepted_checkpoints`. `run_daemon`'s teardown after `wait_for_shutdown()` never waits on `has_inflight_family_effects()` either, and `spawn_shutdown_deadline_enforcer` force-exits (`std::process::exit(70)`) ~5s after the flag flips.

Scenario: a multi-minute monorepo side-effect pass grinding on a detached task, zero outstanding checkpoints (the normal steady state this PR creates), update becomes ready → restart → pass killed mid working-log/notes write — the exact hazard the comment here names. The exposure existed in kind for inline passes on main, but this PR establishes the fence as the safety mechanism and makes long detached passes the steady state, so the restart paths should defer on `has_inflight_family_effects()` (and pending sequencer entries) too.

src/daemon.rs
```diff
@@ -4874,6 +4939,15 @@ impl ActorDaemonCoordinator {
             });
             match ready_entry {
                 FamilySequencerEntry::ReadyCommand(command) => {
+                    // Bound global side-effect concurrency: command passes do
+                    // blocking git work on runtime workers, and with drains
+                    // running on detached tasks an unbounded number of
+                    // simultaneously grinding families could otherwise occupy
+                    // every worker thread and starve the runtime (#2252).
+                    let _side_effect_permit =
```
svarlamov-git-ai: **[P2] Two grinding command passes now starve checkpoint processing in every family**

Merging the checkpoint semaphore into a single 2-permit `side_effect_semaphore` means long command passes and checkpoint passes compete for the same two permits. Two simultaneously grinding families (e.g. two monorepo checkouts/commits, each holding a permit for the full pass duration) leave zero permits, so every checkpoint drain in every other family — plus `run_detached_side_effects_for_applied_command` — blocks for the duration of the shorter grind. On main, checkpoints had two dedicated permits that command work could not consume.

The regression test only exercises one grinding family (one permit stays free); with two grinders its 5s assertion scenario would stall for the grind duration at the processing stage. Since checkpoint passes are short and latency-sensitive while command passes are unbounded, consider keeping a dedicated checkpoint semaphore (or reserving one permit for checkpoint entries) so admission-to-processing latency stays independent of unrelated grinds.

src/daemon.rs
```diff
@@ -6949,17 +7084,26 @@ impl ActorDaemonCoordinator {
...
+            // all already-ingested passes had completed. Ready entries are
+            // executed here or by their detached drains (serialized per
+            // family by the exec lock); passes already handed off are
+            // visible via the effect registrations they take before the
+            // watermark advances (#2252). Entries fail-closed behind a
+            // still-open trace root stay queued without blocking this loop,
+            // exactly as they did pre-detachment.
+            self.drain_all_ready_family_sequencers().await?;
+            if self.unadmitted_checkpoints.load(Ordering::Acquire) == 0
+                && !self.has_inflight_family_effects()
```
svarlamov-git-ai, line 7193: **[P3] sync.family can now be extended indefinitely by work that starts after the sync began**

The loop exits only when `has_inflight_family_effects()` is globally false at a sampled instant — there is no snapshot of the effects that existed when the sync started. Every new command in any unrelated repository registers an effect before its watermark advances, so on a machine with continuous multi-repo git activity (CI runners, several agents), each iteration can observe a fresh in-flight pass and the sync never converges until a global quiet moment. Pre-change, only newly arriving checkpoints extended the loop (`unadmitted_checkpoints`); newly ingested commands did not.

The cross-repo push case justifies waiting on effects that were already in flight, but not on effects registered after the sync's ingest fence passed. A snapshot (e.g. collect the currently registered guard identities/generation counter at loop entry and wait only for those) would preserve the cross-repo guarantee while keeping sync.family bounded.

tests/daemon_mode.rs
```diff
@@ -2126,6 +2126,314 @@ fn daemon_drains_independent_checkpoint_families_concurrently() {
...
+        .unwrap();
+
+    let sync = send_control_request_with_timeout(
+        &daemon_control_socket_path(&repo),
+        &ControlRequest::SyncFamily {
+            repo_working_dir: repo_workdir_string(&repo),
+        },
+        Duration::from_secs(30),
+    )
+    .expect("sync.family should succeed");
```
svarlamov-git-ai, line 2349: **[P3] Flake risk: sync.family is sent with no settle window after `git am`, and the fence can miss the not-yet-ingested root**

The trace reader decrements `root_open_connections` when it processes the socket close, after enqueueing the root's final payload. `wait_for_trace_ingest_processed_through_family` samples its seq target at loop entry, so this interleaving loses the am root: sync samples target M → reader enqueues the am completion payload at seq N > M and processes the close (root no longer open) → sync finishes waiting for M, sees no open roots, and proceeds before the worker processed N — i.e. before the detached pass registered its family effect and before `completion_entries_for_command(&repo, "am")` can ever be non-empty. The assertion then fails spuriously.

The cross-repo test below sleeps 500ms for exactly this reason ("Let the trace reader enqueue the push root before syncing"); this test sends sync immediately after `git am` returns. Adding the same settle sleep (or polling until the daemon has ingested the am root) removes the race.

src/daemon.rs
```diff
@@ -6837,44 +6897,119 @@ impl ActorDaemonCoordinator {
...
+                    %error,
+                    %family,
+                    seq = applied.seq,
+                    "async side-effect error"
+                );
+                return;
+            }
+        };
+        let _guard = exec_lock.lock().await;
+        let Ok(_side_effect_permit) = self.side_effect_semaphore.acquire().await else {
```
svarlamov-git-ai: **[P3] Early returns in the detached runner skip `append_command_completion_log`, which the inline path wrote unconditionally**

On main, every Applied command got a completion log entry even when the side-effect result was an error. Here the `side_effect_exec_lock` error return above and this closed-semaphore return exit before `append_command_completion_log` (and the `self.begin_family_effect(&family)?` at the spawn site skips the whole pass the same way). The test-sync barrier (`wait_for_daemon_total_completion_count`) counts one completion entry per traced top-level command, so an affected command stalls that barrier to its timeout. Reachability is poisoned-lock / closed-semaphore only, but it silently breaks a previously unconditional invariant — cheap to preserve by writing the completion entry with the error before returning.

src/daemon.rs
```diff
@@ -3586,18 +3630,43 @@ impl ActorDaemonCoordinator {
...
-            self.drain_ready_family_sequencer_entries(&family).await
+            self.schedule_family_drain(family);
         } else {
-            self.drain_all_ready_family_sequencers().await
+            self.schedule_all_ready_family_drains();
         }
     }
 
-    async fn replace_pending_root_entry(
+    fn schedule_family_drain(self: &Arc<Self>, family: String) {
```
svarlamov-git-ai, line 3703: **[P3] No coalescing of scheduled drains: tasks pile up on a grinding family's exec lock**

Every cleared trace root and every admission batch spawns a fresh detached drain (here and in `schedule_all_ready_family_drains`), and each spawned task then parks on the target family's `AsyncMutex` waiter queue (or occupies one of `drain_all`'s two `buffer_unordered` slots) for the full duration of any in-progress grind. During a long pass with ongoing machine-wide git/checkpoint activity, parked drain tasks accumulate one per event with no dedup, then stampede through draining nothing when the grind ends. The old inline drains had natural backpressure from the serial worker.

A per-family "drain scheduled" flag (cleared when the drain actually runs) would coalesce these into at most one pending drain per family.

src/daemon.rs
```diff
@@ -6837,44 +6897,119 @@ impl ActorDaemonCoordinator {
...
+            let future = self.maybe_apply_side_effects_for_applied_command(
+                Some(&family),
+                &applied,
+                &mut commit_file_timestamp_snapshots,
+            );
+            let caught = std::panic::AssertUnwindSafe(future);
+            match futures::FutureExt::catch_unwind(caught).await {
+                Ok(result) => result,
+                Err(panic_payload) => {
+                    let panic_msg = if let Some(message) = panic_payload.downcast_ref::<String>() {
```
svarlamov-git-ai: **[P4] Fifth copy of the panic-payload message extraction (CLAUDE.md rule 5: reuse)**

This `downcast_ref::<String>` / `downcast_ref::<&str>` / "unknown panic" block now exists five times in this file (trace ingest worker, checkpoint ingress worker, command drain arm, checkpoint drain arm, and here). Worth extracting a small `fn panic_message(payload: &dyn Any) -> String` helper and using it at all five sites.

tests/daemon_mode.rs
```diff
@@ -2126,6 +2126,314 @@ fn daemon_drains_independent_checkpoint_families_concurrently() {
...
+
+    let destination_path = destination.path().to_string_lossy().to_string();
+    local
+        .git_without_test_sync_for_test(
+            &["push", destination_path.as_str(), "HEAD:refs/heads/pushed"],
+            &[],
+        )
+        .unwrap();
+    // Let the trace reader enqueue the push root before syncing.
+    thread::sleep(Duration::from_millis(500));
```
svarlamov-git-ai: **[P4] Fixed sleeps gate whether these tests exercise the scenario at all**

The four grind-based tests rely on fixed settle sleeps (1500ms in the two grind tests, 1000ms before shutdown, 500ms here). On a loaded runner, if ingestion has not reached the delayed pass by then, the grind tests pass vacuously (nothing is grinding when the checkpoint arrives), while the shutdown and cross-repo tests fail spuriously (the shutdown/sync fence has nothing enqueued to wait on, so no completion entry / no pushed note ever appears). Polling for the daemon-side precondition (e.g. the "checkpoint start"/side-effect-entered log line, or ingest progress) instead of sleeping would make the setup deterministic in both directions.

src/daemon.rs
svarlamov-git-ai: Fixed: `drain_ready_family_sequencer_entries_locked` now creates the RAII effect registration *before* popping entries, so the fences can never observe an empty sequencer with no registered pass while popped entries await execution.

src/daemon.rs
svarlamov-git-ai: Fixed properly: the detached-task path for non-sequencer commands is gone. Their side effects are now sequenced as `FamilySequencerEntry::AppliedSideEffects` entries, ordered by the command's start time in the same BTreeMap as every other pass, so a `git am` pass can never be reordered after (or overlap) a later command's drain — the drain executor runs them strictly in order under the exec lock, restoring the inline worker's ordering guarantee.

src/daemon.rs
svarlamov-git-ai: Fixed: both the update-restart and uptime-restart paths now also defer while `has_inflight_family_effects()`, mirroring the graceful-shutdown fence, so process teardown cannot kill an executing pass mid-write on those paths either.

src/daemon.rs
svarlamov-git-ai: Fixed: the semaphores are split again. Checkpoint side effects keep their own dedicated 2-permit semaphore (they run via `spawn_blocking` on the blocking pool, so they never contribute to runtime-worker starvation), and command/applied passes get a separate `command_side_effect_semaphore` (2 permits) that bounds worker-blocking git work. Two grinding command passes can no longer delay checkpoint processing in other families.

src/daemon.rs
svarlamov-git-ai: Acknowledged, leaving as-is: the exit condition only requires a quiescent instant with respect to *actively executing* passes (bounded at 2 command permits + checkpoint permits — not queued entries, which fail-closed entries never extend), and the pre-change fence had the same shape: `wait_for_trace_ingest_processed_through_family` re-derives its global watermark target on every iteration, so ongoing multi-repo activity could already extend sync before this PR. `sync.family` is a control/test API (no production hot path); if fleet telemetry shows sync latency regressions we can snapshot the family set at entry as a follow-up.

tests/daemon_mode.rs
svarlamov-git-ai: Fixed: the test now polls the daemon log for the new `test side-effect delay started` line (emitted by the delay hook when the pass actually begins executing) before sending sync.family, so the am root is guaranteed ingested and the pass registered — no settle-window race.

src/daemon.rs
svarlamov-git-ai: Mooted by the ordering fix: the detached runner is deleted. Applied passes now execute in the drain executor's `AppliedSideEffects` arm, which writes `append_command_completion_log` unconditionally after the pass, exactly like the `ReadyCommand` arm. The only remaining pre-completion early return is the never-closed semaphore `?`, which matches the pre-existing checkpoint-arm behavior.

src/daemon.rs
svarlamov-git-ai: Acknowledged, leaving as-is: parked drain tasks are intrusively queued futures (~KB each, no allocation while parked), bounded by the trace-ingest event rate for the grinding family — worst observed backlog in the incident was ~245 payloads. On lock release the first waiter drains everything and the rest no-op in ~1-3µs each. A per-family drain-scheduled flag would add dedup state and a re-check race (drain finished but flag not yet cleared strands an entry) for negligible gain.

src/daemon.rs
svarlamov-git-ai: Fixed: extracted `panic_payload_message` and used it at all five sites (trace ingest worker, checkpoint ingress worker, command arm, applied arm, checkpoint arm).

tests/daemon_mode.rs
svarlamov-git-ai: Fixed: all five grind-based tests now gate on the daemon-side precondition — the delay hook logs `test side-effect delay started` when the pass begins executing, and the tests poll for that line (20s budget) instead of fixed sleeps. A loaded runner can no longer make the grind tests pass vacuously or the fence tests fail spuriously.

src/daemon.rs
```diff
@@ -6811,67 +6937,60 @@ impl ActorDaemonCoordinator {
...
+                            );
+                        let started_at_ns = applied.command.started_at_ns;
+                        self.append_family_sequencer_entry(
+                            &family,
+                            started_at_ns,
+                            FamilySequencerEntry::AppliedSideEffects {
+                                applied: Box::new(applied),
+                                commit_file_timestamp_snapshots,
+                            },
+                        )?;
```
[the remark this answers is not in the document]
svarlamov-git-ai: Fixed in the latest push. Root cause: the reader thread cleared `root_open_connections` at socket close, ahead of the ingest worker, so the open-root fence dropped while the root's own command was still queued and an unrelated detached drain could execute a later command's pass first. `record_trace_connection_close` now keeps the root registered as open for roots that get a close marker; the registration is only removed by `clear_trace_root_tracking` when the worker processes the root's emission/close marker — i.e. strictly after the root's own `AppliedSideEffects`/`ReadyCommand` entry is in the sequencer. The fence therefore holds across the reader/worker gap and the earlier command's pass can never be inserted behind already-selected later work. The window cost is bounded (µs–ms: the worker is sequencing-only now), and the checkpoint/sync fences wait for the queued close marker via the watermark before consulting open roots, so no new blocking.

src/daemon.rs
```diff
@@ -9128,11 +9260,13 @@ fn daemon_update_check_loop(coordinator: Arc<ActorDaemonCoordinator>, started_at
             Ok(DaemonUpdateCheckResult::UpdateReady) => {
                 let (outstanding_checkpoints, retained_checkpoint_bytes) =
                     coordinator.outstanding_checkpoint_state();
-                if outstanding_checkpoints > 0 {
+                // Also defer while a detached side-effect pass is executing:
+                // process teardown would kill it mid-write (#2252).
+                if outstanding_checkpoints > 0 || coordinator.has_inflight_family_effects() {
```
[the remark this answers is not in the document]
svarlamov-git-ai: Fixed in the latest push. Both automatic restart branches now defer on `has_pending_attribution_work()`: outstanding checkpoints, queued trace payloads, actionable sequencer entries (covering the append-to-drain-registration window you describe), and executing passes. `PendingRoot` placeholders are deliberately excluded so an idle interactive command (a rebase waiting on an editor overnight) cannot defer updates forever — its entry stays fail-closed in the sequencer and only becomes deferring once actionable. Added a unit test (`pending_attribution_work_fence_ignores_idle_pending_roots`) covering the idle-PendingRoot, queued-entry, and executing-pass cases the restart decision depends on.

src/daemon.rs
```diff
@@ -3586,18 +3678,43 @@ impl ActorDaemonCoordinator {
...
     }
 
-    async fn replace_pending_root_entry(
+    fn schedule_family_drain(self: &Arc<Self>, family: String) {
+        let coordinator = Arc::clone(self);
+        tokio::spawn(async move {
+            if let Err(error) = coordinator
+                .drain_ready_family_sequencer_entries(&family)
+                .await
+            {
```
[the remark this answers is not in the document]
svarlamov-git-ai: Fixed in the latest push. `schedule_family_drain` now coalesces to at most one scheduled-or-running drain task per family via a `scheduled_family_drains` marker. The marker is released atomically with an actionable-front re-check after each pass (mirroring the drain's own gates: unadmitted checkpoints, PendingRoot front, prior-open-root fencing), so an entry appended before any schedule call is always covered — by the running task's next pass or by the fresh task the caller spawns after release. Gated entries deliberately release the marker: the event that lifts their gate (admission completion, root clear) schedules its own drain, so no polling and no busy loop on a blocked family. Error paths release the marker before exiting.

PR #2349, opened 2026-09-14
fix(daemon): spawn Windows bg daemon directly

mgmonteleone:
## Problem

On Windows, `git-ai bg start` currently launches the detached background daemon through `powershell.exe Start-Process`. That adds an extra process hop with higher startup latency, and failures inside `Start-Process` are not surfaced by the wrapper beyond whether PowerShell itself launched.

Replacing that with a direct `Command::spawn` needs one Windows-specific guard: stable Rust's Windows process spawn path inherits handles, so a caller that captures the `bg start` wrapper's stdout or stderr can accidentally leave a pipe write end duplicated into the long-lived daemon. The wrapper exits, but readers waiting for EOF can keep waiting because the daemon still owns an inherited copy of the handle.

## Fix

- Spawn `git-ai bg run` directly from `spawn_daemon_run_detached`, using the same executable resolution, working directory, null daemon stdio, and Git/`GIT_AI` environment scrubbing as the non-Windows path.
- Keep the existing Windows creation-flag policy (`CREATE_NO_WINDOW | CREATE_NEW_PROCESS_GROUP | CREATE_BREAKAWAY_FROM_JOB`) with the existing fallback that retries without `CREATE_BREAKAWAY_FROM_JOB` if that flag combination is rejected.
- Before spawning the daemon on Windows, clear `HANDLE_FLAG_INHERIT` on the wrapper process's own stdin/stdout/stderr handles, scoped to the spawn: an RAII guard reads each handle's original flags with `GetHandleInformation`, clears the inherit flag, and restores the original flags on drop once the spawn attempt (including the `CREATE_BREAKAWAY_FROM_JOB` fallback retry) completes, so the wrapper's own stdio inheritance is never left cleared beyond the spawn itself. Distinct handle values are read/cleared/restored exactly once (stdin, stdout, and stderr can alias the same kernel handle, e.g. `> out.log 2>&1`), and are restored in reverse order of saving on drop. Missing, invalid, already non-inheritable, or otherwise unsupported standard handles are ignored after debug logging; the daemon's explicit `Stdio::null()` handles are still supplied separately.
- Leave the PowerShell-based self-update path unchanged; that path is solving the separate problem of retrying replacement of the running executable after this process releases its file lock.

## Verification

Run on Linux:

- `cargo build --lib` — passed.
- `cargo fmt --check` — passed.
- `cargo clippy --all-targets -- -D warnings` — passed.
- `cargo test --test daemon_mode daemon_start_spawns_detached_run_process -- --test-threads=1` — passed.
- `cargo test --test daemon_mode daemon_ -- --test-threads=1` — 99 passed / 17 failed on this branch; the same 17 failures reproduce on unmodified `main` at `0670e7ef27590af0e8ff5409267f3f4b09b8fcb4` in the same Linux environment, so this did not identify a branch regression.

Also cross-checked with a Windows target added to this Linux environment (`rustup target add x86_64-pc-windows-gnu` plus `gcc-mingw-w64-x86-64`):

- `cargo check --target x86_64-pc-windows-gnu --all-targets` — passed (compiles the Windows-only spawn path, the RAII guard, and both `#[cfg(windows)]` test modules).
- `cargo clippy --target x86_64-pc-windows-gnu --all-targets -- -D warnings` — passed.

## Known limitations

- The regression test, `daemon_windows_bg_start_wrapper_stdout_closes_promptly_while_daemon_stays_up`, and the new unit tests for the inherit-flag guard are `#[cfg(windows)]`. They were not executed in the Linux-only validation environment (only compiled, via the cross-target check above). They should be covered by the upstream `windows-latest core` CI shard, which runs the top-level Rust test targets including `tests/daemon_mode.rs` on pull requests.
- Native Windows smoke validation was not available before opening this PR. The important behaviors for Windows CI or maintainer validation are that `git-ai bg start` exits promptly when its output is captured, the spawned daemon remains reachable after the wrapper exits, and normal `git-ai bg status` / `git-ai bg shutdown --hard` still work.
- Handle-inheritance flags are process-wide kernel state, not scoped to a single spawn call: a concurrent spawn from another thread in the same process during the brief window the guard holds the cleared flag could still observe it. git-ai's daemon auto-start does not run concurrently with a passthrough git spawn in the same process today, so this is not a practical issue currently, but it would need an additional process-wide lock if that assumption ever changes.


mgmonteleone:
Pushed ef3978b2 addressing the review flag on `disable_inherit_for_own_stdio` (standard handles left permanently non-inheritable, which would break later `spawn_git_passthrough` children when the daemon auto-starts inside an ordinary command).

The inheritance change is now scoped to the spawn: `disable_inherit_for_own_stdio()` returns an RAII guard that records each std handle's original flags via `GetHandleInformation`, clears `HANDLE_FLAG_INHERIT` only for handles that actually had it set, and restores the original flags on drop. The guard lives across the preferred-flags attempt and the `CREATE_BREAKAWAY_FROM_JOB` fallback retry and drops on every exit path (success, first failure, fallback failure). Two `#[cfg(windows)]` unit tests cover clear-then-restore around a simulated successful spawn and around an early-`Err` path. The process-wide nature of the flag is called out in a code comment and in the PR body's known limitations; a lock was not added since daemon auto-start does not run concurrently with passthrough spawns in the same process today.

Also cross-checked with `cargo check`/`clippy --target x86_64-pc-windows-gnu --all-targets` so the Windows-only code and tests compile; PR body updated.

mgmonteleone:
Pushed `0af390c` (single commit, amended) addressing the shared-handle review thread on the RAII guard:

- `disable_inherit_for_own_stdio()` is now a thin wrapper over `disable_inherit_for_handles(&[HANDLE])`, which deduplicates by raw handle value before calling `GetHandleInformation`. When two standard streams alias the same kernel handle (e.g. `> out.log 2>&1`), the handle is read, cleared, and restored exactly once, so the second visit can no longer overwrite the saved original flags with the already-cleared value.
- `RestoreStdioInherit::drop()` restores in reverse order of saving (LIFO), so the guard stays correct even if overlapping entries were ever reintroduced.
- New `#[cfg(windows)]` unit test `restores_original_flags_when_handles_alias_the_same_value` builds the guard from `&[stdout, stdout]` and asserts the inherit flag is cleared while held and back to its original value after drop.

`cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` pass on Linux and on `--target x86_64-pc-windows-gnu`; `cargo check --all-targets --target x86_64-pc-windows-gnu` compiles the new test. PR description updated to match.

PR #2351, opened 2026-09-14
feat(checkpoint): add Augment Code preset and hook installer

### LINKED ISSUE #1071: Feature Request: support Augment
I’d like to request support for Augment Code.

References
 - [Augment Code docs](https://docs.augmentcode.com)
 - [Hook](https://docs.augmentcode.com/cli/hooks)

mgmonteleone:
Closes #1071.

## Summary

Adds first-class Git AI support for Augment Code/Auggie by introducing an `augment` checkpoint preset and a hook installer for `~/.augment/settings.json`.

The preset autodetects the hook payload generation from the JSON shape:

- `hook_event_name` without `hook_type` → v1 Auggie hook payloads
- `hook_type` without `hook_event_name` → v2 Auggie/Cosmos Agent hook payloads
- both or neither discriminator → `PresetError` instead of guessing
- `null` discriminator values are treated as absent

For v1 payloads, the preset handles file-edit tools (`save-file`, `str-replace-editor`, `remove-files`, and `apply_patch`) and shell commands (`launch-process`). For v2 payloads, it handles file-edit tools (`write`, `edit`) and the shell tool under either of its two observed names, `terminal` (the documented current name) or `bash` (the ACP/session tool-call model's name). Both shell names route to the same bash/stat-diff attribution path. Read-only tools and lifecycle/notification events are silent no-ops so the catch-all hook does not surface warning noise during normal agent use.

## Approach

The implementation follows the existing agent-preset model: parse hook JSON into `PreFileEdit`/`PostFileEdit` and `PreBashCall`/`PostBashCall` events, then let the existing checkpoint machinery handle attribution. It does not add git work to the trace2 ingestion path.

The installer writes one command hook for each of `PreToolUse` and `PostToolUse` under Augment's catch-all `".*"` matcher:

- command: `git-ai checkpoint augment --hook-input stdin`
- idempotent reinstall: no file change when the desired hooks are already present
- uninstall: removes only hooks owned by the Augment preset, preserving user-defined hooks. Ownership is decided structurally, not by substring: a command's first word (the executable, quoting-aware) must resolve to a `git-ai`/`git-ai.exe` binary, and the next two words must be exactly `checkpoint augment`. A user-defined hook that merely mentions those words in passing (e.g. inside an `echo`) is never mistaken for one the installer owns
- `settings.json` is parsed as JSONC (`//` and `/* */` comments, trailing commas), matching Augment's documented settings-file format, so a docs-valid settings file no longer blocks hook install or mis-reports check state
- install also sweeps owned git-ai commands out of every *other* matcher block for the event (a hand-written legacy matcher, or a duplicate catch-all block) before writing the canonical one, so a stale entry left behind by an earlier install can never keep firing a duplicate checkpoint; `hook_status` reports "needs update" whenever an owned command still lives outside the canonical block

The ordinary v2 hook payload carries no session id, conversation id, cwd, transcript reference, or per-call tool identifier: lifecycle events are only `{hook_type}`, and `PostToolUse` is only `{hook_type, tool_name, tool_input, tool_result/tool_error}`. A sole candidate file under `~/.augment/sessions-v2/` is not a safe substitute either — it can be stale, concurrent sessions can share a workspace, resume reuses a stored id, and fork copies history under a new id — so the preset deliberately never mines the session directory or reads environment variables to recover one. By design, then, an unmodified v2 hook produces a stable, workspace-scoped session id (`generate_session_id(cwd, "augment")`) and `model = "unknown"`, so every v2 conversation in the same workspace currently checkpoints under one shared session.

To support exact per-conversation attribution going forward without requiring a git-ai release, the preset now also honours an explicit, payload-supplied `conversation_id` — the same field the v1 payload already documents — using it verbatim as the session id whenever it is present as a non-empty string. This makes the preset forward-compatible with either a future hook-API addition that supplies a real conversation id, or an editor/extension-side bridge that already knows the id and can pass it through when invoking the hook. A `null` or absent `conversation_id` falls back to the existing workspace-derived hash; a present-but-wrong-type or present-but-empty value is a `PresetError` rather than a silent fallback, consistent with how this preset treats other malformed discriminator fields.

New v2 unit tests cover this: a payload-supplied `conversation_id` is used verbatim as the session id, two payloads with different `conversation_id`s in the same working directory produce different session ids, `null`/absent falls back to the workspace-derived hash, and a non-string or empty-string `conversation_id` is a `PresetError`.

## Shell tool name handling (v2)

Augment's published tool-permissions documentation lists `terminal` as the current shell tool name, while the ACP/session tool-call data model separately uses `bash` for shell calls. Since either name can plausibly appear as `tool_name` in a v2 `PostToolUse` payload, the classifier accepts both `terminal` and `bash` so shell-command attribution is not silently dropped.

Coverage includes:

- a classifier unit-test case for `terminal` alongside `bash`
- an in-process v2 parser test asserting `tool_name: "terminal"` produces a `PostBashCall`
- an integration test sending a v2 `PostToolUse` payload with `tool_name: "terminal"` and a `tool_input.command`, asserting it produces an attributed bash checkpoint

### Known limitation: fixed shell `tool_use_id`

Neither the documented v1 hook payload nor the v2 payload this preset was built against carries a per-invocation shell identifier (no `tool_use_id`/`tool_call_id`/`call_id` field), unlike every other preset with a shell path in this codebase, which reads such a field with a fixed fallback. Both bash branches here always emit a fixed `tool_use_id` ("bash"), documented as a known limitation at the emission site. Since the daemon pairs pre/post snapshots by `(external_session_id, tool_use_id)`, two *concurrent* shell tool calls within one Augment session share this single slot: the first call's post-hook can diff against the second call's pre-snapshot, or find none at all. If Augment's hook payload ever adds a stable per-call id, it should be wired in the same way the other presets already do.

## Settings file format (JSONC)

Augment documents `~/.augment/settings.json` as JSONC — comments and trailing commas are explicitly supported by the CLI configuration format. The installer now parses that format through a shared `parse_jsonc_settings()` helper before converting to `serde_json::Value`, so JSONC settings work consistently across:

- `check_hooks`
- `install_hooks`
- `uninstall_hooks`

Invalid content that still cannot be parsed, and valid JSONC whose root is not an object, is still rejected for install without clobbering the file. Empty or whitespace-only settings content is treated as an empty object, matching the "no settings yet" case. A JSON number `serde_json` cannot represent (e.g. an exponent that overflows to infinity) is also a parse error rather than being silently coerced to `null`, so install can never persist `null` over a value the parser could not represent; uninstall likewise no-ops on such content instead of writing anything.

### Known limitations

- Comment/trailing-comma preservation is not implemented. Parsing accepts JSONC, but if `install_hooks`/`uninstall_hooks` needs to rewrite the file, it reserialises the document as plain pretty-printed JSON — any comments in the original file are dropped on that rewrite. This is covered by a dedicated test (`install_hooks_reserializes_jsonc_settings_as_json`) and can be revisited by migrating to a comment-preserving CST-based rewrite (as already used for VS Code's `chat.useHooks` settings) if this becomes a problem in practice.

## Relationship to #1299

#1299 added v1-only Augment support but has not progressed recently. This PR supersedes that approach with a compact v1+v2 implementation while keeping the scope focused on the preset, installer, tests, and changelog entry.

## Verification

Validated on Linux with Rust 1.93.0:

- `cargo build --lib` — passed
- `cargo clippy --all-targets -- -D warnings` — passed
- `cargo fmt --check` — passed
- `task test TEST_FILTER=augment` — 34 unit tests passed; 8 integration tests passed with `CI=true`. When run as root outside CI, the same command has 3 known stdout-pollution failures from git-ai's elevated-privilege warning, reproduced on the unmodified base checkout and unrelated to this change.
- `task test TEST_FILTER=test_tool_classification_all_agents` — passed, covering the `terminal` classifier case
- `task test TEST_FILTER=install_hooks` — passed twice, covering installer-related tests
- `task test CARGO_TEST_ARGS="--lib"` — completed with one pre-existing failure in `daemon::ref_cursor::...untraced_claim_keeps_claims_made...`, reproduced on the unmodified base checkout

New tests added for the JSONC settings support:

- `jsonc_settings_check_install_check_uninstall_round_trip` — a settings file with `//` line comments, `/* */` block comments, and trailing commas: `check_hooks` reports "not installed" without erroring, `install_hooks` succeeds and preserves all pre-existing user keys and user-defined hooks, a follow-up `check_hooks` reports "installed" and up to date, and `uninstall_hooks` removes only the Augment-preset hook while leaving user keys/hooks intact.
- `install_hooks_reserializes_jsonc_settings_as_json` — documents that installing over JSONC content succeeds but rewrites the file as plain JSON (comments dropped); see Known limitations above.
- `whitespace_only_settings_file_treated_as_empty` — an empty/whitespace-only settings file still behaves as an empty object for both `check_hooks` and `install_hooks`.
- `invalid_or_wrong_shaped_settings_json_errors_or_no_ops` — extended to assert that a failed install never modifies the on-disk file, to cover an additional truly-invalid-JSON shape, and to cover a JSON number `serde_json` cannot represent (`1e400`): install fails with a clear error and leaves the file untouched, and uninstall no-ops without writing.
- `parse_jsonc_settings_rejects_unrepresentable_number` / `parse_jsonc_settings_rejects_unrepresentable_number_nested` — unit tests on the shared parser confirming an unrepresentable number is an `Err`, including when nested inside an array/object, rather than silently becoming `null`.

New tests added for the legacy/duplicate hook sweep on install:

- `legacy_matcher_owned_command_swept_into_canonical_on_install` — a hand-written matcher block holding an owned git-ai command plus an unrelated hook: after install, exactly one owned command exists for the event (in the canonical catch-all block), and the unrelated hook survives untouched in the original block; a follow-up `hook_status` reports fully up to date, and a reinstall is a no-op.
- `duplicate_catch_all_blocks_consolidated_on_install` — two separate `".*"` matcher blocks each holding an owned command: after install, only one catch-all block remains with exactly one owned command; reinstall is a no-op.
- `jsonc_settings_with_legacy_owned_hook_sweeps_via_jsonc_parse_path` — the same legacy-sweep behavior through the JSONC parse path (comments, trailing commas), confirming it isn't specific to strict-JSON input.
- `hook_status_matrix` extended with cases where an owned command lives outside the canonical block (a legacy matcher, or a second catch-all block): `hook_status` now reports "needs update" in both cases even though the canonical block alone would otherwise look current.
- `is_git_ai_augment_command_ownership_matrix` — table-driven coverage of the ownership predicate: bare/quoted-with-spaces/`.exe` paths followed by `checkpoint augment` are owned; a wrapper script whose name merely contains `git-ai`, the right binary with the wrong preset (`checkpoint claude`), and a hook that only mentions the words in passing (e.g. `echo git-ai checkpoint augment`) are all correctly rejected.

I did not run native macOS or Windows CI locally; those platforms should be covered by the repository's normal CI matrix.

## Notes for reviewers

- Augment's hook matcher syntax uses regex; the catch-all matcher is `".*"`, not Claude Code's literal `"*"`.
- Unsupported/read-only tool invocations return no checkpoint events and no stderr. This is intentional because the installed catch-all matcher runs on every tool call.
- v1 can read `context.modelName` only when Augment hook metadata opts into user context; otherwise the model remains `"unknown"`.
- If maintainers want README icon-table parity for Augment, I am happy to add that in a follow-up once the preferred asset/name is available.


mgmonteleone:
Pushed 16fe1386 addressing both review flags:

- **Legacy hooks remain duplicated** (`src/mdm/agents/augment.rs`): `install_hooks_at` now sweeps owned git-ai commands out of every non-canonical block for the event (legacy hand-written matchers and duplicate `".*"` blocks) before writing the canonical one; a block is dropped only if the sweep empties it, unrelated hooks are preserved. `hook_status` reports "needs update" whenever an owned command still lives outside the canonical block. Tests: `legacy_matcher_owned_command_swept_into_canonical_on_install`, `duplicate_catch_all_blocks_consolidated_on_install`, `jsonc_settings_with_legacy_owned_hook_sweeps_via_jsonc_parse_path`, plus two new `hook_status_matrix` cases.
- **Concurrent shell snapshots** (`presets/augment.rs`): checked for an identifier to read — neither the documented v1 payload nor the v2 payload this preset was built against carries a per-invocation id (`tool_use_id` / `tool_call_id` / `call_id`), unlike the other presets with a shell path, which read one with a `"bash"` fallback. Rather than invent a pairing scheme in the daemon, the fixed id stays and the limitation is now explicit: a doc comment on the shared dispatch fn, pointers at both emission sites, and a "Known limitation" section in the PR body. If Augment's payload gains a stable per-call id it wires in the same way the other presets do.

PR body updated accordingly.

mgmonteleone:
Pushed `4316517` (single commit, amended) addressing the two remaining review threads:

- **Ownership predicate** (`src/mdm/agents/augment.rs`): `is_git_ai_augment_command` no longer does substring matching. It splits the command with quoting-aware shell-word semantics, requires the first word's basename to be `git-ai`/`git-ai.exe`, and requires the next two words to be exactly `checkpoint augment`. `echo git-ai checkpoint augment`, `git-ai checkpoint claude`, and wrapper binaries whose name merely contains `git-ai` are no longer treated as owned. Follows the same shape as the Codex installer's structural notify-args check. Covered by `is_git_ai_augment_command_ownership_matrix`.
- **Unrepresentable JSONC numbers** (`src/mdm/utils.rs`): `jsonc_value_to_serde` / `parse_jsonc_settings` now return `Result` and propagate the error (including nested arrays/objects) instead of coercing to `null`. Install fails with a clear error and leaves the file untouched; uninstall keeps its existing no-op-without-writing behaviour on unparseable content. Covered by `parse_jsonc_settings_rejects_unrepresentable_number{,_nested}` and the extended `invalid_or_wrong_shaped_settings_json_errors_or_no_ops`.

`cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, and `cargo test --lib mdm::` (273 passed) are clean. PR description updated to match.

mgmonteleone:
Re the v2 session-id thread (`presets/augment.rs`, workspace-scoped `generate_session_id`):

Confirmed. The current Auggie v2 hook payload carries no session id, conversation id, cwd, transcript reference, or tool-call id (`SessionStart`/`Stop` are only `{hook_type}`; `PostToolUse` is only `{hook_type, tool_name, tool_input, tool_result/error}`). Mining `~/.augment/sessions-v2/` or env vars was considered and rejected as unsafe (stale files, concurrent sessions, resume/fork id reuse).

Fix, amended into this PR's single commit (cb91995b): `parse_v2` now uses a payload-supplied `conversation_id` (same field v1 already documents) verbatim as the session id when present as a non-empty string; absent/`null` falls back to the workspace-derived hash; non-string or empty is a `PresetError`. Unmodified v2 hooks keep today's workspace-scoped id until the hook API or an extension bridge supplies a real one.

Issue #2345, opened 2026-09-12
[Bug]: OpenCode plugin fails to load in OpenCode 2.0 (PluginModule.LoadError)


dsillman2000:
### What broke?

Upgrading OpenCode from 1.18 to 2.0 breaks the `git-ai` plugin. OpenCode 2.0 completely rewrote the plugin architecture, and existing OpenCode 1.x plugins are rejected on startup with a schema validation error.

As a result, `~/.config/opencode/plugins/git-ai.ts` (installed via `git-ai install-hooks` from `agent-support/opencode/git-ai.ts`) fails to load, and no checkpoints are recorded during tool execution.

OpenCode 2 server log error (`~/.local/share/opencode/log/opencode.log`):
```text
level=WARN message="failed to load plugin" target=/Users/.../.config/opencode/plugins/git-ai.ts cause="Cause([Fail(PluginModule.LoadError: Plugin must export a default definition with an id and an effect or setup function. (cause: SchemaError(Expected object at ["default"])))])"
```

### Steps to reproduce

1. Install or upgrade OpenCode to 2.0 (`opencode --version` >= 2.0.0).
2. Install git-ai hooks with `git-ai install-hooks` (writes `~/.config/opencode/plugins/git-ai.ts`).
3. Start OpenCode or restart the service with `opencode service restart`.
4. Inspect `~/.local/share/opencode/log/opencode.log` or run `opencode plugin list`.
5. Observe that the plugin fails to load with `PluginModule.LoadError`, and `git-ai status` records no AI checkpoints for edits made within OpenCode.

### Expected vs actual behavior

- **Expected**: OpenCode loads `git-ai.ts` and invokes `execute.before` and `execute.after` hooks to checkpoint AI edits.
- **Actual**: OpenCode 2.0 rejects the plugin because it exports an async function returning a hook dictionary (OpenCode 1.x format) instead of an object definition with `id` and `setup(ctx)` created via `Plugin.define(...)` (OpenCode 2.0 format).

### Diagnostics (`git-ai debug`)

```shell
git-ai debug report
Generated (UTC): 2026-09-12T14:51:42.888977+00:00

== Versions ==
Git AI version: 1.7.4
Git AI binary: ~/.git-ai/bin/git-ai
Git binary path: /opt/homebrew/bin/git
Git binary realpath: /opt/homebrew/Cellar/git/2.52.0_1/bin/git
Shell git lookup command: /bin/zsh -lc 'which git'
[...]

PR #2364, opened 2026-09-18
fix(opencode): support the OpenCode V2 plugin API

lenucksi:
## Problem

OpenCode 2.x rejects the plugin in its current form:

```
PluginModule.LoadError: Plugin must export a default definition with an id and an effect or setup function.
```

The plugin default-exports an async function returning a `"tool.execute.before"` / `"tool.execute.after"` hook object (the V1 contract). Under OpenCode 2.x it never loads, so no checkpoints are recorded. Tracked in #2345.

## Fix

Support both APIs from one default export, following the documented migration path (OpenCode 2 plugin docs, "Support V1 and V2 from one package"):

```ts
export default {
  ...Plugin.define({ id: "git-ai", async setup(ctx) { /* V2 hooks */ } }),
  async server(ctx) { /* V1 hooks (OpenCode >= 1.18.29) */ },
}
```

The shared checkpoint core is unchanged. Only the payload mapping and the context directory source differ:

| | V1 | V2 |
| --- | --- | --- |
| hook registration | returned object keys | `ctx.tool.hook("execute.before" \| "execute.after", event)` |
| tool call id | `input.callID` | `event.id` |
| tool input | `output?.args ?? input.args` | `event.input` |
| metadata (after) | `output?.metadata` | `event.result?.metadata` when `event.status === "completed"` |
| project dir | `ctx.worktree \|\| ctx.directory` | `ctx.location.project?.directory ?? ctx.location.directory` |

Notes:

- V1 is kept on purpose so `git-ai install-hooks` keeps working on OpenCode 1.18.x; OpenCode calls `server()` for V1 and `setup()` for V2.
- `process.cwd()` is only a last-resort fallback. Unlike when #2205 was written, the V2 context now exposes `ctx.location`, so the project directory no longer has to be guessed.
- Dependency change: `@opencode-ai/plugin` → `@opencode/plugin@^2.0.8`. `package-lock.json` regenerated, and the Rust content assertions in `src/mdm/agents/opencode.rs` updated to match the new export shapes.

## Verification

- `npm run type-check` (`tsc --noEmit`) against `@opencode/plugin@2.0.8`: clean
- `npm ci` lock consistency: clean
- Rust content assertions updated; the Rust toolchain was not available locally, so CI should be the reference run

Refs #2345, #2205, #2171.

---

<a href="https://app.devin.ai/review/git-ai-project/git-ai/pull/2364" target="_blank"><picture><source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=4">Devin Review</picture></a>


MohakBajaj:
Thanks for this — the V1/V2 mapping table matches what I derived from the `@opencode/plugin@2.0.16` type definitions (`event.id`, `event.input`, `event.result?.metadata` on `status === "completed"`, `ctx.location`).

One gap to check before merge, specifically for the **global** install path that `git-ai install-hooks` uses (a bare file at `~/.config/opencode/plugins/git-ai.ts` with no surrounding `node_modules`):

After correcting the default export, the runtime `import { Plugin } from "@opencode/plugin"` cannot be resolved on **OpenCode 2.0.1**. The loader then rejects the module with:

```
level=WARN message="failed to load plugin" target=/Users/.../.config/opencode/plugins/git-ai.ts
cause="Cause([Die(ResolveMessage: Cannot find package '@opencode/plugin' imported from /Users/.../.config/opencode/plugins/git-ai.ts)])"
ref=err_6e7c3263
```

So fixing the export shape alone isn't sufficient for a global `install-hooks` install — the import still fails. I got it loading by dropping the import and exporting the plain object the loader's schema actually asks for (`Plugin.define` is an identity helper):

```ts
export default {
  id: "git-ai",
  async setup(ctx) {
    await ctx.tool.hook("execute.before", (event) => { /* ... */ })
    await ctx.tool.hook("execute.after", (event) => { /* ... */ })
  },
}
```

`opencode plugin list` then reports `git-ai  local  .../plugins/git-ai.ts`.

Questions:
1. Does your PR's dual `Plugin.define(...)` + `server()` export resolve `@opencode/plugin` for a global `~/.config/opencode/plugins/` install, or only for project-local plugins where a `node_modules` exists? If it needs the package present, should the installer drop a minimal `package.json` and install `@opencode/plugin` beside the file?
2. Is OpenCode expected to alias `@opencode/plugin` for local plugin files? It didn't on 2.0.1.

Environment: OpenCode 2.0.1 (latest 2.0.16), git-ai 1.7.4, macOS arm64. Refs #2345, #2171.

